PR #7, opened 2017-01-18
TBM-4693 - Added waveform styles to react components.

troyblank:


components/horizontalTimeline/timelineVisualization.js
```diff
@@ -0,0 +1,91 @@
...
+          <div
+            className={`${styles['timeline-visualization']} js-timeline-visualization`}
+            ref={node => (this.timelineVisualization = node)}
+          >
+            <div
+              style={{ left: this.state.xPos, width: this.state.width }}
+              className={styles['scroll-container']}
+            >
+              <div className={styles['wave-form']} />
+              <div className={styles.cursor} style={{ left: 900 }} />
```
jhweaver: Where did the 900 come from? Does this not depend on the width of the screen?

troyblank: That is just to show the cursor over a bit so to visual output looks decent. Eventually when the story to update the cursor based on playback comes this will be dynamic.

PR #24, opened 2017-01-26
TBM-5045 - added automation for s3 deploy

troyblank:


Gruntfile.js
```diff
@@ -1,5 +1,10 @@
-module.exports = function(grunt) {
-    'use strict';
+module.exports = (grunt) => {
+    const keyPath = 'security/keys.json';
```
jhweaver: Can we name this file something else? It has the keys and the other S3 configs in it (region, bucket). Maybe something like "S3Config"?

troyblank: Yea totally agree. Just pushed up the change.

PR #87, opened 2017-03-20
TBM-5284 - Allow action pane to add a sentiment marker

troyblank:


components/interactionEditor/util.js
```diff
@@ -2,6 +2,17 @@ export function sortMarkersByTime(markers) {
     return markers.sort((a, b) => a.time - b.time).slice();
 }
 
+export function getAUniqueId(markers) {
```
troyblank: I wouldn't say this is a common way of doing it in React. I agree though this is a little brittle. I will revise using a more forever approach.

troyblank: This has been altered for the better; thanks for bringing this up.

PR #103, opened 2017-03-29
TBM-5338 - added bare bones auto complete input

troyblank:


assets/js/mockData/states.js
```diff
@@ -0,0 +1,52 @@
...
+    'New Hampshire',
+    'New Jersey',
+    'New Mexico',
+    'New York',
+    'North Carolina',
+    'North Dakota',
+    'Ohio',
+    'Oklahoma',
+    'Oregon',
+    'Pennsylvania Rhode Island',
```
jhweaver: Nope.

troyblank: daaamn! thanks lol

troyblank: @jhweaver fixed!

components/form/autoComplete.js
```diff
@@ -0,0 +1,103 @@
...
+    onEnterKey() {
+        this.selection = this.state.options[this.state.focus] ? this.state.options[this.state.focus] : '';
+    }
+
+    updateOptions(value) {
+        let regex;
+
+        this.state.options = [];
+        this.props.options.forEach((option) => {
+            regex = new RegExp(value, 'gi');
```
jhweaver: Are we sure we want to do these with regexs? We won't be able to search for some special characters. I'm not sure what kind of characters are normally in there, so I'm not sure if it's really a concern.

troyblank: Yea not fully sure currently; this iteration is very basic so I am fine with regexes for now.

PR #108, opened 2017-03-30
TBM-5346 - added ability to click on option

troyblank:


components/form/autoComplete/option.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+import classNames from 'classnames';
+import styles from './autoComplete.scss';
+
+export default class AutoComplete extends Component {
```
jhweaver: wut

troyblank: oops! fixed!

PR #111, opened 2017-03-31
TBM-5365 - added campaigns to the context bar

troyblank:


components/interactionEditor/contextBar.js
```diff
@@ -16,42 +17,45 @@ export default function ContextBar({
...
-            </button>
-          </li>
-          <li>
-            <button onClick={onClose} data-tip={'Cancel any interaction work.'}>
-              <IconClose />
-            </button>
-          </li>
-        </ul>
-      </header>
+      <div>
```
troyblank: This div is to shield the rest of contents from a flexbox in flexbox issue.

PR #125, opened 2017-04-11
TBM-5425 - added toggle component

troyblank:


assets/js/components.js
```diff
@@ -28,6 +28,7 @@ import {
     RadioGroup,
     Select,
     TextArea,
+    Toggle,
```
jhweaver: We should spell this "toogle" in half the places it's used to keep it consistent with RCI

PR #129, opened 2017-04-14
TBM-5442 - added expanding context component

troyblank:


components/interactionEditor/contextBar/expandingContext.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import classnames from 'classnames';
+import styles from './expandingContext.scss';
+
+export default function ExpandingContext({ label, items }) {
+    const isMoreThanOne = 1 < items.length;
+    let i = 1;
+    let dataTip = '';
+
+    for (i; i < items.length; i += 1) {
```
jhweaver: Won't this leave out the first element? Isn't this whole loop the same thing as a .join(', ')?

troyblank: hmmm yea it is except the first element needs to be left out. I could clone the array, remove the first element and do a join() and that woudl remove the loop which would be better. Will do in next PR

PR #155, opened 2017-04-28
TBM-5639 - converted current mock data to better match real data

troyblank:


assets/js/mockData/interactionData.js
```diff
@@ -0,0 +1,103 @@
...
+            id: 5,
+            type: 'topic',
+            time: 40000,
+            name: 'Some Topic'
+        },
+        {
+            id: 6,
+            type: 'sentiment',
+            time: 70000,
+            sentiment: 'positive',
```
jhweaver: We aren't going to have a special "sentiment" attribute on sentiment events. Is this a place holder?

components/interactionEditor/interactionEditor.js
```diff
@@ -187,9 +182,9 @@ export default class InteractionEditor extends Component {
               />
               <HorizontalTimeline
                 duration={duration}
-                audio={audio}
+                audio={resources.audioUrl}
```
jhweaver: While this makes total sense, this isn't what that value is right now. It's a link to the S3 file which is actually pretty much useless to you. I'll change the API sometime soon so that you get a link for the actual audio.

troyblank: Noted.

PR #156, opened 2017-04-28
TBM-5345 - stoped transcript search form select from generating errors

troyblank:


test/components/form/select.spec.js
```diff
@@ -20,7 +20,7 @@ describe('Select component', () => {
 
     it('should render', () => {
         const wrapper = shallow(
-          <Select id={id} name={name} label={label} options={options} onChange={externalOnChange} />
+          <Select id={id} name={name} label={label} options={options} />
```
jhweaver: gotcha

PR #236, opened 2017-06-27
TBM-5857 - matched mock data properties to data API

troyblank:


troyblank:
Need to change some things based on conversations with team.

PR #250, opened 2017-07-05
TBM-4925 - global note form added to interaction editor

troyblank:


components/interactionEditor/horizontalTimeline/horizontalTimeline.js
```diff
@@ -209,8 +210,8 @@ export default class HorizontalTimeline extends Component {
                       <IconInfo />
                     </button>
                   </li>
-                  <li data-tip={'Comments'}>
-                    <button>
+                  <li data-tip={'Call Notes'}>
```
jhweaver: Can we call this "Interaction Notes" instead?

troyblank: @jhweaver good catch; changed.

PR #304, opened 2017-08-03
TBM-6444 - added chat content to body of timeline node

troyblank:


test/components/icons.spec.js
```diff
@@ -242,3 +243,9 @@ describe('IconSuccess', () => {
         assertIcon(<IconSuccess color={'green'} />, 'success', 16);
     });
 });
+
+describe('IconChate', () => {
```
jhweaver: "Chate"? is that a typo?

troyblank: It is! fixed

test/components/icons.spec.js
```diff
@@ -242,3 +243,9 @@ describe('IconSuccess', () => {
         assertIcon(<IconSuccess color={'green'} />, 'success', 16);
     });
 });
+
+describe('IconChat', () => {
+    it('should render the chate icon', () => {
```
jhweaver: Another chate here yo

PR #481, opened 2018-01-09
TBM-6000 horizontalTimeline pauses and unpauses audio when spacebar is pressed

johncipponeri-topbox:
…ssed.


components/interactionEditor/horizontalTimeline/horizontalTimeline.js
```diff
@@ -111,6 +120,10 @@ export default class HorizontalTimeline extends PureComponent {
         }
     }
 
+    handleKeyDown(event) {
+        if ('p' === event.key) this.onPlayPause();
```
troyblank: We have talked a lot about this in the past and unfortunately you have not been included, but the space bar should be included as something that pauses/plays the timeline. I am fine with 'P' being something that also works, but we need to insure this works for uppercase as well if we want both P and spacebar.

troyblank: There needs to be a test for this.

components/interactionEditor/horizontalTimeline/horizontalTimeline.js
```diff
@@ -57,6 +62,10 @@ export default class HorizontalTimeline extends PureComponent {
         }
     }
 
+    componentWillUnmount() {
+        document.removeEventListener('keydown', this.handleKeyDown);
```
troyblank: There needs to be a test for this.

components/interactionEditor/horizontalTimeline/horizontalTimeline.js
troyblank: Let's use curly braces here; we can add a lint rule for it but no where do we use if statements without curlys throughout the code. If we end up using P and spacebar we should move to a `switch` statement anyway.

troyblank:
Also @johncipponeri-topbox  note that using a form field should not trigger this hotkey, currently filling out a form causing the audio to pause or play if you type 'p'.

components/interactionEditor/horizontalTimeline/horizontalTimeline.js
```diff
@@ -111,6 +116,16 @@ export default class HorizontalTimeline extends PureComponent {
         }
     }
 
+    handleKeyDown(event) {
+        const key = event.key;
+        const tag = event.target.tagName.toLowerCase();
+
+        if (' ' === key && 'input' !== tag && 'textarea' !== tag) {
```
troyblank: I think it would be better to use a formal `keyCode` here instead of the literal string ` `. This way we can get around encoding issues etc and make it less about the string comparison and more about the key that was pressed. What do you think?

troyblank: Instead of `'input' !== tag && 'textarea' !== tag` .. I think we can build this out as something that can be reusable, more functional and easier tested such as something like this.

`if (' ' === key && doesNotHaveInputFocus()) { ... } `

what do you think?

PR #489, opened 2018-01-18
TBM-7218 - Convert metadata datetime to local datetime

johncipponeri-topbox:


components/interactionEditor/interactionEditor.js
```diff
@@ -139,6 +140,13 @@ export default class InteractionEditor extends PureComponent {
     }
 
     componentDidMount() {
+        const { metadata } = this.props;
+
+        let startDate = metadata['Interaction Start Date'];
+        startDate = moment(startDate).format('MM-DD-YYYY h:mm A');
+
+        metadata['Interaction Start Date'] = startDate;
```
troyblank: This is done by the API, there is no need to calculate this on the frontend.

Also, not sure why we are manipulating a prop in this matter, but a prop can change anytime from the parent.

components/interactionEditor/interactionEditor.js
```diff
@@ -447,6 +450,22 @@ export default class InteractionEditor extends PureComponent {
         return deepEqual(this.untouchedValues, valuesToCheck);
     }
 
+    convertMetadataToLocalTime() {
```
troyblank: To me this does not make sense here, this is not where the data originates and is not where it is rendered. I would say it makes more sense down at the end of the road, so in the `MetaModal` right before it's rendered is less of a surprise. We can even use the render loop already there which makes things more findable/readable.

components/interactionEditor/interactionEditor.js
```diff
@@ -447,6 +450,22 @@ export default class InteractionEditor extends PureComponent {
...
+        const { metadata } = this.props;
+
+        if (metadata) {
+            Object.keys(metadata).map((fieldName) => {
+                const fieldValue = metadata[fieldName];
+
+                metadata[fieldName] = moment(fieldValue).isValid() ?
+                    moment(fieldValue).format('MM-DD-YYYY h:mm A') :
```
troyblank: This logic has been repeated in such places like [this](https://github.com/topbox-rci/rci-components/blob/c99b82faaeda8c4943ce2590e3d5e79c8bd13e09/components/interactionEditor/contextBar/contextBar.js#L93).

I feel to enforce this format/conversion of local dates, we should put all three of these guys in a place like [here](https://github.com/topbox-rci/rci-components/blob/c99b82faaeda8c4943ce2590e3d5e79c8bd13e09/components/interactionEditor/util.js) if not further up.

test/components/interactionEditor/interactionEditor.spec.js
```diff
@@ -1376,6 +1376,31 @@ describe('Interaction Editor', () => {
         assert.isFalse(instance.state.saveErrorShowing);
     });
 
+    it('should convert metadata to local time', () => {
```
troyblank: This fails for mr. travis and I > 

`  1) Interaction Editor should convert metadata to local time:
     AssertionError: expected false to be true
      at Function.assert.isTrue (node_modules/chai/lib/chai/interface/assert.js:332:31)
      at Context.<anonymous> (test/components/interactionEditor/interactionEditor.spec.js:1400:1`

troyblank: This is a time zone issue, just FYI, figured that out, it will only work in central.

PR #495, opened 2018-01-22
TBM-7508 - Removed convertToLocalTime and implementations

johncipponeri-topbox:


components/interactionEditor/util.js
```diff
@@ -381,7 +381,7 @@ export function filterOutNullJudgments(object) {
 export function convertToLocalTime(pDate) {
     let date = pDate;
 
-    date = moment(date).isValid() ?
+    date = moment(date).isValid() && isNaN(date) ?
```
troyblank: because `moment(NaN).isValid()` returns `false`. `&& isNaN(date)` appears to be superfluous here, do we need it?

troyblank: Ah I see, this is for metas so a number could been seen as seconds.

PR #505, opened 2018-02-21
Added auto fail styles

johncipponeri-topbox:


components/interactionEditor/actionPane/autoFailToggle.scss
```diff
@@ -0,0 +1,62 @@
+@import '../../../assets/sass/base/base';
+
+.autofail-toggle {    
+    margin-bottom: -20px;
+
+    background-color: #ececec;
+    border-radius: 3px;
+
+    max-height: 0;
+    transition: max-height $animation-time-subtle $animation-curve-subtle, padding $animation-time-subtle $animation-curve-subtle;
```
troyblank: `max-height` transition requires the element to be a specific height which is not great for responsiveness. There is no way to test this with this PR (no autoFail code to show it) so I could not test this but I assume this might fail at certain widths.

components/interactionEditor/actionPane/autoFailToggle.scss
```diff
@@ -0,0 +1,62 @@
...
+    overflow: hidden;
```
troyblank: As stated above this might not be great for responsiveness, will have to see when the code for the actual element is used alongside of this.

components/interactionEditor/actionPane/autoFailToggle.scss
```diff
@@ -0,0 +1,62 @@
...
+
+    &.auto-fail {
+        position: absolute;
```
troyblank: It's hard to tell without the elements but why is the positioned absolutely?

components/interactionEditor/actionPane/formPane/formPane.scss
```diff
@@ -176,3 +204,8 @@
         float: none;
     }
 }
+
+.autofail-prompt-backing {
+    margin-left: -10px;
+    background-image: repeating-linear-gradient(45deg, transparent, transparent 30px, $color-stripe-black 30px, $color-stripe-black 60px);
```
troyblank: I think we already have a style for this, we should combine the two in a mixin if we do.

troyblank:
@johncipponeri-topbox  it's hard to review this without actual elements, I don't know what these styles are for and if some are superfluous. I did whoever write possible issues that I saw but they were a little bit of a stab in the dark so take them with that in mind.

components/interactionEditor/actionPane/autoFailToggle.scss
troyblank: @johncipponeri-topbox  actually after this in action it seems like `max-height` would be fine. Given how small it can compress to width wise like the comment box.

PR #506, opened 2018-02-21
Added auto fail functionality

johncipponeri-topbox:


troyblank:
This guy won't build so I will just look over the code without actually testing it in a browser FYI.

ERROR in ./components/interactionEditor/actionPane/autoFailToggle.js
Module not found: Error: Can't resolve './autoFailToggle.scss' in 'rci-components/components/interactionEditor/actionPane'
 @ ./components/interactionEditor/actionPane/autoFailToggle.js 15:22-54
 @ ./components/interactionEditor/actionPane/formPane/question.js
 @ ./components/interactionEditor/actionPane/formPane/category.js
 @ ./components/interactionEditor/actionPane/formPane/formPane.js
 @ ./components/interactionEditor/actionPane/actionPane.js
 @ ./components/interactionEditor/interactionEditor.js
 @ ./components/index.js
 @ ./assets/js/logos.js
 @ ./assets/js/main.js
Warning:  Use --force to continue.

components/interactionEditor/actionPane/formPane/formPane.js
```diff
@@ -22,8 +22,16 @@ export default class FormPane extends Component {
     constructor(props) {
         super(props);
 
+        const { formNavigationIndexes, interactionForm } = this.props;
+        const activeForm = interactionForm[formNavigationIndexes.form];
+
+        this.state = {
+            autoFail: activeForm.isAutoFail
```
troyblank: To maintain this all in one place this should be part of the form prop further up > https://github.com/topbox-rci/rci-components/blob/master/components/interactionEditor/interactionEditor.js#L87.

The reason for this is the form can be effected by outside elements.

components/interactionEditor/actionPane/formPane/formPane.js
```diff
@@ -69,6 +77,14 @@ export default class FormPane extends Component {
         this.props.onInteractionFormChange(interactionForm);
     }
 
+    onAutoFailChange(autoFail) {
+        const interactionForm = this.props.interactionForm;
```
troyblank: We are getting two things from this props here. I feel like destructuring would be great here instead of the two this.props calls. What do you think?

components/interactionEditor/actionPane/formPane/formPane.js
```diff
@@ -96,6 +112,7 @@ export default class FormPane extends Component {
         this.updateNavigationBasedOnNa(value);
         updateCategoriesBasedOnNa(activeForm.categories, value);
 
+        this.onAutoFailChange(false);
```
troyblank: This seems like something that can be handled by state of the form itself. Why is this needed when changing forms?

components/interactionEditor/actionPane/formPane/formPane.js
```diff
@@ -126,14 +144,23 @@ export default class FormPane extends Component {
...
+                    <Notification small={true}>This form has been auto-failed</Notification>
+                  </div>
+                }
                 { form.isWeighted &&
                   <div className={styles.score}>
                     <span>Score</span>
                     <br />
-                    <span>{selectedWeight}/{maxWeight} <span>({weightPercentage}%)</span></span>
+                    { !autoFail && <span>{selectedWeight}/{maxWeight} <span>({weightPercentage}%)</span></span> }
+                    { autoFail && <span>0/{maxWeight} <span>0%</span></span> }
```
troyblank: This is something that can be handled by state, we should reflect the actual value and not fake something based on it's status. An autofail interaction should have a real score of 0.

components/interactionEditor/actionPane/formPane/question.js
```diff
@@ -16,22 +17,34 @@ export default class Question extends PureComponent {
...
+        const { question } = nextProps;
+
+        if (this.state.commentsShowing && !question.comment) {
             this.commentNeedsToBeRemoved = true;
         }
+
+        if (question.answers) {
+            const selectedAnswer = question.answers.filter(answer => answer.answer === question.selectedAnswerText)[0];
+
+            this.autoFailAnswerSelected = selectedAnswer ? selectedAnswer.canAutoFail : false;
```
troyblank: Should autoFail of the question be managed by the question and not props? (and communicated up when it is changed).

components/interactionEditor/actionPane/formPane/question.js
```diff
@@ -58,6 +71,34 @@ export default class Question extends PureComponent {
         this.props.onQuestionChange(question, this.props.index);
     }
 
+    onAutoFailChange(autoFail) {
+        const { question, index, onAutoFailChange, onQuestionChange } = this.props;
+
+        if (question.answers) {
+            question.answers.map((answer) => {
```
troyblank: I am not fully sure about this logic, why would we not want to change autofail based on the current question answer changed? This way we would not have to go through all answers to know about autofail.

assets/js/mockData/interactionForms.js
```diff
@@ -26,10 +26,10 @@ const interactionForms = [
                     {
                         question: 'Duis consectetur nec velit eu suscipit. Nunc id dui egestas, ultricies mi vitae, iaculis ligula. In rhoncus massa egestas nisi elementum lobortis. Integer ultrices neque ut ipsum ultricies.',
                         type: 'combobox',
-                        selectedAnswerText: 'no',
+                        selectedAnswerText: 'yes',
                         maxWeight: 2,
                         answers: [
-                            { answer: 'yes', weight: 2 },
+                            { answer: 'yes', weight: 2, canAutoFail: true, isAutoFail: true },
```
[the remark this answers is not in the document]
johncipponeri-topbox: @troyblank That is how the data is from the API so that is how I mocked it up. 😄

components/interactionEditor/actionPane/autoFailToggle.js
```diff
@@ -0,0 +1,44 @@
...
+              [styles['view-mode']]: viewMode,
+              [styles['auto-fail']]: autoFail,
+              [styles['toggle-showing']]: toggleShowing
+          }
+      )}
+    >
+      { !viewMode &&
+        <div>
+          <div className={styles.label}>
+            <p>AUTO FAIL</p>
```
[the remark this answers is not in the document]
johncipponeri-topbox: @troyblank Good catch.

components/interactionEditor/actionPane/formPane/formPane.js
johncipponeri-topbox: @troyblank I'll see what I can do. Wasn't sure how many things touched the form. Shouldn't be a big deal to yank the prop up a level.

components/interactionEditor/actionPane/formPane/formPane.js
johncipponeri-topbox: @troyblank Good catch. 👍

components/interactionEditor/actionPane/formPane/formPane.js
johncipponeri-topbox: @troyblank This is some code from early on, I'll look into it, you're right it's probably redundant. 😄

components/interactionEditor/actionPane/formPane/formPane.js
johncipponeri-topbox: @troyblank Thanks for clearing that up, it's a question I originally had whether we'd be retaining the non autofail score or not.

components/interactionEditor/actionPane/formPane/question.js
johncipponeri-topbox: @troyblank I'll see what I can do. You're right setting isAutoFail should be handled somewhere more like onAnswerChange.

assets/js/mockData/interactionForms.js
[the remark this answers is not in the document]
jhweaver: canAutoFail is right here. We have to know individually which answers are actually valid to auto fail. It's sort of like canNA, but at a different level.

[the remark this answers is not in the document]
troyblank: Ah ok I think i see so `canAutoFail` can be on question and answer then, as text questions will need something. I think I smell what you are steppin' in.

PR #508, opened 2018-02-23
TRW-231 - added form styles for auto fail

troyblank:


components/interactionEditor/actionPane/formPane/formPane.scss
```diff
@@ -176,3 +180,13 @@
         float: none;
     }
 }
+
+.auto-fail {
+    border-left: 14px solid $color-blue;
+    background-image: repeating-linear-gradient(45deg, transparent, transparent 30px, $color-stripe-grey 30px, $color-stripe-grey 60px);
+    background-size: 100% 9999px;
```
troyblank: This is to 'stabilize' the background so it does not shift during animation - I originally tried to put this as a var `$really-big-pixel-value` but realized this exact number was a bit of a sweet spot for this background so I kept it to a literal.

jsheebs: good call on the non px value

jsheebs:
Styling looks good to me.

PR #517, opened 2018-02-28
TRW-263 - swapped signin form state to change and not blur for return key

troyblank:


troyblank:
@scott-emery-topbox  correct, only enter key or button are ways of submitting this form.

PR #528, opened 2018-03-22
TRW-412 - added language toggle to interaction editor

troyblank:


troyblank:
@jsheebs Sorry I am not sure I am following you, but the mock data has 'korean' hardcoded so we can show something using the components. Please note the mock data is just that, data that is dummy and is hard coded only for presentation purposes. There reference for the language will be done at the timeline and actionPane levels.

PR #536, opened 2018-04-13
TRW-600 - show attempt countdown on signin form

troyblank:


troyblank:
@jsheebs  it does not, I prefer to be more specific so the prop is clear. Because my prop is "signInAttemptsLeft" does not mean the api has to be that as well, it can be "attempt". But one thing I advise though is if you didn't know anything about our signIn form would you assume attempt meant sign in attempts?

jsheebs:
That is not what I was trying to say. I was just saying that you will need to do the math on your end to determine the number of signin attempts left since this is now a patch request.

troyblank:
@jsheebs Ah gotcha yea that math can be done on the RCI side if need be, the components only care about logic that effect look and feel for the most part. Thanks!

PR #545, opened 2018-05-09
TRW-816 - do not show failed login attempt accounts if the count is off

troyblank:


components/auth/signIn.js
```diff
@@ -71,7 +71,7 @@ export default class SignIn extends PureComponent {
         const { errors } = this.state;
         const usernameError = get(errors, 'username');
         const passwordError = get(errors, 'password');
-        const showAttemptsCountDown = Boolean(3 >= signInAttemptsLeft && signInAttemptsLeft);
+        const showAttemptsCountDown = Boolean(3 >= signInAttemptsLeft && 0 < signInAttemptsLeft && signInAttemptsLeft);
```
bunnyLord: can we flip the order of variables and numbers we are comparing them to? the way it is now seems kind of backwards and hard to think about.
so like signInAttempts > 0 instead of 0 < singInAttempts

bunnyLord: also why do we need '&& signInAttemptsleft' at the end?

test/components/auth/signIn.spec.js
```diff
@@ -146,4 +146,17 @@ describe('Sign In', () => {
 
         assert.deepEqual(instance.state.errors, { username: usernameError });
     });
+
+    it('should not show failed attempt countdown if numbers are out of wack', () => {
+        const wrapper = shallow(
+          <SignIn
+            forgotPasswordLink={forgotPasswordLink}
+            pending={pending}
+            signInAttemptsLeft={0}
```
bunnyLord: What about a test for the less than or equal to 3 and whatever '&& signInAttemptsLeft' adds?

components/auth/signIn.js
troyblank: @bunnyLord per the condition flip; currently we are linting uses yoda, and there is good reason to do so. Basically you want to insure there is never an assignment by mistake. Read more here > https://eslint.org/docs/rules/yoda.

bunnyLord: Ohhhhh ok. That makes sense but I don't think we could have an assignment happen here with <= and <.
Why do we need &&signInAttemptsLeft at the end? is it making sure it has a value/not null?

test/components/auth/signIn.spec.js
troyblank: There is, it's above this one. https://github.com/topbox-rci/rci-components/blob/master/test/components/auth/signIn.spec.js#L25

Granted it's a render test,  you think these should be separated? I usually tie in a few display tests in with render and insure coverage is at 100%.

components/auth/signIn.js
troyblank: `'&& signInAttemptsleft'` is no longer needed as we are doing a range, so yea good call on that! Removed!

test/components/auth/signIn.spec.js
bunnyLord: No it's more like you test that it won't render when it's <= 0 but you don't test the other part of the conditional where it should also not render when it's more than 3.

troyblank: Ah, sorry I misunderstood! Yea good call I added a another test along with better values! Thanks for this.

bunnyLord: kind of surprised you have 100% coverage without that test actually? Coveralls didn't catch it for some reason?

test/components/auth/signIn.spec.js
```diff
@@ -146,4 +146,30 @@ describe('Sign In', () => {
...
+            pending={pending}
+            signInAttemptsLeft={chance.integer({ max: 0 })}
+          />
+        );
+        const notificationEle = wrapper.find(`.${styles.notification} span`);
+
+        assert.equal(notificationEle.length, 0);
+    });
+
+    it('should not show failed attempt countdown if numbers are out of wack', () => {
```
bunnyLord: out of wack is not very descriptive. 'if sign in attempts go above 3'?

components/auth/signIn.js
troyblank: Yea we definitely do not have an assignment, but the yoda practice is to help protect a mistake of assignment ever to happen holistically. Switching them will currently fail our lint process. I feel like the yoda rule is a good one and is adopted by our use of the AirBnb defaults, but it's up for debate and our lint rules can change if as a team we feel strongly it's not good practice.

test/components/auth/signIn.spec.js
troyblank: oops yea i mean to change that, copy and paste fail. updated!

PR #549, opened 2018-05-20
TRW-476 fixed comment not deleting when focused.

tVeratti:
## Comment Delete
`onMouseDown` has higher event precedence than both `onClick` and `onBlur`, so changing the delete button to `onMouseDown` makes its logic execute above `onBlur`. 

The risk in this approach is that the precedence of events may someday change in future DOM implementations. However, I'd like to see this component (and a few that connect with it) get a review on how we can changed to a controlled value for the textarea because there seems to be some patches to deal with loss of focus when the parent components update.

## Babel
I added [transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties/) as a plugin to the babel configuration. This allows the arrow syntax for self-binding component methods.


troyblank:
Love this, much more clean. We should remove all the binds.

troyblank:
Made a quick demo video > https://drive.google.com/file/d/1gG3fdhkF_aYxqo3XzFg9T9szUVi7ktEw/view?usp=sharing

troyblank:
Sorry it appears my local branch was jacked, it works!

PR #551, opened 2018-05-22
 TRW-904 - added universal search demo shell

troyblank:
This is just some pre-work before piecing together the universal search app demo. The goal is to next be able to add the welcome graphic and other components as they become available to insure they work well together. Everything in the demo directories are just for demonstration and because they will be implemented similar but different in RCI they have no tests.


views/index.hbs
```diff
@@ -5,24 +5,31 @@
             <h1>RCI Components</h1>
         </hgroup>
         <section>
+            <style>
```
troyblank: I realize this is slightly bad form, but the intention is not to make a build step for this page. This index just shows a simple navigation to the components on the root. That being said you can call me out on it :)

PR #552, opened 2018-05-22
TRW-904 - added welcome splash to universal search

troyblank:


assets/sass/base/silents/_typography.scss
```diff
@@ -1,6 +1,6 @@
 %header-secondary {
-    margin: 0 0 15px;
-    font-size: rem(20px);
+    margin: 0 0 1.6rem;
+    font-size: 2rem;
```
troyblank: SOOO currently this is too large to match design, this will work when I change the base font size tomorrow. Everything will have to be altered anyway so I just stuck with what it will be.

components/universalSearch/universalSearch.scss
```diff
@@ -0,0 +1,34 @@
...
+    margin-bottom: 10.6rem;
+}
+
+.universal-search__welcome-splash-header {
+    @extend %header-secondary;
+}
+
+.universal-search__welcome-splash-grapic {
+    margin-top: auto;
+    margin-bottom: -2.14rem;
```
tVeratti: Same here. Have -2.14rem written out is just another place we're setting ourselves up to maintain. A var for here and for the section with the padding using the same one will save us from potential defects of the mismatching in the future, and save us from the extra work of modifying in multiple places.

tVeratti: Alternatively, we could use positioning instead of having the flex box here.

```
.universal-search__welcome-splash-grapic {
  position: absolute:
  bottom: 0;
}
```

The section already has `position: relative`.

components/universalSearch/universalSearch.scss
```diff
@@ -0,0 +1,34 @@
+@import '../../assets/sass/base/base';
+
+.universal-search {
+    display: flex;
+    flex: 1;
+    flex-direction: column;
+    justify-content: center;
+}
+
+.universal-search__welcome-splash {
```
tVeratti: I think we could consider the splash as block level element. I'm not sure if that means we should create a component just for it, though that might not be a bad idea? Either way, renaming it like this might help
```
.welcome-splash
.welcome-splash__text
.welcome-splash__header
.welcome-splash__graphic
```
or
```
.splash
.splash__text
.splash__header
.splash__graphic
```

components/universalSearch/universalSearch.scss
```diff
@@ -0,0 +1,34 @@
...
+    display: flex;
+    flex: 1;
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+}
+
+.universal-search__welcome-splash-text {
+    margin-bottom: 10.6rem;
```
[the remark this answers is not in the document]
troyblank: The problem I see with this is design has to be on board first, if we make variables for spacing then our designs need to also only use that spacing. So until we have a talk with Casey I am not sure making variables will help anything. `10.6rem` is the amount of space between the text and the graphic when the graphic is no longer force to the bottom.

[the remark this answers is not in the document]
tVeratti: Design definitely should be on board. I'd see this as a long-term goal, something to tackle together over time.

components/universalSearch/universalSearch.scss
troyblank: The problem with `bottom: 0` is it will always be bottom zero. We need the graphic to expand downward as content grows if it exits, or for example on a smaller width screen. The solution I have come up with maintains normal padding of any body of content but allows the graphic to stick to the bottom ONLY if there is not enough content to fit the page. Thus the `-2.14rem` is to get past that padding and be at the bottom. Granted like you said that number could be in a var once we have a bigger discussion about making padding consistent in design.

components/universalSearch/universalSearch.scss
troyblank: Yea I went back and forth on that since it's all contained as a component. But that seems fine to me. I will make the splash a block in BEM. Code change incoming in a few.

components/universalSearch/universalSearch.scss
troyblank: For example: 
screen shot 2018-05-23 at 2 52 42 pm

components/universalSearch/universalSearch.scss
troyblank: @tVeratti cool thanks for this just updated the class names. Let me know if you see anything else.

components/universalSearch/universalSearch.scss
tVeratti: If the first relative parent is the overall section, it would grow with the child's content and therefore keep the graphic underneath it all still right?

troyblank: For example if you add more stuff in any part above the graphic including it's parents it stays to the bottom but does not give room for that stuff. You can see this behavior in your codepen example. Having something absolutely positioned ignores content in general. What I mean by expand downward is this graphic has to do a couple things: 1. it has to be anchored to the bottom of the window if the window's content does not have a scroll bar. (this will some what work with absolute positioning) 2. the graphic has to be at the bottom of the scrollable area if there is a scrollbar. (this is where absolute positioning fails).

PR #553, opened 2018-05-23
TPBP-730 - cookie policy notice on sign-in

tVeratti:
Added the notification the sign-in form. I also took the opportunity to add some BEM naming to the component styles, and to update the silent for floating actions to use `flex` instead of floating the elements.


tVeratti:
I accidentally branched off of TRW-904 so I'll make sure to update that before merging.

troyblank:
I will wait till TRW-904 is in so I can review this mo' betta.

PR #554, opened 2018-05-23
TRW-915 - converting all margin, padding, and font stuff to rems

troyblank:
There is some places margin is centering things that I did not convert but most of it's all rems now. I made the base 10px and the body `font-size: 1.4rem` so that all default text will not be so tiny. At some point we can bust our paddings out into some vars, but this is a step in making that easier at least.

I wanted to do this before moving forward with universal search as it would have been a pain in building it with rems.


assets/sass/base/silents/_containers.scss
```diff
@@ -67,7 +49,7 @@
         padding: 8px 10px;
         float: left;
         border-right: 1px solid $color-grey;
-        line-height: rem(17px);
+        line-height: 1.6rem;
```
tVeratti: Is it okay for these to be off 1px after the change?

assets/sass/demo/partials/_colors.scss
```diff
@@ -2,20 +2,20 @@
...
+        margin-bottom: 2rem;
     }
 
     li {
         width: 100%;
         height: 32px;
         border-bottom: 1px solid $color-black;
-        font-size: rem(14px);
+        font-size: 1em;
         list-style: none;
```
tVeratti: `em` here

tVeratti:
Is this giving the desired effect without setting `html { font-size: 62.5%; }`?

assets/sass/base/vars/_typography.scss
```diff
@@ -1 +1 @@
-$base-font-size: 14px;
+$base-font-size: 10px;
```
troyblank: @tVeratti so this should be the same as `{ font-size: 62.5%; }`.  Although I have never got why people use that, it assumes browsers will always keep `16px` as the default font size. I don't like to assume and thus I am trying to take matters into our own hands with the `10px`. That is unless there is something magic about that percentage that I am not understand.

assets/sass/base/silents/_containers.scss
troyblank: Nope it's not, ooop. Pushed up a fix.

assets/sass/demo/partials/_colors.scss
troyblank: Good catch, pushed up a change.

assets/sass/base/silents/_forms.scss
```diff
@@ -127,7 +127,7 @@
     left: 0;
     transition: top $animation-time-crisp $animation-curve-crisp, font-size $animation-time-crisp $animation-curve-crisp, color $animation-time-crisp $animation-curve-crisp;
     color: $color-grey-dark;
-    font-size: rem(14px);
+    font-size: 1.6rem;
```
tVeratti: `1.4rem`

assets/sass/base/silents/_lists.scss
```diff
@@ -14,31 +14,31 @@
...
     background: $color-grey-lighter;
 
     button {
         @extend %naked-button;
         flex-basis: 0;
         flex-grow: 1;
-        padding: 12px 20px 12px 10px;
-        font-size: rem(14px);
+        padding: 1.2rem 2rem 1.2rem 0;
+        font-size: 1.2rem;
```
tVeratti: 1.4rem

assets/sass/base/silents/_typography.scss
```diff
@@ -5,8 +5,8 @@
 }
 
 %header-tertiary {
-    margin: 0 0 15px;
-    font-size: rem(14px);
+    margin: 0 0 1.5rem;
+    font-size: 1.6rem;
```
tVeratti: 1.4rem

assets/sass/base/vars/_typography.scss
tVeratti: That was always the implementation I'd seen passed around. I'm fine with the 10px if it accomplishes the same thing.

assets/sass/demo/partials/_colors.scss
```diff
@@ -2,20 +2,20 @@
...
-        margin-bottom: 20px;
+        margin-bottom: 2rem;
     }
 
     li {
         width: 100%;
         height: 32px;
         border-bottom: 1px solid $color-black;
-        font-size: rem(14px);
+        font-size: 1rem;
```
tVeratti: 1.4rem

assets/sass/demo/partials/_senchaPage.scss
```diff
@@ -28,15 +30,15 @@
...
+    padding: 0 1rem;
     background-color: $color-orange;
     color: $color-white;
 
     li {
         height: 39px;
         padding: 13px 22px 0;
         float: left;
-        font-size: rem(13px);
+        font-size: 1.6rem;
```
tVeratti: 1.3

PR #560, opened 2018-05-24
TRW-479 -  added clone info to edit mode

tVeratti:


test/components/interactionEditor/contextBar/contextBar.spec.js
```diff
@@ -558,4 +558,27 @@ describe('Context Bar', () => {
           <li><strong>Cloned From:</strong> {'abcx'}</li>
         ));
     });
+
+    it('should render clonedFrom if provided in edit mode', () => {
+        const data = {
+            ...contextBarData,
+            clonedFrom: 'abcx'
```
troyblank: I have been using change.js to generate strings. So instead of `'abcx'` it would be a something like:

```
const cloneFrom = chance.word();

const data = { clonedFrom };

...

<li><strong>Cloned From:</strong> {cloneFrom}</li>
```

If you don't mind I think we should keep the generation of test data consistent and more random to help with testing in general.

tVeratti: Yes I love that - updated!

PR #561, opened 2018-05-24
TRW-902 - added universal search input styles

troyblank:
I tried to BEM-ify everything I touched. But this just gives us an auto complete input that does nothing right now. Mainly just wanted to match the styles of our design.

screen shot 2018-05-24 at 4 36 57 pm


components/form/input.scss
```diff
@@ -21,7 +21,52 @@
...
+        padding: .5rem 0;
+
+        & ~ .input__bar {
+            top: 39px;
+        }
+    }
+}
+
+.input--horizontal {
     & > div:first-child  {
```
tVeratti: I don't think the `div` part of the `:first-child` selector is necessary.

troyblank: Yea you are 100% right, i removed it along with 3 others, thanks.

PR #563, opened 2018-05-25
TRW-939 - fixed some minor issues with create/clone app typography

troyblank:


assets/sass/partials/_typography.scss
```diff
@@ -1,12 +1,12 @@
...
 
-html, button, input, optgroup, select, textarea {
-    color: $color-black;
-    font-family: 'Open Sans', sans-serif;
+html {
     font-size: $base-font-size;
 }
 
-body {
+body, button, input, optgroup, select, textarea {
```
troyblank: Moved this down a bit as browsers fit with these things. `html` still contains `10px` base and the reset inherit `1.4rem` so stuff like default buttons are not super tiny font wise.

tVeratti: Awesome, good move

PR #564, opened 2018-05-25
TRW-943 - hide notifications from print view

tVeratti:


assets/js/pages/interactionEditorWithNotification.js
```diff
@@ -0,0 +1,55 @@
...
+import InteractionData from '../mockData/interactionData';
+import InteractionEventSchemas from '../mockData/interactionEventSchemas';
+import styles from '../../sass/demo/style.scss';
+import { InteractionEditor, NotificationBar } from '../../../components/index';
+
+const { topBoxId, agentName, pKey, duration, startDate } = InteractionData;
+const contextBarData = { topBoxId, agentName, pKey, duration, startDate };
+
+let alert;
+setTimeout(() => {
```
troyblank: Why timeout at all? Just curious seems like it would be fine without it. Not a big deal for a demo though so I am not going to request a change.

tVeratti: Since the demo part isn't a full component, I would have preferred to hook into `componentDidMount` so that the alert ref could be set properly. Went with a timeout as a kinda janky way to let it mount before calling :)

troyblank: If you need more control of them, I have created demo components in the past just to do such a thing. https://github.com/topbox-rci/rci-components/tree/master/assets/js/demo/components

PR #565, opened 2018-05-25
TRW-940 - added search title to universal search bar

troyblank:
This adds the title to the search bar, aka the word search and the icon.

This does not cover two things yet to come. 1. an issue where the blue highlight does not get remove when using the clear input button. 2. the responsiveness of the input itself (to come soon).
Main goal was to get the icon in there and get it to a point it will work with with the dimensions picker.


assets/sass/base/vars/_spacing.scss
```diff
@@ -1,2 +1,4 @@
+$medium-tiny: .8rem;
+$medium-small: 1.6rem;
```
tVeratti: :) :)

components/form/autoComplete/autoComplete.js
```diff
@@ -70,24 +70,26 @@ export default class AutoComplete extends PureComponent {
...
         const key = e.key;
-        const lastFocus = this.state.options.length - 1;
-        let newFocus = this.state.focus;
+
+        const lastFocus = options.length - 1;
+        let focus = oldFocus;
 
         if ('ArrowDown' === key) {
-            newFocus += 1;
+            focus += 1;
```
tVeratti: `focus++` ?

components/form/autoComplete/autoComplete.js
```diff
@@ -121,17 +123,18 @@ export default class AutoComplete extends PureComponent {
     }
 
     updateOptions(value) {
```
tVeratti:
```
updateOptions(value) {
    const options = this.props.options.filter((option) => {
        const regex = new RegExp(value, 'gi');
        return regex.test(option);
    });

    this.setState({ options });
}
```

I will add this in a pr on its own if you don't want to deal with it here. Otherwise, it might be easier for you to add it and avoid another story + conflicts :) Just let me know how you prefer it.

components/form/autoComplete/autoComplete.js
troyblank: This is due to the airbnb eslint setting https://eslint.org/docs/rules/no-plusplus. It's argument is slightly weak so we can talk about removing if you like `++` better; I have no strong opinion.

components/form/autoComplete/autoComplete.js
troyblank: Another PR would be great to keep stuff small. Thanks!

PR #566, opened 2018-05-29
TRW-941 - hide modal backing when in print mode

tVeratti:
This is a bit of a guess.

I have not been able to reproduce the original issue, but I'm sure that if this were to be showing in print mode, it would result in a big gray background.

I will test with Josh to see if this resolves it when he goes to print.


tVeratti:
@jakehurst1 tested and he no longer sees the black box.

PR #567, opened 2018-05-30
TRW-931 - added dimension auto picker

troyblank:
This covers the adding of the auto picker input itself and the styles around it.

This has yet to cover the actual styles when you see the drop down of dimensions I made that another ticket as it will be slightly complex.

Note: currently the design has smaller fonts that this implementation, talking with Case about the 16px iOS zoom issue before we decided to make them smaller.

screen shot 2018-05-30 at 1 00 07 pm


troyblank:
I noticed the coverage dropped on this PR but locally it's fine, and the branch I am building on it's fine so I am going to assume it was a fluke at this poing.

PR #568, opened 2018-05-30
TRW-963 - added Paginator component and tests

tVeratti:


assets/js/demo/components/paginator.js
```diff
@@ -0,0 +1,19 @@
...
+
+    changePage = (activePage) => {
+        this.setState({ activePage });
+    };
+
+    render() {
+        const { activePage } = this.state;
+        return (
+          <div>
+            <Paginator totalRows={120} activePage={activePage} onPageChange={this.changePage} />
```
troyblank: Considering if we re-use pagination on say a bunch of logos in an admin. This would not indicate rows. Rows seems like we are assuming pagination will only be used for data tables. Should we consider making it something like 'totalItems" ?

components/paginator/paginator.js
```diff
@@ -0,0 +1,124 @@
...
+import { debounce } from 'lodash';
+import styles from './paginator.scss';
+
+export default class Paginator extends Component {
+    static defaultProps = {
+        activePage: 1,
+        totalRows: 0,
+        pageSize: 10,
+        pageRange: 10
+    };
```
troyblank: I need to use this more often.

components/paginator/paginator.js
```diff
@@ -0,0 +1,124 @@
...
+        );
+        const label = isNamed ? index : page;
+
+        const pageClassName = classNames(styles.paginator__page, {
+            [styles['paginator__page--active']]: isActive,
+            [styles['paginator__page--named']]: isNamed
+        });
+
+        if (!isActive) {
+            const onClick = this.onPageChange.bind(this, page);
```
troyblank: Why would we not bind this in the constructor?

components/paginator/paginator.js
```diff
@@ -0,0 +1,124 @@
...
+
+        // Return the active page without the button
+        // and click handler, preventing the same
+        // page from being re-rendered by accident.
+        return (
+          <li key={label}>
+            <span
+              className={pageClassName}
+              aria-label={`Current page, page ${page}`}
+              aria-current="true"
```
troyblank: yay! I need to bone up on aria roles.

assets/js/demo/components/paginator.js
```diff
@@ -0,0 +1,19 @@
+import React, { Component } from 'react';
+import { Paginator } from '../../../../components/index';
+
+export default class PaginatorDemo extends Component {
+    state = { activePage: 1 };
```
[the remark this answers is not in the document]
tVeratti: I'm a fan of property intializers when reference to props isn't needed. https://babeljs.io/blog/2015/06/07/react-on-es6-plus

If you feel strongly about keeping all of these in constructors I don't mind sticking to it, but I believe this is slightly newer syntax - which comes with the territory of still being a proposal. However, since it's a recommended use case even in official react docs, I feel okay about using a proposal.

assets/js/demo/components/paginator.js
tVeratti: Yeah I sat and thought about what to name it for a while. `items` seemed a bit generic so I swapped to `rows`. I half expected this to come up in the PR so I'm all good with changing it to `items` since it stood out to you.

components/paginator/paginator.js
tVeratti: Because it binds the page (number) for the click, which is context that isn't available in the constructor.

assets/js/demo/components/paginator.js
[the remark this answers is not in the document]
troyblank: Ah cool, yea I like them as well; let's just be consistent as we go forward we should lean towards these and not using constructor.

components/paginator/paginator.js
troyblank: Gotcha, I guess I was more asking why we are not binding once and sending the page onClick, but that would require function literals which are not that pretty either. This works.

tVeratti:
@troyblank updated based on comments :)

PR #569, opened 2018-05-31
 TRW-965 - styled dimension focus styles sans selections on universal search

troyblank:
This handles the styling of everything but the selection pane.

screen shot 2018-05-31 at 10 59 36 am


components/universalSearch/searchBar.scss
```diff
@@ -58,15 +65,48 @@
     }
 
     display: flex;
+    position: absolute;
     align-items: center;
+    width: 320px;
```
tVeratti: `32rem` here and line 38?

troyblank: The problems with `rem` here is that changing the size of the base font could have this move past the actual window causing horizontal scroll bars. Since we are using media queries in px I feel it makes sense to keep this `px`. I did think about adding it to a `var` but we have not done that yet with any specific css so I shy-ed away.

PR #575, opened 2018-06-01
TRW-980 - created Datepicker component, initial tests, and demo

tVeratti:


assets/js/navigation.js
```diff
@@ -21,6 +21,7 @@ renderHtml(
       <li><a href="#preloader">Preloader</a></li>
       <li><a href="#modal">Modal</a></li>
       <li><a href="#paginator">Paginator</a></li>
+      <li><a href="#datepicker_2">Datepicker v2</a></li>
```
troyblank: I assume this will replace our current date picker. I am fine with calling this v2 while we work on it but I think we should just call it "date picker" when it's done. What do you think?

components/datepicker/datepicker.js
```diff
@@ -0,0 +1,25 @@
+import React, { Component } from 'react';
+import classNames from 'classnames';
+import styles from './datepicker.scss';
+
+export default class Datepicker extends Component {
+    state = { open: false }
+
+    onToggle = () => {
```
troyblank: This is not covered well with unit tests.

test/components/datepicker/datepicker.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('Datepicker', () => {
+    // it('should render', () => {
+    //     const wrapper = shallow(<Datepicker />);
+    //     assert.isTrue(wrapper.contains(
+    //       <div className={styles.datepicker}>
+    //         <span className={styles.datepicker__toggle} />
+    //       </div>
+    //     ));
+    // });
```
troyblank: Was this being commented out a mistake? seems like we do need a test for rendering.

troyblank:
The commit "TRW-980 - fix linting errors" does not seem relevant, would you be against squashing it?

assets/js/navigation.js
tVeratti: I definitely will fully replace it. For now having both will allow some comparisons.

test/components/datepicker/datepicker.spec.js
tVeratti: It's intentional. I only meant to stub the file out and get it ready for tests. This is not the entirety of the tests I plan to write :)

troyblank: Ah, would you mind removing it for just this PR? It's almost like a todo comment. I am always afraid those will linger forever. I know this will not, but I guess I am also a clean freak.

components/datepicker/datepicker.js
tVeratti: Yeah I haven't written all tests yet, and this component is nearly empty for now. Tests soon™

PR #576, opened 2018-06-01
TRW-964 - adds dimension drop down

troyblank:
This adds dimension drop down styles / mock data.

This work forced me to rethink the styles of the search bar, so I improved them to be a little more fluid. I did have to add an explicit height, but it's better than using js to focus the hidden input. I tried to keep normal form flow in tact.

Also I noticed a bug with the `AutoComplete` clear button that needs to be fixed outside of this work.


components/universalSearch/universalSearch.js
```diff
@@ -3,10 +3,14 @@ import { GraphicRocket } from '../';
 import SearchBar from './searchBar';
 import styles from './universalSearch.scss';
 
-export default function UniversalSearch() {
+export default function UniversalSearch({ dimensions }) {
+    const dimensionsOptions = dimensions.map(dimension => dimension.name);
```
tVeratti: Do we not have any unique ids for these?

components/universalSearch/searchBar.scss
```diff
@@ -65,48 +87,71 @@
...
+    right: -1px;
+    left: -1px;
+    width: 304px; // for animation only
+    transition: width $animation-time-crisp $animation-curve-crisp;
+    opacity: 0;
+    z-index: 1;
+
+    input {
+        height: $input-explicit-height + 2px; // to overcome bottom border :/
+        padding-left: 114px;
```
tVeratti: What's the importance of `114px`? Could it be a variable even just for this file? And rem?

components/universalSearch/searchBar.scss
```diff
@@ -65,48 +87,71 @@
...
+    top: -1px;
+    right: -1px;
+    left: -1px;
+    width: 304px; // for animation only
+    transition: width $animation-time-crisp $animation-curve-crisp;
+    opacity: 0;
+    z-index: 1;
+
+    input {
+        height: $input-explicit-height + 2px; // to overcome bottom border :/
```
tVeratti: Just a thought, would `calc(100% + 2px)` do the job?

components/universalSearch/searchBar.scss
```diff
@@ -19,30 +19,52 @@
...
+    display: flex;
 }
 
 .search-bar__title {
-    padding: 0 $small-spacing 0 $tiny-spacing;
+    display: flex;
+    position: relative;
+    align-items: center;
+    justify-content: space-between;
+    width: 86px;
```
tVeratti: This file has a lot of hard-coded px widths. Would it help to define them at the top of the file so that we can start to see any kind of mathematical relationships? It would also help maintenance and future work to be able to modify variables at the top instead of searching throughout the css to find where it has been used or referenced.

components/universalSearch/searchBar.scss
```diff
@@ -65,48 +87,71 @@
...
     }
 
-    .dimension-picker {
-        @extend %stand-out-shadow;
+    .search-bar__input {
+        opacity: $disabled-opacity;
+    }
+
+    .dimension-picker__button {
+        opacity: 0;
```
tVeratti: This won't stop the button from being focusable. Adding `display: none` or `visibility: hidden` would fix this.

troyblank: At, good catch this class doesn't exist anymore actually since I switched attempts a few times. Now I  do not have a `<button/>` in this component to help with this issue. Going to remove this css rule.

components/universalSearch/searchBar.scss
troyblank: Yea makes sense, I had the thought of adding vars at the top but since it hasn't been done before I was just making explicits. Next PR you should see something along the lines of.

```$search-bar-title-width: 86px;
$dimension-picker-input-space-for-title: 114px;
$dimension-picker-input-expand-animation-start-width: 304px;
$dimension-picker-input-expand-animation-end-width: 540px;```

components/universalSearch/searchBar.scss
troyblank: Unfortunately it will not. The way this area has to support `position: absolute;` with the dimension auto complete field means we need a pixel value here.

components/universalSearch/universalSearch.js
troyblank: I am sorry I do not follow. The purpose of this currently is to show the drop down of dimensions visually. Currently there can be duplicates and the data structure is in flux, thus I am keeping it simple and just making an array of dimension objects with `name` and `type`. This might change in the future.

tVeratti: If you're going to render a bunch of options, what will you use for the key? (index?)

Also, when they select it, if there are two options called `Doug` that might be from slightly different dimensions, do we have any way of detecting that if we're only passing `Doug`?

I was just thinking that maybe have unique ids will be a good safety net for duplicates that may not actually be the same thing.

tVeratti:
Honestly nice job for rewriting this when it seemed the original approach wasn't going to work. Not always the easiest thing to start over, but I think it hardens the approach too :) This is a tough one, nice work.

components/universalSearch/universalSearch.js
troyblank: Ah I gotcha, currently key is the index > https://github.com/topbox-rci/rci-components/blob/master/components/form/autoComplete/autoComplete.js#L221

It appears there would be no conflict with that currently but it might not be the best way of doing things. We can change but it's out of scope for this as it's handled with the `<AutoComplete />` component.

PR #578, opened 2018-06-05
TRW-981 - created Calendar component (sub component of Datepicker)

tVeratti:
This will be used (x2) by the overall `Datepicker` component to render all dates and the selection ranges, along with providing callback handlers for users to click and select date ranges, go to previous month, and go to next month.

The current demo only renders one. The next part of the work will be consuming the `Calendar` within `Datepicker` and managing pre-made selections.

Todos for this component include: Adding the arrow `SVG` icons (placeholders atm), refining some of the sizing styles (height/width of dates), and adding *A11y* attributes.

I wanted to get the core part of the functionality/styles in for review before it gets any bigger as a review.


components/datepicker/calendar.js
```diff
@@ -0,0 +1,189 @@
...
+        const { first } = this.range;
+
+        // `first` represents the first possible
+        // day of the month. Anything before that
+        // is from the previous month.
+        const offset = (first - index) * -1;
+        return new Date(year, month, offset + 1);
+    };
+
+    checkWithinRange = (date) => {
```
tVeratti: This function felt a bit meh. Any ideas on improving the efficiency are welcome. Perhaps this is just a cost of calculating within a range for each date.

components/datepicker/calendar.js
```diff
@@ -0,0 +1,189 @@
...
+
+        return (
+          <div className={styles.calendar__header}>
+            {/* Month Name & Controls */}
+            <div className={styles.calendar__controls}>
+              <button onClick={clickPrev}>p</button>
+              <span className={styles.calendar__name}>
+                {this.months[month]}
+                &nbsp;
+                {`${year}`.slice(-2)}
```
tVeratti: This should be safe until we hit another decimal in years. I don't think this code will be running in year 10000 so... :D

components/datepicker/calendar.js
```diff
@@ -0,0 +1,189 @@
...
+        // Get the first day of the month.
+        // This is used to understand what day
+        // of the week the month begins on.
+        const firstDay = new Date(year, month, 1);
+
+        // Get the last day of the month.
+        // Go up one month, then pass 0 as the date
+        // and it will return the last day of the
+        // the previous month (28, 31, etc).
+        const lastDay = new Date(year, month + 1, 0);
```
tVeratti: https://stackoverflow.com/a/222439

components/datepicker/calendar.js
```diff
@@ -0,0 +1,189 @@
...
+            newMonth = 11;
+        }
+        if (onChangeMonth) onChangeMonth(newMonth, year);
+    }
+
+    getDate = (index) => {
+        const { year, month } = this.props;
+        const { first } = this.range;
+
+        // `first` represents the first possible
```
troyblank: so instead of explaining what `first` means here, why not name the prop firstNameOfTheMonth? That way when handling the prop there is no question what it is.

components/datepicker/calendar.js
troyblank: The first thing I question is why are we doing string compares instead of just casting everything to numbers and then running the `x > y && x < z`? There might be something I am missing but I don't see the value of the string compares. Also I think the previous is good but you could also get cute and do something like `(x-y) * (x-z) < 0` but honestly I think the  `x > y && x < z` format is more readable.

components/datepicker/calendar.js
```diff
@@ -0,0 +1,189 @@
...
+        return (date >= selection[0] && date <= selection[1]) || (
+            date.toString() === selection[0].toString()
+            || date.toString() === selection[1].toString()
+        );
+    };
+
+    calculateRange = (props) => {
+        const { year, month } = props;
+
+        // Get the first day of the month.
```
troyblank: Same here for me, I love funcs and vars that are descriptive. Compilers make things small, so why not name this `firstDayOfTheMonth` so someone can quickly understand what the var represents.

components/datepicker/calendar.js
troyblank: Creating another y2k huh? I am all about project mayhem.

components/datepicker/calendar.scss
```diff
@@ -0,0 +1,78 @@
...
+        background-color: $color-blue;
+        color: $color-white;
+
+        &:first-child {
+            border-radius: $selection-radius 0 0 $selection-radius;
+        }
+
+        // There is no selector for previous sibling,
+        // so this class needed to be added by JavaScript.
+        &.calendar__day--last,
```
troyblank: I have often used css classes such as `js-some-class`. When indicating css that is manipulated or added by js. What do you think? This might help let us know which css is battered around by the javascripts.

components/datepicker/datepicker.js
```diff
@@ -2,23 +2,67 @@ import React, { Component } from 'react';
...
+    onChangeMonth = (month, year) => {
+        this.setState({ month, year });
+    }
+
+    onSelectDate = (date) => {
+        let range = [...this.state.selectionRange];
+
+        const dateString = date.toString();
+        if (dateString === range[0].toString() || dateString === range[1].toString()) {
+            // The selection is on the beginning or ending node,
```
troyblank: As above I might suggest making a variable that makes the code more readable if you think it makes sense such as

```
const isOnBeginningOrEnd = dateString === range[0].toString() || dateString === range[1].toString();

if (isOnBeginningOrEnd) {
    .....

File: components/datepicker/calendar.scss [hunk 8]
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+.calendar__name {
+    flex-grow: 1;
+    text-align: center;
+}
+
+.calendar__week {
+    display: flex;
```
troyblank: This generally seems like a great time to use `display: grid;`. Without trying it I feel like it would be easier and less verbose than flex. But that being said you might have already thought, tried and moved on, what do you think?

components/datepicker/calendar.js
tVeratti: I suppose I lean towards brevity in a lot of my variable naming. I could see the benefit from being more verbose sometimes to explain the meaning of a variable without commenting. But perhaps a compromise, as for some reason the `firstNameOfTheMonth` just feels like it would be a pain to read within further calculations. Granted this  function is short, but in another scenario where it gets used more, it'd be dizzying.

components/datepicker/calendar.js
tVeratti: Those are fine until the date is `==` or `===`.
`new Date() === new Date()` is `false`, so the `toString` is specifically for matching on the first/last date.

components/datepicker/calendar.js
tVeratti: I could rename it to `firstDayOfTheMonth` but does that truly provide more information than `firstDay`? This is probably a naming thing we're gonna have to practice agreeing on since we have slightly different styles, which I am totally fine with. On this particular one, I believe the comment explains the future purpose of the variable which may not be conveyed even by the full name. I do tend to over-explain in comments so let me know if you still think it could just be replaced by a more verbose name.

components/datepicker/calendar.scss
tVeratti: I think grid is a bit overkill for this. All we need are evenly sized boxes, each on a non-wrapping row. If we had to create unique sizes and customize specific columns sizes/positions, I might lean towards grid.

components/datepicker/calendar.js
troyblank: What about `date.date.getTime() >= y && date.getTime() <= z` ?

tVeratti: 10/10 I love it, will update

components/datepicker/calendar.js
troyblank: I guess I i see `firstDayOfTheMonth` telling you that the var is the first day of the month and `firstDay` only tells us we have a day and it is first. The comment might not be necessary with a more verbose var name. My thought here is that it takes less time to read a variable that explains what it is than a variable that needs a comment of what it is.

components/datepicker/calendar.scss
troyblank: Coolness.

components/datepicker/calendar.scss
troyblank: Yea you are right all css is being spit out by JS technically, but I am just referring to things that are manipulated by OUR js, so it's easier to inventory and point out a style that is "special" in the way that we are actually messing with it in real time.

PR #579, opened 2018-06-05
TRW-993 - updated auto complete options pane height based on viewport

troyblank:


components/form/autoComplete/autoComplete.scss
```diff
@@ -14,7 +14,7 @@
     @extend %shadow;
     position: absolute;
     width: 100%;
-    max-height: 192px;
+    max-height: 500px;
```
troyblank: I struggled in making a var for this because it is arbitrary based on design and the var would just be like `option-pane-max-height`. But really it's just to insure the pane does not get super big and over whelming. But feel free to ask me to make it that var, it just seemed silly when I tried.

components/form/autoComplete/autoComplete.js
```diff
@@ -112,25 +121,34 @@ export default class AutoComplete extends PureComponent {
...
+        const { optionPaneWindowPadding } = this.props;
+        const windowHeight = window.innerHeight;
+        const position = this.autoComplete.getBoundingClientRect().top;
+        const spaceToExpand = bottomOfWindow ? position : (windowHeight - position);
 
-            this.select.scrollTop = newOffSetTop;
+        this.setState({ optionsPaneHeight: (spaceToExpand - optionPaneWindowPadding) });
+    }
+
+    scrollSelectionIntoView() {
```
tVeratti: How does the user navigate this with the keyboard? I don't see a key press handler that would accept arrow keys for going up/down options or anything - what am I missing?

components/form/autoComplete/autoComplete.js
```diff
@@ -7,6 +7,10 @@ import styles from './autoComplete.scss';
 import { isAtBottomOfWindow } from '../../util/dom';
 
 export default class AutoComplete extends PureComponent {
```
tVeratti: With the DOM manipulation and height calculations, I don't think this can really be considered a `PureComponent` anymore, might be best to swap over to a regular `Component`

components/form/autoComplete/autoComplete.js
```diff
@@ -112,25 +121,34 @@ export default class AutoComplete extends PureComponent {
...
-            if (activePos >= viewRectEnd) {
-                newOffSetTop = (activePos - selectHeight) + activeLiHeight;
-            } else if (activePos <= viewRectStart) {
-                newOffSetTop = activePos;
-            }
+    setOptionPaneHeight() {
+        const { bottomOfWindow } = this.state;
+        const { optionPaneWindowPadding } = this.props;
+        const windowHeight = window.innerHeight;
+        const position = this.autoComplete.getBoundingClientRect().top;
```
tVeratti: It might be stylistic discussion, but I have generally been more comfortable with letting refs be refs without setting them to a `this` property. The thing that provides is an understanding of what it is. At this point, I only know that `this.autoComplete` is a ref if I look at the `render` function. Perhaps you feel that's sufficient, but I'll present the alternative.

If we set refs the classic way `ref="autoComplete"` we can get it elsewhere with `const { autoComplete } = this.refs;`. Both ways are fine, it just made me double check here so figured I'd mention it.

components/form/autoComplete/autoComplete.js
```diff
@@ -112,25 +121,34 @@ export default class AutoComplete extends PureComponent {
...
+        const spaceToExpand = bottomOfWindow ? position : (windowHeight - position);
 
-            this.select.scrollTop = newOffSetTop;
+        this.setState({ optionsPaneHeight: (spaceToExpand - optionPaneWindowPadding) });
+    }
+
+    scrollSelectionIntoView() {
+        // This allows a selection made by keyboard or otherise to always be scrolled into view.
+        const activeLi = this.select.querySelector(`.${styles.active}`);
+        const activeLiHeight = get(activeLi, 'offsetHeight', 0);
```
tVeratti: What are the benefits of `get` aside from preventing mutation? If you're getting the raw value it won't mutate the original anyway, so you'd be safe with `const { offsetHeight = 0, offsetTop = 0 } = activeLi;`

I recognize that those names probably aren't descriptive enough, but mainly I'd like to understand if we need `get` everywhere as it seems to be a layer of abstraction that might not be needed.

test/components/form/autoComplete/autoComplete.spec.js
```diff
@@ -545,4 +553,55 @@ describe('Auto Complete', () => {
...
+            allowAnyValue={true}
+            options={options}
+            onUpdate={onUpdate}
+            optionPaneWindowPadding={optionPaneWindowPadding}
+          />
+        );
+        const instance = wrapper.instance();
+
+        window.innerHeight = windowHeight;
+        instance.autoComplete = { getBoundingClientRect: () => ({ top: position }) };
```
tVeratti: Nice way to handle mocking that internal call. 👍

components/form/autoComplete/autoComplete.scss
tVeratti: How about just making it rem? :)

tVeratti: Or even alternatively, since this is a matter of going off-screen, a unit like `50vh`

components/form/autoComplete/autoComplete.js
troyblank: It was made a longtime ago https://github.com/topbox-rci/rci-components/blob/master/components/form/autoComplete/autoComplete.js#L83

tVeratti: Oh damn, yeah I realize now it was collapsed in the diff, yikes sorry about that ;)

components/form/autoComplete/autoComplete.js
troyblank: I put the height in `state` intentionally for this reason. This will behave just fine if the height does not change and there is not a re-render. I did test this. So I think `PureComponent` still works here unless I misunderstand what you are saying.

components/form/autoComplete/autoComplete.js
troyblank: Actually the reason `get` is used here is because not all browsers have `offsetHeight` (meaning virtual browsers). So it's not about mutation but more about assuring it returns something rather than throws x of undefined type errors.

`const { offsetHeight = 0, offsetTop = 0 } = activeLi` seems awesome to me, I will try to add it, but for some reason I didn't think dom elements could be deconstructed like that. But I am happy to find out :)

components/form/autoComplete/autoComplete.js
troyblank: Actually this is discourage by react the project https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-string-refs.md. Sorry Charlie.

components/form/autoComplete/autoComplete.js
tVeratti: Ok nevermind on this comment, it being in state overs my initial concern.

PR #581, opened 2018-06-11
TRW-1143 - added caret icon with four directions, and added to calendar

tVeratti:


components/datepicker/calendar.scss
```diff
@@ -29,12 +29,16 @@ $selection-radius: 50%;
         @extend %naked-button;
         padding: $tiny-spacing;
         cursor: pointer;
```
troyblank: per sass lint there is a line expected here between blocks

components/datepicker/calendar.scss
```diff
@@ -29,12 +29,16 @@ $selection-radius: 50%;
...
+        svg {
+            vertical-align: middle;
+        }
     }
 }
 
 .calendar__name {
     flex-grow: 1;
     text-align: center;
```
troyblank: per sass lint `text-align` should come last

components/icons/caret.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { greyDark } from '../util/colors';
+
+export default function Caret({ direction = 'down' }) {
```
troyblank: Sorry should have fond this before but there is no test for this icon, please see https://github.com/topbox-rci/rci-components/blob/master/test/components/icons.spec.js to see how we have been doing icon tests in the past. They are basic sanity checks but are valuable to have in case some icon gets mangled on accident.

PR #582, opened 2018-06-11
TRW-993 - seperated select drop down from auto complete

troyblank:
This PR is to ONLY separate two components, the select drop down and auto complete. There is probably tons of refactoring, accessibility we probably need to do but because it was already a lot of work I did not focus on changing anything that didn't directly help make the drop down options be it's own component.


components/selectDropDown/selectDropDown.js
```diff
@@ -0,0 +1,142 @@
...
+            style.bottom = height + optionsPaneBottomUpPadding;
+        }
+
+        return (
+          <ul
+            className={classNames(styles['select-drop-down'], {
+                [styles['select-drop-down--showing']]: showing
+            })}
+            style={style}
+            onKeyDown={this.onKeyDown}
```
tVeratti: My previous understanding has been that an element like `ul` doesn't really deal with focus/onKey events like an `input` would. This is totally working as expected?

tVeratti: I looked down into `<Option />` and didn't any input that might've bubbled the event up. How is this working? :D

test/components/selectDropDown/selectDropDown.spec.js
```diff
@@ -0,0 +1,263 @@
...
+            options={options}
+            showing={showing}
+            parentLayoutInformation={layoutInformation}
+            onItemSelect={onItemSelect}
+          />
+        );
+
+        wrapper.setProps({ itemFocus: chance.natural() });
+
+        assert.isTrue(componentDidUpdate.called);
```
tVeratti: Too bad we can't test scrollTop and stuff, that'll be something we need to be careful if we change the logic in the future.

components/selectDropDown/selectDropDown.js
troyblank: This is working just based on keybindings, I am find changing this but if possible I would love to defer to another PR as the focus of this one was separation. Also the problem with an `<input>` that contains other `<inputs>` is that focus of autocomplete would not work correctly.

components/selectDropDown/selectDropDown.js
```diff
@@ -0,0 +1,142 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            itemFocus: 0
+        };
+    }
+
+    componentWillReceiveProps(nextProps) {
+        const { showing: oldShowing, options: oldOptions } = this.props;
```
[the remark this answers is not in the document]
troyblank: Yea I actually wasn't fully sure if it was the best way to compare but I didn't see anything better out there in the wild :)

components/selectDropDown/selectDropDown.js
troyblank: I will need to go in and add the "popular" titles to these 'options' at some point. That might be a good time to consider making these radios or something more accessible.

tVeratti: Honestly I'm doubting the use of radios for this based on your research and the results of the qlik extension. I am mainly curious how it's working without a focusable thing inside, because that is useful info :D I think this is great for now.

PR #583, opened 2018-06-12
TRW-942 -  removed display: flex from main container to fix page-break

tVeratti:
I honestly do not have a good explanation for this.

`page-break: avoid` was not being respected on interactions under view mode for some clients. In addition, when I added `page-break-after: always` it would start a new page after each question, but it would still pull the top ~20px of the next page's question onto the first page. I wasn't able to find anything similar through stack overflow or on any browser bugs. We have a -lot- of layers of `display: flex` so all I can theorize is that there is some kind of browser failure handling things in this scenario.

By removing `display: flex` from `.main-container` the page-break property was being completely respected. It didn't seem to affect the overall layout. That's not too surprising since it only had flex and `flex-direction: column`.


tVeratti:
Nope that did affect the scrolling hook for the timeline. Per @troyblank 's suggestion, I moved the override to the print media query.

PR #585, opened 2018-06-12
TRW-1016 - added select button

troyblank:
This work uses the separated `<SelectDropDown />` and ties it to a button. Used area role `listbox` to help with accessibility.

One thing I did not follow on design is that the options are currently as wide as the bottom. This is something I feel like we can tweak later if needed, but would love to get design feedback before making it work differently than the `<AutoComplete />` component.


components/selectDropDown/selectDropDown.js
```diff
@@ -0,0 +1,169 @@
...
+        const style = { maxHeight: optionsPaneHeight };
+
+        if (!isAtBottomOfWindow) {
+            style.top = height;
+        } else {
+            style.bottom = height + optionsPaneBottomUpPadding;
+        }
+
+        return (
+          <ul
```
tVeratti: I think `role="listbox"` belongs on the `<ul>` element. The `aria-activedescendant` can stay higher up I believe. Then we can also add `aria-expanded={showing}` to the `<ul>`

[Source: W3](https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology)

components/selectButton/selectButton.js
```diff
@@ -0,0 +1,94 @@
...
+        const selectedOption = options[selectedIndex];
+
+        return (
+          <div
+            className={styles['select-button']}
+            role={'listbox'}
+            aria-activedescendant={`${ariaID}-${itemFocus}`}
+          >
+            <Button text={selectedOption} onClick={this.onClick} />
+            <button
```
tVeratti: This guy could use `aria-haspopup="listbox"` :)

[Source: W3](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html)

tVeratti: This part seems a bit different based on the examples provided. That link shows that it would also recommend adding `aria-expanded="true"` to the toggle button, but the other example has it on the `<ul>` itself. As long as we have it on one of them we're probably okay.

components/selectDropDown/option.js
```diff
@@ -10,15 +10,19 @@ export default class Option extends PureComponent {
...
           <li
-            className={classNames({
-                [styles.active]: index === focus
+            className={classNames(styles.option, {
+                'js-is-active': isActive,
+                [styles['option--active']]: isActive
             })}
             onMouseDown={this.itemSelectHandler}
             key={index}
+            role={'option'}
```
tVeratti: We can also add `aria-selected` if this is the currently selected option. That would require an extra check and prop though :(

components/selectDropDown/selectDropDown.js
```diff
@@ -0,0 +1,169 @@
...
+    onKeyDown = (e) => {
+        const { itemFocus: oldItemFocus } = this.state;
+        const { options, onItemSelect } = this.props;
+        const key = e.key;
+        const lastItemFocus = options.length - 1;
+        let itemFocus = oldItemFocus;
+
+        e.preventDefault();
+
+        if ('ArrowDown' === key) {
```
tVeratti: `switch` might be a slightly better option here - though I wouldn't really worry about it unless we planned to add more possible keys (tab, space, etc.)

troyblank: The problem with that is the `'Enter' === key && onItemSelect` at the end, so it would be awkward in my eye. I agree though switch usually better for anything above 2 if than else checks, but I am going to leave this.

components/selectDropDown/option.js
troyblank: Great idea, thanks, just added.

components/selectButton/selectButton.js
troyblank: Great idea, thanks, just added aria-haspopup.

components/selectDropDown/selectDropDown.js
troyblank: Great idea, thanks, just added.

troyblank:
@tVeratti  added some of the stuff you suggested, thanks a lot for that. I wish git would kick the PR to not approved after someone updates the code. Do you mind giving it a quick once over after those adds? Thanks!

components/selectDropDown/option.js
tVeratti: I was under the impression that `isActive` represents focus and not actual selection. It should match up to `activedescendant`

troyblank:
@tVeratti OK cool I was confused about `aria-selected` there for a second. But all should be good now I hope. Thanks!

PR #586, opened 2018-06-13
TRW-1161 - change the style of query nodes

troyblank:
Updated query node's style > currently used by transcript search . These will be used on universal search and I made them use 'dimension' and 'value' props that work very well with transcript search.

Also not the move from icon sizes by indicating the 'small-legacy' as the old size.


troyblank:
Coverage report must by lying, it's 100% locally 🤷‍♂️

PR #588, opened 2018-06-14
TRW-1018 - allowed universal search to add query nodes

troyblank:
The goal of this PR is to add the basics of the query adding flow in universal search without actually searching anything.

This work adds the add to search button to universal search as well as adds query nodes when it is clicked. There is some work that needs to be done that this does not cover such as removing queries, omitting queries, and clearing out the inputs when a query is made.


test/components/universalSearch/searchBar.spec.js
```diff
@@ -94,8 +104,38 @@ describe('Universal Search - Search Bar', () => {
...
+            dimensionsOptions={dimensionsOptions}
+          />
+        );
+        const queries = [{
+            id: chance.natural(),
+            dimension: chance.word(),
+            value: chance.word()
+        }];
+
+        wrapper.setState({ queries });
```
tVeratti: Would it be helpful to provide a prop or function that can set the queries in state? I can see that being useful for loading saved query sets, or preloaded sets?

components/universalSearch/searchBar.js
```diff
@@ -27,46 +39,77 @@ export default class SearchBar extends PureComponent {
...
+        this.setState({ queries, queryIdCount: nextID });
+    }
+
     render() {
-        const { dimensionSelected, dimensionInputFocused } = this.state;
+        const { dimension, dimensionInputFocused, queries } = this.state;
         const { dimensionsOptions } = this.props;
-        const dimensionPickerButtonText = dimensionSelected || 'Choose a dimension...';
+        const dimensionPickerButtonText = dimension || 'Choose a dimension...';
+        const hasQueries = Boolean(queries.length);
```
tVeratti: Alternatively, `!!queries.length`

components/universalSearch/searchBar.js
```diff
@@ -27,46 +39,77 @@ export default class SearchBar extends PureComponent {
         this.setState({ dimensionInputFocused: false });
     }
 
+    onSearchQueryAppend = () => {
+        const { dimension, value, queryIdCount, queries: oldQueries } = this.state;
+        const queries = cloneDeep(oldQueries);
```
tVeratti: I understand why you use `cloneDeep` here, but I still think it's overkill. As far as I can tell, there is no code that compares or mutates the queries themselves, we are really only concerned with modifying the array once things have been added. If it were me, I may have written it like:
```
const queries = [
    ...oldQueries,
    {
        id: queryIdCount,
       // etc...
    }
];
```

test/components/universalSearch/searchBar.spec.js
troyblank: Yea eventually this will have to be shared with the parent to make the query so it will so have to happen at some point.

components/universalSearch/searchBar.js
troyblank: ooo, yea I like that better. Good call, pushed up changed, you mind re-approving?

PR #589, opened 2018-06-14
TRW-1073 - classification form renders AutoComplete for large choice sets

tVeratti:
Initially I was modifying the behavior in the marker pane, but I believe after some testing that was a level above. It's possible, though, that the design wants these changes on both layers. I'm going to start at this level and we can add these in any other place that gets specifically pointed out in other stories.


components/interactionEditor/actionPane/classificationForm.js
```diff
@@ -18,11 +40,11 @@ export default class ClassificationForm extends PureComponent {
             { classifications.map((classification, i) => (
               <div key={i}>
                 <h1>{classification.name}</h1>
```
troyblank: These might be word of mouth and not privy to you but the agreed on rules from my understanding is ALL options become autocompletes IF there is ONE classification greater than 9. This only covers individual classifications

screen shot 2018-06-14 at 4 40 56 pm
.

tVeratti: Ahh interesting - yeah I didn't gather that at all. I will make the changes :)

tVeratti:
@troyblank updated per your comment!

PR #592, opened 2018-06-18
TRW-1207 - ensure the minimum padding generated is large enough

tVeratti:
This test failed whenever the generated padding was not actually enough for the top of the element to be considered at the bottom. This needed to be forced to a minimum value based on the top value so that it would actually be a valid test case.

This also means I had to generate the values once and then save, instead of making the getter call `chance` so that the calculated padding is still large enough in the later util calls.


tVeratti:
I'm also thinking this caused the coveralls check to be less than 100% when this fail case hit, because `return true` was never hit within the `if` check of `util > dom > isAtBottomOfWindow`

troyblank:
@tVeratti  thank you so much for fixing this!

PR #597, opened 2018-06-20
TRW-1021 - hooked up onSearchQueryDelete to nodes

tVeratti:


components/universalSearch/searchBar.js
```diff
@@ -61,6 +61,15 @@ export default class SearchBar extends PureComponent {
         this.setState({ queries, queryIdCount: nextID });
     }
 
+    onSearchQueryDelete = (id) => {
+        const queries = [...this.state.queries];
+
+        const index = queries.map(q => q.id).indexOf(id);
```
jsheebs: Could you do ```queries = queries.filter(q => q.id !== id)```? Six in one, half dozen in the other, but makes it a one liner. :)

tVeratti: Well what I need to extract is the `index` of the id. If I filter the array first, the index is no longer accurate when I use it to pull the query out by index. It's a good idea though :)

jsheebs: Yea but your end game is removing that element from the array, yes? This does that.

tVeratti: Yeah I just realized I misunderstood your comment - I'm updating to use filter, great idea.

PR #599, opened 2018-06-22
TRW-1243 - added table and pagination to ResultsGrid

tVeratti:


components/table/table.scss
```diff
@@ -20,16 +20,14 @@
...
-
     li {
         @include breakpoint-below($size-tablet) {
             display: block;
         }
 
         display: flex;
         flex-wrap: wrap;
-        padding: 1rem 2rem;
+        padding: 1rem 0;
```
troyblank: Since we have introduced variable spacing is there any reason this should not use the variables?

components/universalSearch/resultsGrid.js
```diff
@@ -1,11 +1,45 @@
 import React, { PureComponent } from 'react';
 import styles from './resultsGrid.scss';
+import { Table, Paginator } from '../';
 
 export default class ResultsGrid extends PureComponent {
-
     state = { loading: false };
 
+    titles = [
+        'Topbox ID',
```
troyblank: I feel like this should be passed down so whatever consumes universal search can change this data without updating the visual components given this repos goal is more data agnostic. Thoughts?

tVeratti: I completely agree! I'll move the data up and plop it in through the demo.

components/table/table.scss
tVeratti: I think I was just keeping the same spacing as before - I'm fine with replacing with the variables though they are just a bit different.

troyblank: I have been trying to replace them when I can, if it looks too weird though I am fine with keeping the old values, or even making a new var possibly. But considering this is a new component I feel like we should definitely try to mold it into the new spacing var process thingie thing if possible.

PR #600, opened 2018-06-26
TRW-1028 - added labels to dimension drop down in universal search

troyblank:


components/selectDropDown/selectDropDown.js
```diff
@@ -120,19 +131,57 @@ export default class SelectDropDown extends PureComponent {
...
+
+        return labelsAndOptions;
+    }
+
+    renderOptions(options, indexOffset = 0) {
+        const { itemFocus } = this.state;
+        const { selectedIndex, onItemSelect, ariaID } = this.props;
+
+        return (
+          <span key={indexOffset} >
```
troyblank: When we upgrade react we can remove this span :/

test/components/selectDropDown/selectDropDown.spec.js
```diff
@@ -300,4 +303,65 @@ describe('Select Drop Down', () => {
...
+            ariaID={ariaID}
+          />
+        );
+        const instance = wrapper.instance();
+        const { itemFocus } = instance.state;
+        const indexOffset = options.length;
+        const optionsWithLabelsDoms = instance.renderLabelsAndOptions();
+        const labels = Object.keys(optionsWithLabels);
+
+        /* eslint-disable react/jsx-indent */
```
troyblank: I was in an infinite loop with indents here so I just disabled it... I started to just do lookups and ifExists but it didn't cover the `Options` so I decided to keep this as it is currently.

components/form/autoComplete/autoComplete.js
```diff
@@ -96,18 +117,18 @@ export default class AutoComplete extends PureComponent {
...
-        let regex;
-
-        options.forEach((option) => {
-            regex = new RegExp(value, 'gi');
-            if (null !== regex.exec(option)) {
-                newOptions.push(option);
-            }
-        });
+        const { options: oldOptions = [] } = this.props;
+        const options = cloneDeep(oldOptions);
```
tVeratti: If we don't mutate things, we shouldn't need to perform deep cloning logic. While it's true that each object in the array is still technically a reference, that's only a problem if we're directly mutating those references.

I would insist that we should put ourselves in the habit of not writing side-effect-y / mutate-y code as much as possible, so that we can write functional code with confidence that our objects aren't changing somewhere sneaky.

components/universalSearch/universalSearch.js
```diff
@@ -5,11 +5,15 @@ import Splash from './splash';
 import styles from './universalSearch.scss';
 
 export default function UniversalSearch({
+    popularDimensions,
     dimensions,
     queries = [],
     onQueriesChange,
     results }) {
-    const dimensionsOptions = dimensions.map(dimension => dimension.name);
+    const dimensionsOptions = {};
```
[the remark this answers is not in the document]
troyblank: This can probably wait till we get real data as it's probably easy to sprinkle in, but I would like to push back and really insure we need this data before we have to burden our uses with the extra bandwidth.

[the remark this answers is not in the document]
troyblank: Just to document; we talked and we agree we should keep the human readable name with the actual dimension. But this will be implemented in the future when we are dealing with data.

components/form/autoComplete/autoComplete.js
```diff
@@ -87,7 +101,14 @@ export default class AutoComplete extends PureComponent {
     onItemSelect = (i) => {
         const { options } = this.state;
 
-        this.selection = options[i];
+        if (Array.isArray(options)) {
+            this.selection = options[i];
+        } else {
+            const allOptions = flatten(Object.values(options));
+
+            this.selection = allOptions[i];
```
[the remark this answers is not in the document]
troyblank: Yea this once did serve somewhat of a purpose before the `SelectDropDown` component existed. But def needed to go so thanks for pointing it out. I refactored it out so that all relies on `this.state.value` now!

components/form/autoComplete/autoComplete.js
troyblank: I am still a bit confused on this. So given you have an object of arrays, you still have to mutate the object to alter it using `map` as we discussed.

```const myObject = { 'a': 1, 'b': 2, 'c': 3 }

Object.keys(myObject).map(function(key, index) {
   myObject[key] *= 2; // this is a mutation - and maps should return something not just mutate an object
});

console.log(myObject);```

I see no way to not mutate the `prop` without a clone at this point, but I could be missing something or over-thinking it. I did try NOT using map but still at some point you need to make a new object to build it with the filtered options, and at this point clone makes more sense to me. What am I missing?

I agree with the sentiment, I don't want to direct mutate an object, and I thought that is what clone was achieving in this example, using a new object to set state with.

Comment (troyblank) [reply components/form/autoComplete/autoComplete.js]: I updated this so it no longer uses `deepClone`. The result is the same though a object gets constructed and the direct options is not touched. That being said this is an improvement as the clone was just being over-written anyway. I might have mis-understood you and maybe this is what you were referring to but let me know if it's not.

**Comment by troyblank** [comment]
This should be ready for review again when you get time @tVeratti

File: components/form/autoComplete/autoComplete.js [hunk 6]
```diff
@@ -86,46 +97,58 @@ export default class AutoComplete extends PureComponent {
...
-
-    updateSelection(inputValue) {
-        const { name, onUpdate, allowAnyValue } = this.props;
-        let value = this.selection;
+    addInputedTextAsValue = (inputedValue) => {
+        const { options = [] } = this.state;
+        const { allowAnyValue, name, onUpdate } = this.props;
+        const allOptions = flatten(Object.values(options));
+        const isInOptions = allOptions.includes(inputedValue);
+        let value = '';
```
tVeratti: `let value = allowAnyValue || isInOptions ? inputedValue : '';` If you like ternary statements :) It's simple enough that I think it's a fitting case.

troyblank: ah yea that seems obvious now >< ... ok updated with ternary, please re-approved if you are good. I made it a `const` though ;)

PR #601, opened 2018-06-27
TRW-1267 - added pending animation for universal search

troyblank:


components/universalSearch/universalSearch.js
```diff
@@ -20,10 +22,16 @@ export default function UniversalSearch({
...
           onQueriesChange={onQueriesChange}
         />
-        {
-          queries.length
-              ? <ResultsGrid {...gridProps} />
-              : <Splash />
+        { showSplash &&
+          <Splash />
+        }
+        { pendingResults &&
```
tVeratti: While we don't expect these to overlap based on the logic - would it make sense to code it in a way that absolutely ensures that? Meaning - either a `switch(true)` or an `if/else` chain so that if by some mistake or something two of these end up being true, we don't end up with stacked wonky renders.

troyblank: Yea good call, ok updated this to be a `switch` to insure now two 'contents' show at the same time.

components/universalSearch/universalSearch.js
```diff
@@ -1,29 +1,60 @@
...
-        <SearchBar
-          dimensions={dimensions}
-          onQueriesChange={onQueriesChange}
-        />
-        {
-          queries.length
-              ? <ResultsGrid {...gridProps} />
-              : <Splash />
+        switch (true) {
+        case pendingResults:
```
tVeratti: Since you've put this in a function, you could just return it right away
`case pendingResults: return <div />;`

troyblank: ah yea, silly me, sorry! adjusted

PR #605, opened 2018-06-29
Porting jenkins stuff from master to develop

mathom:


grunt/mocha_istanbul.js
```diff
@@ -3,7 +3,8 @@ module.exports = {
         src: 'test/**/*.spec.js',
         options: {
             reporter: 'spec',
-            coverage: true
+            coverage: true,
+            reportFormats: ['cobertura', 'lcovonly']
```
tVeratti: Cobertura is one of the reporters that comes with istanbul/nyc right?

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-605/1/)

grunt/mocha_istanbul.js
troyblank: Yea, it's a format it allows as a final result. All supported are `clover, cobertura, html, json-summary, json, lcov, lcovonly, none, teamcity, text-lcov, text-summary, text.`

mathom: Cobertura is the format most Jenkins plugins understand so if you've got something to hook up, that's one to keep an eye out for.

PR #607, opened 2018-07-03
TRW-1349 - fixed malformed transcript search payload

troyblank:
This is a hotfix to fix the payload that is giving a 500 on prod this morning.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-607/1/)

PR #609, opened 2018-07-03
TRW-1063 - fixed more malformed transcript search payload properties

troyblank:
missed the "text" property.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-609/1/)

PR #610, opened 2018-07-03
TRW-1349 - refactored transcript search form to use dimension and value instead of text and speaker

troyblank:
OK, I learned my lesson today. More HOTFIX fun!

I did some refactoring here to do two things:

- Insure the node list works right on transcript search but reverting back to the previously used props
- use `setState` instead of manipulating state directly.

That being said this is basically a revert of what I did earlier this morning. I think we should do a re-`map` of data on the rci side before passing to the API. Meaning we alter the prop names of the query nodes slightly to cater to the API and keep the internal prop names consistent with all components here so they are more easily passed in harmony.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-610/2/)

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-610/3/)

test/components/forms/transcriptSearchForm.spec.js
```diff
@@ -129,23 +130,23 @@ describe('Transcript Search Form', () => {
...
+        const value = chance.word();
+        const value2 = chance.word();
         const wrapper = mount(
           <TranscriptSearchForm onSubmit={onSubmit} />
         );
         const instance = wrapper.instance();
         const form = wrapper.find('form');
 
-        instance.searchTerms.state.value = text;
+        instance.searchTerms.state.value = value;
```
tVeratti: no `setState` here?

components/forms/transcriptSearchForm.js
```diff
@@ -22,20 +22,25 @@ export default class TranscriptSearchForm extends Component {
...
-            this.setState({ queryNodes: this.state.queryNodes });
-            this.idCount += 1;
+        if (value) {
+            const { queryNodes, idCount } = this.state;
 
-            this.props.onSubmit(this.state.queryNodes);
+            this.searchTerms.setState({ value: '' });
+            this.setState(
+                {
+                    queryNodes: queryNodes.concat([{ id: idCount, value, dimension, type }]),
```
tVeratti: If it were me, I'd use the spread operator purely because I find it more readable. What do you think?
```
this.setState({
    queryNodes: [ 
        ...queryNodes,
        { id: idCount, value, dimension, type }
    ],
    idCount: (idCount + 1)
})
```

troyblank:
@tVeratti I updated the tests to use all setState, thanks for pointing out the ones I missed. Also I updated to use `spread` instead of `concat`. I think spread makes more sense but it's not quite a habit for me with arrays, so keep me on track to be consistent :D

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-610/4/)

PR #613, opened 2018-07-06
TRW-1369 - Added HelperSwitch to start rendering conditional helpers

tVeratti:
No styles started for how the helpers will be contained within the little helper button, the primary goal here was to start a very flexible (but sturdy) approach that will let us render N helpers in a consistent way, despite being very different interfaces. This will be accomplished by customizing the `props` object in the `helpers` property, which essentially our front-end map of component-to-type. This does require some careful configuration on our part.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-613/1/)

PR #614, opened 2018-07-06
TRW-1369 updated HelperSwitch styles, tests, and demo dims

tVeratti:
I've also moved the dimensions/helper map into its own file. This necessitated passing the onChange as a parameter - and possibly more parameters in the future.

`Select` is now used instead of `SelectDropDown`, which will need an alternative style created to match the designs.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-614/1/)

PR #615, opened 2018-07-06
TRW-1064 - added highlights to table - structure only no styles

troyblank:
This alters the data structure for search results to allow adding transcript highlights as well as adds the html structure to show highlights.

This has very little styling - going to do that in another PR.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-615/1/)

components/table/table.js
```diff
@@ -69,9 +80,19 @@ export default function Table({
...
-            <ul className={styles.table__list}>{rows.map((row, i) =>
-              <li key={i}>
-                {row.map((value, j) => renderColumn(value, j, exports.getColumnClasses(columnStyles[j]), titles[j]))}
+            <ul>{rows.map((row, i) =>
+              <li
+                className={styles['table__list-item']}
+                key={i}
+              >
+                <div className={styles.table__cells}>
+                  {row.cells.map((value, j) => renderColumn(value, j, exports.getColumnClasses(columnStyles[j]), titles[j]))}
```
tVeratti: Does it really need `exports.` or can you just refer to the local definition?

tVeratti: Also, would it make sense to much up the naming between `cells` and `columns`?

components/table/table.scss
```diff
@@ -19,20 +19,22 @@
...
+        display: block;
     }
+
+    display: flex;
+    flex-wrap: wrap;
+    align-items: center;
+    padding: $tiny-spacing 0;
+
+    color: $color-black;
+    cursor: pointer;
```
tVeratti: Are the cells all going to be clickable? `pointer` would make me expect a behavior on left click.

components/table/table.scss
```diff
@@ -90,3 +92,15 @@
     display: none;
     text-decoration: underline;
 }
+
+.highlights {
+    display: grid;
```
tVeratti: Nice! Stepping into `grid` fun!

tVeratti: What made you decide on `grid` instead of `flex` here, out of curiosity?

troyblank: `grid` comes with a lot of responsiveness for free in the case you have things being stacked in very specific "amounts per row". Thus my thought is these little things are def layed out in a grid and using `minmax` we could write cleaner css without a lot of media queries. If these were going to need to flow differently based on media queries I would have done `flex` but alas they will not. Something like what this person is doing > https://medium.com/samsung-internet-dev/common-responsive-layouts-with-css-grid-and-some-without-245a862f48df

Please note though the styles are not really done yet so this is just an initial "get things sorta looking ok" step.

components/table/table.scss
troyblank: They will be clickable, but there are not right now so I will remove this as it doesn't make sense currently.

components/table/table.js
troyblank: Apparently we do not need `exports` :) removed that... and also great call on the naming, I moved everything to use `cells` as they are more descriptive for what the styles are targeting.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-615/2/)

PR #617, opened 2018-07-09
TRW-1378 - added styling to grid highlights

troyblank:
screen shot 2018-07-09 at 5 05 15 pm


components/table/table.scss
```diff
@@ -92,13 +92,26 @@
 }
 
 .highlights {
+    @include breakpoint-below($size-tablet) {
+        margin-left: 0;
+    }
+
     display: grid;
+    // calc decently fakes cell alignment with 
+    margin: 0 0 $small-spacing calc(#{$medium-spacing} + 4.5%);
```
troyblank: Calc is a little wonky, but I feel it beats passing in a prop for this style at this juncture given highlights are only for universal search at this point or muddying up the html.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-617/1/)

components/interactionEditor/HTMLText/HTMLText.js
```diff
@@ -64,7 +65,9 @@ export default function HTMLText({ text, citations = true, emSelector = ':scope
       <div className={styles['html-text']}>
         <div
           // eslint-disable-next-line react/no-danger
-          dangerouslySetInnerHTML={{ __html: html }}
+          dangerouslySetInnerHTML={{
+              __html: dompurify.sanitize(html)
```
tVeratti: Good thinking for the security here.

components/table/table.scss
tVeratti: What's the goal here, to have a minimum space of `medium-spacing` plus 4.5%, essentially?

components/interactionEditor/HTMLText/HTMLText.js
troyblank: Goal was to achieve the red space below:
space

The reason it is wonky because I am leaving the html structure so the view button (the orange eye one) is part of the content above. This presents an issue with alignment that is solved pretty convincingly with that not super normal css. That being said we could make that view button a separate element and not with the content of the rest of the stuff. This is doable and makes this css more clean cut but makes our render logic more complicated. I was on the fence but leaned towards making our css slightly strange. Admittedly the alignment is not perfect but it is near perfect, this is something I would love for the design team to review.

troyblank:
In reply to the css comment (not sure why the reply text box is not showing for that one....)

Goal was to achieve the red space below:
space


The reason it is wonky because I am leaving the html structure so the view button (the orange eye one) is part of the content above. This presents an issue with alignment that is solved pretty convincingly with that not super normal css. That being said we could make that view button a separate element and not with the content of the rest of the stuff. This is doable and makes this css more clean cut but makes our render logic more complicated. I was on the fence but leaned towards making our css slightly strange. Admittedly the alignment is not perfect but it is near perfect, this is something I would love for the design team to review.

components/table/table.scss
troyblank: Goal was to achieve the red space below:
space

The reason it is wonky because I am leaving the html structure so the view button (the orange eye one) is part of the content above. This presents an issue with alignment that is solved pretty convincingly with that not super normal css. That being said we could make that view button a separate element and not with the content of the rest of the stuff. This is doable and makes this css more clean cut but makes our render logic more complicated. I was on the fence but leaned towards making our css slightly strange. Admittedly the alignment is not perfect but it is near perfect, this is something I would love for the design team to review.

troyblank: Actually, after pondering this and throwing some other cases at it this morning this is not going to work anyway. I will need to make the view button it's own thing. I will probably supply a prop that if used shows it.

tVeratti: A thought - I've worked with a similar problem before but I was using `display: table`. I had an empty cell at the start for the *child* row that allowed the sizing to match up. I can't say exactly how you might get the same effect with flex + grid, but maybe just the idea of an empty `cell` could open another branch of thought.

PR #618, opened 2018-07-10
TRW-1296 - added date type and datepicker helper to search bar

tVeratti:
This may need some slight data massaging before we pass it to the queries, but https://topbox.atlassian.net/browse/TRW-1388 will handle making all `onChange` calls normalized for data that we can return to the queries.


tVeratti:
Getting the core of the helper system tested out with this helper/component was the goal here. The last piece to complete this one is to add the interaction/observation radio selection, sub-task here: https://topbox.atlassian.net/browse/TRW-1389

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-618/1/)

components/datepicker/datepicker.js
```diff
@@ -64,9 +65,20 @@ export default class Datepicker extends Component {
...
         }
 
+        const value = this.formatSelection(range);
+        if (onChange) onChange({ value, range });
         this.setState({ selectionRange: range });
     }
 
+    formatSelection = (range) => {
+        const toFormat = date => `${date.getMonth()}/${date.getDate()}/${date.getFullYear()}`;
+
```
troyblank: why a space here?

components/icons/calendar.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
troyblank: I feel like this svg can be improved. The use of lines instead of fills are generally bad if the svg happens to scale. You can supply a counter argument but I personally tend to convert all my lines into fills so scaling is never an issue. That being said if we do make these fills there is a lot of shape combining we can do so that many lines can become two shapes for this graphic. I feel like doing this not only makes the svg more flexible but also tends to make it smaller and easier to render thus less likely for any issues in most cases.

There also is a lot of odd artifacts that I feel can be removed during this optimization process like `translate(0.000000, 1.000000)`.

components/universalSearch/helperSwitch.js
```diff
@@ -4,6 +4,27 @@ import styles from './helperSwitch.scss';
 import helperMap from './helperMap';
 
 export default class HelperSwitch extends PureComponent {
+    state = {
+        helperValues: {}
+    };
+
+    /* eslint-disable camelcase */
+    onChange_date = e => this.onChange('date', e);
```
troyblank: Why are we breaking the lint rule we have for these?

components/universalSearch/helperSwitch.scss
```diff
@@ -11,5 +11,8 @@
 }
 
 .helper-switch__helper {
+    display: flex;
+    align-items: center;
+    &>span { margin-left: $small-spacing; }
```
troyblank: should this be it's own class for helper values?

components/icons/calendar.js
```diff
@@ -0,0 +1,29 @@
...
+            <path d="M10.5,11.5 L10.5,6.5 L10.5,11.5 Z" strokeLinecap="round" strokeLinejoin="round" />
+            <path d="M11.5,3 L11.5,0 L11.5,3 Z" strokeLinecap="round" strokeLinejoin="round" />
+            <path d="M0,4.5 L16,4.5" strokeLinecap="round" strokeLinejoin="round" />
+            <path d="M2.5,10.5 L13.5,10.5" strokeLinecap="round" strokeLinejoin="round" />
+            <path d="M2.5,7.5 L13.5,7.5" strokeLinecap="round" strokeLinejoin="round" />
+          </g>
+        </g>
+      </svg>
+    );
+}
```
troyblank: I don't see this icon in our visual inventory of icons. I feel like we should always keep track visually of everything we are using currently as it allows a quick way to see everything we have for many reasons.

screen shot 2018-07-10 at 9 50 56 am

components/universalSearch/helperSwitch.js
tVeratti: I'm using a dynamic way of assigning the function so that we don't have to `.bind` the `onChange` in the render. If I were to make ti camelcase, then I'd need to start formatting/manipulating the "types" instead of just going lowercase. I also think it helps differentiate it a little bit from a standard function, since this is like a feeder that leads right back into the base `onChange`.

components/icons/calendar.js
tVeratti: I will add it!

components/universalSearch/helperSwitch.scss
tVeratti: Yeah may as well :)

components/datepicker/datepicker.js
tVeratti: It felt like a definition kind of `pre` to the core of the rest of the function. Everything else uses this as a mini utility, and it's a bit heavy with symbols so for readability it seemed nicer to break it apart.

tVeratti:
@troyblank updated per your comments.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-618/2/)

PR #620, opened 2018-07-10
TRW-1397 - refactored the way table handles link rendering

troyblank:
This makes the style of the table with links less fragile but also have a tad more markup.


assets/js/demo/components/universalSearch.js
```diff
@@ -43,8 +32,7 @@ export default class UniversalSearchDemo extends Component {
             dimensions={dimensions}
             pendingResults={pendingResults}
             onQueriesChange={this.getResults}
-            titles={this.titles}
-            results={results}
+            tableData={{ titles, results, links }}
```
troyblank: @tVeratti  please don't hate me for this... I know it will sorta break our current tagged demo though but with all three of these props I feel like it makes things more tidy and it's early enough it won't break anything for users :D

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-620/1/)

assets/js/demo/components/universalSearch.js
tVeratti: What about
```
onQueriesChange={this.getResults}
{ ...{ titles, results, links }}
/>
```
Not a common syntax but I have played with the idea fo a long long time. I don't hate yours, but wanted to suggest this in case it clicks for you as well.

assets/js/mockData/searchResults.js
```diff
@@ -1,46 +1,58 @@
...
+            cells: ['IKGac37144612810', 'Smith, Katrina', '06/04/17 - 6:37 am', <IconFlag color={'blue'} key={'flag-5'} />, 'Sentiment', 'Johnson, Erik', '06/04/14 - 6:37 am'],
             highlights: [['...my son took <span>product</span> test tennis shoes in january of...']]
         },
         {
-            cells: [<ViewButton href={'/pages/universalSearch.html'} key={'view-6'} />, 'IKGac37144612810', 'Smith, Katrina', '06/04/17 - 6:37 am', <IconFlag color={'blue'} key={'flag-6'} />, 'Sentiment', 'Johnson, Erik', '06/04/14 - 6:37 am']
+            cells: ['IKGac37144612810', 'Smith, Katrina', '06/04/17 - 6:37 am', <IconFlag color={'blue'} key={'flag-6'} />, 'Sentiment', 'Johnson, Erik', '06/04/14 - 6:37 am']
         },
         {
-            cells: [<ViewButton href={'/pages/universalSearch.html'} key={'view-7'} />, 'IKGac37144612810', 'Smith, Katrina', '06/04/17 - 6:37 am', <IconFlag color={'blue'} key={'flag-7'} />, 'Sentiment', 'Johnson, Erik', '06/04/14 - 6:37 am']
```
tVeratti: I like that you moved this out from being expected in the data. Good move.

components/table/table.js
```diff
@@ -60,13 +61,25 @@ function renderHighlight(value, key) {
     );
 }
 
+function renderLink(link) {
+    return (
+      <div className={styles.table__link}>
```
tVeratti: Do you want this div wrapper to render when there is no link? Would that leave a blank space/clutter divs?

components/table/table.js
```diff
@@ -79,7 +92,11 @@ export default function Table({
           </div>
         </header>
         <div>
-          <div className={styles.table__labels}>
+          <div
+            className={classNames(styles.labels, {
```
tVeratti: I tend to define the classNames above since it feels like it adds a lot of logic in the already busy HTML. Do we want to pick one way or the other? I'd argue for defining complicated things above and using a cleaner variable name, but I'm open to hearing why inlining it is better as well.

components/table/table.js
troyblank: Yes, we definitely need the blank space to insure the alignment of the label vs value.

assets/js/demo/components/universalSearch.js
troyblank: The thing I like about mine is that it is more descriptive. When I added `links` I thought to myself links for what? It wasn't apparent it was for the table. So this made more sense to me. Granted maybe now that I typed this `resultTableData` might even be more specific....

components/table/table.js
troyblank: Yea that is cool with me, I am all for less complicated HTML.  Updated!

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-620/2/)

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-620/3/)

PR #621, opened 2018-07-10
TRW-1387 - updated mobile styles for universal search results

troyblank:
This builds off TRW-1397!!! so wait till that is merged if you want this PR to look smaller.

Update mobile styles for search results to be much better than before.


components/table/table.scss
```diff
@@ -84,11 +114,12 @@
 
 .table__vertical-label {
     @include breakpoint-below($size-tablet) {
```
tVeratti: I general I think an approach that has worked a lot better for me is to use breakpoints targeted for this component only, that represent the points at which the component fails in the UI, instead of global device sizes. This would be a pretty big change but I'd like to start the conversation about it.

test/components/table/table.spec.js
```diff
@@ -50,15 +51,15 @@ describe('Table', () => {
         assert.isTrue(wrapper.contains(
           <div className={styles.table}>
             <header className={styles.table__header}>
-              <div>
+              <div className={styles.table__header__count}>
```
tVeratti: So this is a style of BEM that I don't think we've been using, but I'm open to discussing in. In the style I've advocated, this would be `table__count` or `table__header-count` (always only one deep).

troyblank: Yea that works for me, sorry team I was with one time used this style so I need to break myself of it. Updated!

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-621/2/)

components/table/table.scss
troyblank: Seems smart to me, we should discuss this with Jason about how we want to do it and go forward. But generally speaking thinking about break points on a component level would probably help everything overall look and work better responsively so it would be a good change in my eyes.

PR #623, opened 2018-07-11
TRW-1408 - passed options to helperSwitch & changed value rendering

tVeratti:
Minor fixes to polish things up for the Universal Search helpers. Options being passed will allow us to query the values on the server-side and pass it to the helper(s).


components/universalSearch/helperSwitch.js
```diff
@@ -8,6 +8,9 @@ export default class HelperSwitch extends PureComponent {
         helperValues: {}
     };
 
+    // These specific onChange functions allow us to
```
tVeratti: I added this comment because it was a good question you asked on the initial review. I believe in *why* comments, and this might be useful to explain in case someone doesn't understand the purpose in the future.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-623/1/)

PR #624, opened 2018-07-11
TRW-1409 - added prop to stop filtering on autocomplete

troyblank:


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-624/1/)

assets/js/components.js
```diff
@@ -258,6 +258,15 @@ renderHtml(
       onUpdate={e => print(e, `AutoComplete: { ${e.name} ${e.value} }`, false)}
     />
     <br />
+    <AutoComplete
+      id={'autoCompleteNoFilter'}
+      name={'autoCompleteNoFilter'}
+      label={'Auto Complete Without Filtering'}
+      activeFiltering={false}
```
tVeratti: So kind of a minor nit-pick, I feel like this is the kind of prop that we should force to `true` and is `falsey` by default. The nice thing about that is that you don't have to do the `activeFiltering = true` defaulting and allow undefined (all existing components) to make all of the existing logic work the same until we force it differently.

components/form/autoComplete/autoComplete.js
```diff
@@ -143,18 +143,20 @@ export default class AutoComplete extends PureComponent {
     }
 
     updateOptions(value) {
-        const { options: oldOptions = [] } = this.props;
+        const { activeFiltering = true, options: oldOptions = [] } = this.props;
```
tVeratti: Continuing from my comment above, if we change it to a default false value, you can avoid adding an extra if/else depth by going the `if (noFilter) return;` path.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-624/2/)

components/form/autoComplete/autoComplete.js
troyblank: Ok! sorry so missed this one, i love it! updated!

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-624/4/)

PR #625, opened 2018-07-11
TRW-1034 - pass dimension onChange handlers up to UniversalSearch

tVeratti:
This is how we will perform the dimension autocomplete queries based on the typed value in the helper. Just needed to bubble that all the way up to where RCI could take it and send to the API.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-625/2/)

PR #633, opened 2018-07-12
TRW-1419 - dimension changes to US root, reset helpers, autocomplete value held

tVeratti:
These are a few of the little things I've come across needing fixes for the overall Universal Search workflow. I'm sure there will be more.

Most notable changes:
I added keys to both the `HelperSwitch` and `AutoComplete` components within the `SearchBar`. This forces them to reset when the dimension changes or when a query is added, respectively. This helps usability in not having to remove old values, and it also reset the inner states that hold onto the info about the previous query data.


components/universalSearch/helperSwitch.js
```diff
@@ -14,20 +14,18 @@ export default class HelperSwitch extends PureComponent {
     /* eslint-disable camelcase */
     onChange_channel = e => this.onChange('channel', e.value);
     /* eslint-disable camelcase */
```
troyblank: this should all be wrapped in a `eslint-disable` with a `eslint-enable` at the end. From my understanding there is no reason to do this three times. I think what you were assuming you were doing was `eslint-disable-line`.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-633/2/)

PR #635, opened 2018-07-12
TRW-1412 - fixed small issue where autocomplete corners did not match design

troyblank:


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-635/1/)

troyblank:
nope @jasonhale it's the parent controlling a child, and we use css modules :/

PR #637, opened 2018-07-16
TRW-1429 - double clicking a table row with a link navigates to link

troyblank:
Adds the double click link for universal search rows.

In order to get double click link to work well with testing for our universal search table I had to separate  `<Table />` into two different components adding `<Row />`. This allows us to handle the click event more cleanly.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-637/1/)

components/table/row.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.highlights {
+    @include breakpoint-below($size-tablet) {
+        margin-left: 0;
+    }
+
+    display: grid;
+    margin-bottom: $small-spacing;
+    grid-gap: 1rem;
+    grid-template-columns: repeat(auto-fit, minmax(300px, 1fr));
```
tVeratti: Fancy. I like the use of grid here. I wasn't familiar with this way of setting the columns - just did a bit or reading on this. Noice.

test/components/table/row.spec.js
```diff
@@ -0,0 +1,123 @@
...
+            row={row}
+            cellStyles={cellStyles}
+            link={link}
+          />
+        );
+        const linkItem = wrapper.find(`.${styles.row}`);
+
+        window.location.assign = sinon.spy();
+        linkItem.simulate('dblclick');
+        assert.isTrue(window.location.assign.calledWith(link));
```
tVeratti: Thorough tests, nice.

components/table/row.scss
troyblank: Yea, the huge benefit is a lot less media queries.

PR #641, opened 2018-07-17
TRW-1442 - added view interaction option for ugly context menu on table component

troyblank:
The actual menu is ugly for now, but this allows a user to click and select "view interaction" which just changes the window location. Some style work will need to come to help polish this.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-641/1/)

components/table/table.scss
```diff
@@ -28,20 +28,24 @@
...
+
 .table__column {
     @include breakpoint-below($size-tablet) {
         padding: $base-spacing 0;
         border-bottom: 1px dashed $color-grey;
         white-space: normal;
     }
 
+    flex: 1;
     flex-basis: 0;
```
jasonhale: `flex` is shorthand property that includes `flex-basis`... shouldn't this be more like `flex: 1 1 0`?

troyblank: Yup it should be, totally overlook this thanks. I am going to fix this and a small alignment issue I introduced with this PR with header titles vs content.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-641/3/)

PR #642, opened 2018-07-18
TRW-1490 - allowed table cell to be able to copy the value

troyblank:
Copy on cell menus are now working.

This still does not make the cell menu pretty, nor does handle the weirdness of copying icons that come out to "[Object, Object]". These two things will be handled very soon in other PRs.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-642/1/)

components/table/cell.js
```diff
@@ -1,28 +1,93 @@
...
+            ariaID: uniqueId()
+        }
+    };
 
     onClick = () => {
         const { showingContextMenu } = this.state;
 
-        this.setState({ showingContextMenu: !showingContextMenu });
+        this.setState({
+            showingContextMenu: !showingContextMenu,
```
jasonhale: I've seen this pattern a fair bit, and doubt you've had issues yet.  But since React batches state updates, and this specifically updates the state relative to it's current state, may look to use `prevState`.
```js
this.setState((prevState) => {
  return {
    showingContextMenu: !prevState.showingContextMenu,
    layoutInformation: getLayoutInformation(this.cell)
  }
);

troyblank: @jasonhale what you say makes too much sense. I feel like we should do this way from now on and should be a request for change. I updated the code if you don't mind re-reviewing please?

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-642/2/)

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-642/4/)

PR #643, opened 2018-07-18
TRW-1490 - copy option on table cell does not show if value is not a string

troyblank:
For universal search this means you can not copy the flag icon. I did talk to design about this to insure it seemed like the best UX.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-643/2/)

PR #646, opened 2018-07-19
TRW-1066 - report sorting changes to UniversalSearch

tVeratti:


assets/js/demo/components/universalSearch.js
```diff
@@ -36,6 +37,7 @@ export default class UniversalSearchDemo extends Component {
             dimensions={dimensions}
             pendingResults={pendingResults}
             onQueriesChange={this.getResults}
+            onTitleClick={(title, index) => print(null, { title, index })}
```
troyblank: This confuses me. Wouldn't the table only care about a title click? I would assume universal search should only care about sorting actions.

components/table/table.js
```diff
@@ -17,58 +17,65 @@ export function formatPageInfo(pageInfo) {
...
+              </div>
+            </header>
+            <div>
+              <div className={labelClasses}>
+                {titles.map((title, i) =>
+                  <TitleCell
+                    key={title}
+                    title={title}
+                    index={i}
+                    onClick={onTitleClick}
```
troyblank: As above I still think this can be encapsulated and bubble up a a "sortChange" event of some form.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/2/)

components/table/table.js
tVeratti: Agreed, working on making this change now - thanks!

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/3/)

tVeratti:
@troyblank  - So I U-turned on how this was starting. I decided to use the `Cell` component for the title cells after all. This meant adding a `sort` action of course, which will come in very handy for the next step which is to add those menu options to individual regular cells. I think this tackles two birds with one stone, and reduces the code duplication I had with `TitleCell`.

Kudos to the great work you did on `Cell` to make this reusable with very minor changes.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/5/)

tVeratti:
Renamed the PR because it's original name is no longer relevant since that component has been deleted.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/7/)

components/icons/arrow.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default function IconArrow({ direction = 'up' }) {
+    const directionDegrees = {
+        up: 0,
+        down: 90
+    };
+
+    const rotation = directionDegrees[direction] || 0;
+
+    return (
```
troyblank: This can be optimized from my perspective, I was able to get this using svgomg + cleaning some stuff in illustrator:
```
<svg
  xmlns={'http://www.w3.org/2000/svg'}
  width={16}
  heigh={16}
  viewBox={'0 0 16 16'}
  style={{ transform: `rotate(${rotation}deg)` }}
>
  <title>{direction} arrow</title>
  <path
  	fill={'currentColor'}
  	d={'M13 7.5c0-.3-.2-.5-.5-.5s-.5.2-.5.5v3.8L2.9 2.1c-.2-.1-.6-.1-.8 0s-.1.6 0 .8l9.1 9.1H7.5c-.3 0-.5.2-.5.5s.2.5.5.5h5.2c.1-.1.2-.1.3-.3V7.5z'}
  />
 </svg>
```

components/icons/sort.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export default function IconSort() {
+    return (
+      <svg
+        xmlns={'http://www.w3.org/2000/svg'}
+        width="16"
```
troyblank: This is probably super nit picket but I like to keep everything consistent as js values such as width={16}. Let me know if you agree :) or disagree.

components/table/table.js
```diff
@@ -17,58 +19,83 @@ export function formatPageInfo(pageInfo) {
...
+        } = this.props;
+        const { sort } = this.state;
+
+        const formatedPageInfo = formatPageInfo(pageInfo);
+        const hasLinks = Boolean(links && links.length);
+        const labelClasses = classNames(styles.labels, {
+            [styles['labels--has-links']]: hasLinks
+        });
+
+        return (
```
troyblank: For some reason this extra structure broke the small screen layout. The label and value should be side by side compared to this > https://s3-us-west-2.amazonaws.com/rci-components/tiles.html#searchResults

screen shot 2018-07-23 at 4 58 50 pm

components/icons/arrow.js
tVeratti: I can remove the properties. My hesitation is that last time I did, Casey pointed out that it was longer rounded. I couldn't notice the difference until he mentioned it - so I decided to avoid removing some of the properties that came with his stuff. However, if you have confidence that those tools wouldn't remove something with a real effect then I don't mind taking them out.

components/table/table.js
tVeratti: Ahhh -- I needed `inline-flex` instead of `flex`. Fixing!

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/9/)

components/icons/arrow.js
troyblank: I am pretty confident as this is what I have been doing to all the svgs. Here is a screen render of my above code. Round corners stayed intact.

screen shot 2018-07-24 at 9 19 38 am

troyblank: If you slap a `transform: scale(10)` on this you can see the rounded corners for future use. Granted scale does not let it anti-alias well but you totally can see the roundies 👯‍♂️ 

screen shot 2018-07-24 at 9 30 53 am

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/10/)

components/icons/arrow.js
tVeratti: Noice! Wouldn't want to mess with the roundies 👯 
Thanks for cleaning that SVG up!

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-646/11/)

PR #647, opened 2018-07-20
TRW-1515 -- refactored select drop down styles to be more flexible

troyblank:
Noticed a lot of rough edges around the select drop as it relates to universal search. This is work to smooth everything out.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-647/1/)

components/selectDropDown/selectDropDown.js
```diff
@@ -183,34 +184,46 @@ export default class SelectDropDown extends PureComponent {
...
             aria-expanded={showing}
           >
-            <ul className={styles['select-drop-down__content']}>
-              { this.renderLabelsAndOptions() }
-            </ul>
+            <div
+              className={styles['select-drop-down__chrome']}
+              style={maxHeight}
+            >
+              <ul className={styles['select-drop-down__content']}>
```
tVeratti: This is a thought not directly tied to these changes - what do you guys think about not rendering things like options until they open? My only concern atm is accessibility, but it seems that perhaps with all the proper labels and things like `aria-expanded` we should be covered... though I'm speaking from theory and not actual fact.

The benefits would include:
- drastically reducing the amount of elements on a page at load and in general
- less js work (not open, don't bother looping through render logic for each child)

Thoughts?

components/selectDropDown/selectDropDown.js
```diff
@@ -183,34 +184,46 @@ export default class SelectDropDown extends PureComponent {
...
-            optionsPaneBottomUpPadding
+            optionsPaneBottomUpPadding,
+            styleVariations
         } = this.props;
         // padding is added as a bridge from content to select drop for more friendly mouse events.
+        // see `.select-drop-down` styles.
         const paddingAmount = 10;
         const { isAtBottomOfWindow, isAtRightOfWindow, height } = parentLayoutInformation;
-        const style = { maxHeight: optionsPaneHeight };
+        const maxHeight = { maxHeight: optionsPaneHeight };
```
tVeratti: I would consider naming this something like `chromeStyle` so that it not only tells you what it's built for upfront, but is also more ready for any additional inline styles in the future. The name right now doesn't give any extra info that the declared value doesn't already provide.

components/selectDropDown/selectDropDown.scss
```diff
@@ -1,18 +1,28 @@
 @import '../../assets/sass/base/base';
 
 .select-drop-down {
+    // This outside element + chrome seperation is to give increase the mouse area of the drop down.
+    // Allowing more friendly mouse events.
```
tVeratti: This is good thinking. I'm always a fan of giving some slack to click areas.

components/selectDropDown/selectDropDown.js
troyblank: Seems pretty darn logical to me. I would want to make a new story for it though. I assume you are down so I will go ahead and make the story.

components/selectDropDown/selectDropDown.js
troyblank: OK changed this to `__wrapper`. I didn't like `__chrome` a lot so this is great.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-647/4/)

PR #648, opened 2018-07-23
TRW-1335 - converted universal search bar to a form

troyblank:
In order to get return key working I turned `<SearchBar />` to use `<form onSubmit />`.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-648/1/)

PR #649, opened 2018-07-23
TRW-1521 - made select drop down max pane height test more solid

troyblank:
Looks like there was potential for a little overlap here depending on virtual DOM. I made this a bit more conservative so the test hopefully will not fail anymore.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-649/1/)

PR #650, opened 2018-07-23
TRW-1032 - added freetext helper to HelperSwitch

tVeratti:


components/universalSearch/helperMap.js
```diff
@@ -10,9 +10,10 @@ function Popout(ComponentClass, Icon, props) {
     );
 }
 
-export default function (onChange, options) {
-    return {
-        channel: {
+export default function (type, onChange, options) {
+    switch (type) {
```
tVeratti: While I'm a big fan of dictionaries for mappings, I decided to change a `switch` mainly for the `default` use case - allowing use to render a `freetext` helper if for some reason we get a type that doesn't match our existing list.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-650/3/)

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-650/4/)

PR #651, opened 2018-07-24
TRW-1523 - added timepicker component that only shows am and pm hour selections as first step

troyblank:
This just adds time buttons along with the new `<timePicker />` component. Next step will be to allow the component to make selections and report to the parent what selection was made.

screen shot 2018-07-24 at 11 17 03 am


jasonhale:
Got some file name conflicts - camelCase vs lowercase

components/timePicker/timepicker.js
```diff
@@ -0,0 +1,51 @@
...
+        this.timeOptions = Array.from(Array(24), (x, hour) => ({
+            label: moment(hour, 'h').format('h:mm'),
+            value: hour
+        }));
+        this.amHourOptions = this.timeOptions.splice(0, 12);
+        this.pmHourOptions = this.timeOptions.splice(0, 12);
+    }
+
+    onHourSelect = () => {
+        // selection logic coming soon to a PR near you.
```
tVeratti: In a world... where times must be selected...

components/timePicker/timepicker.js
```diff
@@ -0,0 +1,51 @@
...
+import moment from 'moment';
+import HourOption from './hourOption';
+import styles from './timePicker.scss';
+
+export default class Timepicker extends PureComponent {
+    /* istanbul ignore next */ // https://github.com/gotwarlost/istanbul/issues/690
+    constructor(props) {
+        super(props);
+
+        this.timeOptions = Array.from(Array(24), (x, hour) => ({
```
tVeratti: I've done the same with ` [ ...Array(24) ].map` - as a suggestion of an alternate syntax.

tVeratti: I've also created a tiny shorthand util for myself in the past when I found myself needing this more often:
`const arr = count => [...Array(count)];` (arrrr like a pirate... I know that name wouldn't satisfy you so feel free to imagine your own in its place :D )

components/timePicker/hourOption.js
```diff
@@ -0,0 +1,23 @@
+import React, { PureComponent } from 'react';
+import styles from './hourOption.scss';
+
+export default class HourOption extends PureComponent {
+    onClick = () => {
+        const { value, onHourSelect } = this.props;
+
+        onHourSelect(value);
```
tVeratti: I've been experimenting with defining functions like this in the `defaultProps`. I feel like always having checks if they exist starts to clutter things a bit, but we still wouldn't want this to explode if it didn't have one. HOWEVER, this might not be a good example because this component most likely won't ever be used outside of this very specific context so there won't be room for mistakes like that anyway.

troyblank:
@jasonhale  sorry I do not follow you. Do you mean `timepicker` vs `timePicker`. If so I followed what veratti did with `datepicker`. There is already a `timePicker`, so any `timepicker` is intentional.

components/timePicker/timepicker.js
troyblank: This is def a better route. Change.

PR #653, opened 2018-07-25
TRW-1537 - time picker is now selectable.

troyblank:


assets/js/components.js
```diff
@@ -434,7 +434,7 @@ renderHtml(
 
     <hr id="timepicker_2" />
     <h2>Timepicker v2</h2>
-    <Timepicker />
+    <Timepicker onUpdate={e => print(e, `Timepicker: { ${e.value} }`, false)} />
```
troyblank: @tVeratti  I know this is different from the datepicker which is `onChange`, but is the same as all the form inputs we have built previously. I have find going one way or the other but I was attempting to be as consistent as possible here for elements that make up forms.

tVeratti: Hmm yeah, I like consistency. My only argument is that `onChange` is more globally consistent with natural elements. If we want to explicitly separate those concepts then sure, but otherwise `onUpdate` is always going to create inconsistency. So whichever way we choose I'm cool with sticking to in the future.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-653/1/)

assets/js/components.js
troyblank: Ah, I sort of saw those as conflicts and not consistency. But we will never have a `onChange` prop on a raw input so I can get behind this. Changing over to `onChange`.

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-653/2/)

PR #655, opened 2018-07-26
TRW-1547 - added quick time selectors to time picker

troyblank:
@jasonhale This adds the quick select to the right of the time picker. Next will be adding it to universal search then adding the radios for date category.


mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-655/1/)

components/timePicker/hourOption.scss
```diff
@@ -1,36 +1,45 @@
 @import '../../assets/sass/base/base';
 $selection-radius: 20px;
+$above-small-breakpoint: 351px;
+$small-breakpoint: 350px;
+
+@mixin selected-border-radius($row-size, $one-less-than-row-size) {
+    &:nth-child(#{$row-size}n - #{$one-less-than-row-size}),
```
jasonhale: Would it be worth it to, instead of passing in 2 args, make it one argument and just calculate the "one-less-than-row-size"?

components/timePicker/hourOption.scss
```diff
@@ -1,36 +1,45 @@
...
-        border-radius: 0;
-        background-color: $color-blue;
-        color: $color-white;
-        font-weight: 700;
-
-        &:nth-child(6n - 5),
-        &.hour-option--first {
-            border-radius: $selection-radius 0 0 $selection-radius;
-            content: 'first';
+        @include breakpoint-above($above-small-breakpoint) {
```
jasonhale: Kinda piggy-backing on another comment about variables... just calc this variable instead of making a new one to keep track of?  Just a suggestion.  It ain't breakin' nothin'.

components/timePicker/hourOption.scss
troyblank: How do you find a length of a prop in css though? I am all for doing this if it's possible.

PR #656, opened 2018-07-26
TRW-1507: Coloring Sentiment Selector

jasonhale:
- Modified RadioGroup to allow 'sentiment' as a _type_
- Added colors to styles of RadioGroup for 'sentiment' type
- Modified IconSentiment to use `fill="currentColor"` as default.
- Modified timeline to utilize new colored IconSentiment
- Update to component library examples


mathom:
[99% (-0.79%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/1/)

assets/js/components.js
```diff
@@ -206,6 +206,7 @@ renderHtml(
       disabled={true}
     />
     <br />
+    <h2>{'RadioGroup - \'pill\' style'}</h2>
```
troyblank: This being a string literal seems inconsistent; the others are not also 'RadioGroup - default value set' in this pr is not as well. Is there a reason for going this way? If so let's talk about moving that way and if it should be done soon.

mathom:
[99% (-0.79%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/2/)

assets/js/components.js
jasonhale: Because of how React handles quotes, had to use string literal.  Not ideal, true.  But wanted to keep quotes for more definition in that `'pill'` was a type, as opposed to a general description.

Open to suggestions.

troyblank: Ah, gotcha. Well I personally would have just left it without quotes. But I am not great with fancy wordage so if you want to keep the quotes this makes sense. Thanks for explaining it.

components/icons/sentiment.js
```diff
@@ -28,49 +28,50 @@ function largeIcon(status) {
...
         />
       </svg>
     );
 }
 
-function mediumIcon(status) {
-    let mouth, color, backingColor, title;
+function mediumIcon(status, color) {
+    let mouth, iconColor, backingColor, title;
+    const customBackingColor = (color && color.indexOf('#')) ? lightenDarkenColor(color, 50) : color;
```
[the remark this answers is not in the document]
jasonhale: the `lightenDarkenColor()` function is to modify a custom color being passed into this component.  It doesn't modify our standard color palette, it just allows colors other than the default red/green/grey to be used on the medium-sized sentiment icons if needed and a lighter version of the custom color to be used as the background fill.  Otherwise the 'medium' sentiment icon - if it were to allow custom colors - would have to have 2 colors passed (fill and background) causing more complexity.

This feature was only added to the medium sentiment icons to keep consistent with the ability for the sentiment icon to have _any_ color as implemented into the 'large' sentiment icons... which falls back onto our conversations before about allowing svg icons to default to the text color of their container (specifically `fill = 'currentColor'`).

In all honesty, having the medium and large sentiment icons in the same component with different types of presentation made this a bit problematic.  This was the course I thought would be best without bringing in breaking changes.

Granted, I should've brought it up sooner with ya.  If you have a preferred approach, let me know and I'll pivot these changes.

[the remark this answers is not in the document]
troyblank: Yea, I would definitely prefer to have these new colors documented and cataloged. We can definitely use these lighter sentiment elsewhere and if or when we do I could love to have only 3 lighter versions and not 6, then 9 if you get where I am going with this :D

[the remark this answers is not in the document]
jasonhale: Side note... this may all be moot since Casey has a change for sentiment icons coming soon. ¯\_(ツ)_/¯

mathom:
[99% (-0.79%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/3/)

components/form/radioGroup.js
```diff
@@ -51,7 +52,14 @@ export default class RadioGroup extends PureComponent {
             )}
           >
             {values.map(v => (
-              <label key={`${v.value}-${this.state.value}`} data-tip={getDataTip(v.value, isPillStyle)}>
+              <label
+                key={`${v.value}-${this.state.value}`}
+                className={classNames(
+                  { [styles['sentiment--positive']]: ('positive' === v.value && isSentiment) },
```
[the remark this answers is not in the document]
jasonhale: Eh.  Fair.  Changing.

mathom:
[99% (-0.79%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/4/)

mathom:
[100% (-0.03%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/7/)

components/icons/sentiment.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
-import { black, mutedGreen, mutedRed } from '../util/colors';
+import { black, mutedGreen, mutedRed, subtleGreen, subtleRed, greyLight } from '../util/colors';
 
-function largeIcon(status) {
+function largeIcon(status, color = 'currentColor') {
```
troyblank: Once tiny thing left; it appears this default is not covered in testing.

assets/js/colors.js
```diff
@@ -22,11 +22,14 @@ renderHtml(
     <ul>
       <li className={styles['color-green']} data-name={'green'} />
       <li className={styles['color-muted-green']} data-name={'muted-green'} />
+      <li className={styles['color-subtle-green']} data-name={'subtle-green'} />
```
troyblank: This is not a show stopper, but this appears to break the wonderful order I had with colors :) Granted I flipped the order as sentiments are light to dark not dark to light. which now I feel silly about. Anyway no change here if not desired just expressing my intent to be anal-y organized.

See blue order for example:

screen shot 2018-07-27 at 2 41 09 pm

components/icons/sentiment.js
jasonhale: 👍

mathom:
[100% (-0.03%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/8/)

mathom:
[100% (0.0%) vs master 100%](https://jenkins.topbox.io/job/Topbox/job/rci-components/job/PR-656/9/)

PR #659, opened 2018-07-28
TRW-1410 - mobile styles for date helper, helperSwitch mobile updates

tVeratti:
I took some first steps toward using breakpoints that fit the component instead of devices. This was particularly important for the date picker, because standard device sizes still allowed some ranges where the component was extremely squished and unpleasant.

I increased the minimum size for the searchbar to flip into mobile mode for two reasons:
1. The only helper that would render nicely in the current minimum was a simple text, and there was very little room even for that.
2. The Date picker would (because of the left-hand content) not fit on the screen. If you want to see what I mean you can change the breakpoint size var back to tablet and see the small gap where things are rendered for a full screen and don't fit.

I noticed an issue with the `switchHelper` where the actual helper wasn't taking full space, and therefore the component within was being extremely squished unnecessarily. To fix this, I force the popout to take the full amount of space. The issue was the `__value` piece that was pushed out - so I allowed this to position itself over the top of the blank container with `absolute`.


troyblank:
noticed this weirdness

screen shot 2018-07-25 at 10 14 15 am

PR #661, opened 2018-07-30
TRW-1580 - added number picker to new duration picker

troyblank:
This is not the full duration picker; but rather it's heart and soul a number picker to be used by the mother component.


components/durationPicker/numberPicker.js
```diff
@@ -0,0 +1,75 @@
...
+    state = {
+        value: 0
+    }
+
+    onBlur = (e) => {
+        this.updateValue(Number(e.value));
+    }
+
+    onAdd = () => {
+        const { value } = this.state;
```
tVeratti: Just a thought since you've already created a function they can both call - what if you passed a `delta` value to `updateValue` instead?

With that you could remove even more repeated code and handle within `updateValue`

`this.updateValue(-1)`
`this.updateValue(1)`

Then:
```
updateValue(delta) {
    const { value } = this.state;
    this.setState({ value: this.constrainValue(value + delta) });
}
```

tVeratti: And you can do the callback `setState` but I never really worry about that for 99% of cases.

components/durationPicker/numberPicker.js
```diff
@@ -0,0 +1,75 @@
...
+
+    updateValue(value) {
+        this.setState({ value: this.constrainValue(value) });
+    }
+
+    constrainValue(value) {
+        const { max } = this.props;
+        let contrainedValue = value;
+
+        if (0 > value) contrainedValue = 0;
```
tVeratti: Alternatively:
`constrainedValue = Math.max(0, value);`
`constrainedValue = Math.min(max, constrainedValue);`

I don't think there's really much of a difference other than avoid `if` statements, which isn't a stylistic thing (fewer branches of code)- which I'm not really worried about but wanted to toss out there for consideration.

tVeratti: PS `constrained` missing an `s` :)

components/durationPicker/numberPicker.js
```diff
@@ -0,0 +1,75 @@
...
+
+    render() {
+        const { value } = this.state;
+        const { max } = this.props;
+
+        return (
+          <div className={styles['number-picker']}>
+            <button
+              className={classNames(styles.button, {
+                  [styles['button--active']]: 0 !== value
```
tVeratti: I'd suggest going with `0 > value` so that if somehow the number ends up below 0 (maybe future requirements make it possible, or who knows) then our button is still disabled properly.

components/durationPicker/numberPicker.js
troyblank: @tVeratti  not sure this will work, if I make `updateValue` use a delta then the input would need a separate flow when the input changes. My goal here was to have a single place that the parent would get notified of an update based on any of the three forms of input. How would you handle the input given the use of delta?

tVeratti: Yep actually I didn't think of the direct input value. You're right that this wouldn't fit that approach so feel free to discard the above.

components/durationPicker/numberPicker.js
troyblank: Yea I agree avoiding `if` is and improvement. Also fixed typo >< changed.

components/durationPicker/numberPicker.js
troyblank: Yea this way is more ironglad, changed.

PR #666, opened 2018-07-31
TRW-1584 - remove iconPopout breakpoint

tVeratti:


troyblank:
screen shot 2018-07-31 at 2 50 03 pm


looks like we have horizontal scroll bars with this :(

PR #668, opened 2018-08-01
TRF-1602 - updated loading styles on universal search to play well with main app styles

troyblank:
The loading graphic was off center in main app so I removed positioning from the loader in universal search. In an ideal world this would not have been an issue post full react app. But alas roll with the punches.


troyblank:
The tests that are failing are also failing on develop FYI.

troyblank:
@tVeratti just merged in your fix, if jenkins is cool with this i will merge this in since you already approved just FYI.

PR #671, opened 2018-08-01
TFW-1609 - made cell styles for universal search a prop

troyblank:
This will allow us to utilize space a lot better on RCI's dynamic table labels.


troyblank:
@tVeratti  yea exactly. I can't make them dynamic without data, so what this does is allow RCI to pass in the cell styles along with the data itself. Components do not mess with that sorta thing so this is 1 of 2 steps.

PR #672, opened 2018-08-01
TRW-1611 - Allow results cells to update sorting (pre: sub-menu)

tVeratti:


components/table/row.js
```diff
@@ -49,6 +49,18 @@ export default class Row extends PureComponent {
             });
         }
 
+        menu.push({
```
tVeratti: These two inserts are semi temporary. These will move into their own sub menu.

PR #674, opened 2018-08-02
TFW-1617 - fixed pending button styles

troyblank:
Oops, messed up some styles when adding BEM to preloaded. Easy fix.


components/preloader/preloader.scss
```diff
@@ -52,6 +52,7 @@
     &.preloader--light {
         &:before {
             border-top: 2px solid $color-white;
+            background-color: transparent;
```
tVeratti: Out o curiosity, why background-color instead of background: none?

troyblank: Because it is only the `background-color` that needs to be changed. If you look at the normal `light` style it is what is added.

PR #677, opened 2018-08-06
TFW-1638 - even-ed out padding in universal search bar

troyblank:
Per convo with Casey, made bottom padding of search bar even with top.


components/queryNodeList/queryNodeList.scss
```diff
@@ -2,7 +2,7 @@
 
 .list {
```
tVeratti: While you're here, can you make this BEM?

troyblank: sure thang chicken wang. Done! I updated it so it is not longer `list` which seemed to generic if we want to avoid conflicts post css modules and went with `query-list`.

tVeratti: exactamundo my thought process

PR #679, opened 2018-08-06
TFW-1570 - Add decimal places to Summary Form Scores

jasonhale:


test/components/interactionEditor/actionPane/formPane/formPane.spec.js
```diff
@@ -131,7 +131,7 @@ describe('Form Pane', () => {
         );
         const percent = wrapper.find(`.${styles.score} span span`).at(2).text();
 
-        assert.equal(percent, '(50%)');
+        assert.equal(percent, '(50.00%)');
```
troyblank: So question, if the score comes out to 50%, do we really want to show `50.00`. What is the point of the zeros in this situation?

jasonhale: Yes.  Per the ticket and convo with Blake (wrote the ticket), the preference is to always show 2 decimal places.

PR #682, opened 2018-08-07
TFW-1649 - updating mockData / demo to match popular changes

tVeratti:
I have these change ready locally on RCI as well, but wanted to make sure it looks good here since I'm having issues testing with the API.


assets/js/demo/components/universalSearch.js
```diff
@@ -25,13 +25,15 @@ export default class UniversalSearchDemo extends Component {
...
-            { label: 'Sentiment', type: ['text'] },
-            { label: 'Interaction Type', type: ['text'] }
-        ];
+            'topBoxId',
+            'interactionStartDate',
+            'callDuration',
+            'agentName',
+            'sentiment',
+            'campaignIds'
+        ].map(pD => dimensions.filter(d => d.name === pD)[0])
```
troyblank: I have been spelling these out, for example `popularDimension` instead of `pd`; as the team I was on previously was not a fan of  any abbreviation and argued readability. To me the context is already here, you can see `popularDimensions` is what is being mapped. So I am in favor of this route personally. I will try not to be so verbose going forward to match.

tVeratti: Eh yeah, honestly I'm almost always against abbreviations too, except for tiny arrow functions like this. I could be convinced to stay verbose here as well, but I love a quick & tidy arrow func

troyblank: I am all for quick tidy arrow funcs, no convincing here :D

PR #685, opened 2018-08-08
TFW-1644 - update shadows to match design for selectDropDown

tVeratti:


assets/sass/base/silents/_containers.scss
```diff
@@ -42,7 +42,7 @@
 }
 
 %shadow-box-spread {
-    @extend %spread-shadow;
+    @extend %action-shadow;
```
troyblank: action shadow!

components/selectButton/selectButton.scss
```diff
@@ -3,6 +3,10 @@
 .select-button {
     display: inline-flex;
     position: relative;
+    // Without this border radius, the shadow
+    // extends to a rectangle over the top of
+    // the buttons, leaving a tiny white ~1px triangle.
+    border-radius: $border-radius;
```
troyblank: thanks for fixing this!

assets/sass/base/silents/_containers.scss
tVeratti: It's so actiony

PR #687, opened 2018-08-09
TFW-1624 - add conditional showing of GDPR notice / styles

tVeratti:


components/auth/signIn.js
```diff
@@ -4,11 +4,15 @@ import styles from './signIn.scss';
 import { Logo, Input, Button, Notification } from '../index';
 import formValidation, { removeError } from '../util/formValidation';
 
+const GDPR_AGREE = 'GDPR_AGREE';
```
troyblank: So, we going to use this style for all local store keys?

components/auth/signIn.scss
```diff
@@ -19,10 +19,20 @@
 }
 
 .sign-in__notice {
-    display: none;
-    margin-top: 2.5rem;
-    color: $color-grey-dark;
+    margin: 2.5rem -5rem -2rem;
```
troyblank: any chance these can be added as spacing vars?

components/auth/signIn.js
tVeratti: Only if you like it. Always up for discussion.

components/auth/signIn.js
```diff
@@ -41,6 +45,7 @@ export default class SignIn extends PureComponent {
         if (hasErrors) {
             this.setState({ errors });
         } else {
+            window.localStorage.setItem(GDPR_AGREE, 'true');
```
troyblank: This seems a bit annoying for the user, I wish we could disable this forever, but alas this is not a signed in page. I can't think of a way of making it go away for good.

tVeratti: I'm sure we could if we created an api endpoint that stored this flag in the db... but that might be somewhat overkill.

components/auth/signIn.scss
tVeratti: Yeah - maybe they should also fit our 4px multiplier size thingy thang?

components/auth/signIn.js
tVeratti: Good point - and anything we'd use to track them wouldn't be any better than a fresh browser storage.

components/auth/signIn.scss
tVeratti: Ok so I added new spacing var and slightly tweaked the overall style to match our base 4px sizes.

PR #689, opened 2018-08-10
TFW-1685 - BEMify Preloader, adjust Demo props

tVeratti:


components/interactionEditor/horizontalTimeline/horizontalTimeline.js
```diff
@@ -188,6 +188,8 @@ export default class HorizontalTimeline extends PureComponent {
         const visualizationWidth = convertTimeToPixel(duration);
         const hasAudioPlayback = !isFetching && audio;
         const loaded = audio ? this.state.loaded : true;
```
tVeratti: First issue is here, for the demo at least. `loaded` is always `true` because `audio` is `undefined`.

components/interactionEditor/horizontalTimeline/timelineVisualization.js
```diff
@@ -245,6 +245,7 @@ export default class TimelineVisualization extends Component {
                 onTimeChange={onTimeChange}
               />
               { audioType && !loaded && !errored &&
+                // ISSUE: !loaded is always false
```
tVeratti: The above comment leads to the issue here, where `!loaded` breaks out of the checks and `<Preloader />` was never actually rendered

tVeratti: So the question is how RCI is passing these props - is `audio` undefined in RCI as well?
( `audio={get(resources, 'audioUrl')}` within `interactionEditor.js`)

troyblank: So I think the problem here is we used to only have audio interactions. Now as it stands the audio is not required. So we are going to need some way to determine the loading graphic without audio, It probably should just be based on any data being loaded at all for the whole interaction as this horizontal bar is used for everything.

tVeratti: Hmm ok that is useful information. But isn't the issue showing for audio interactions too?

troyblank: Yes correct there is definitely something wrong also with audio interactions and showing the preloader graphic.

PR #692, opened 2018-08-15
TFW-1722 - render Icons by name in cells

tVeratti:
Hear me out.

We do some mapping over data to prepare it for the table which is fine, but I really do not like the idea of requiring React and using JSX in our reducers. They should be completely isolated from the presentational code - this way we can keep things basic JS and still render any icon with a string.


assets/js/mockData/searchResults.js
```diff
@@ -17,11 +14,11 @@ export const queryResults = {
     offset: 0,
     rows: [
         {
-            cells: ['IKGac37144612810', 'Smith, Katrina', '06/04/17 - 6:37 am', <IconFlag color={'blue'} key={'flag-1'} />, 'Sentiment', 'Johnson, Erik', '06/04/14 - 6:37 am'],
+            cells: ['IKGac37144612810', 'Smith, Katrina', '06/04/17 - 6:37 am', 'IconFlag', 'Sentiment', 'Johnson, Erik', '06/04/14 - 6:37 am'],
```
troyblank: Talked to you over PM, but just insure we don't look for any old value called `'IconFlag'` as other values could have this theoretically.

PR #701, opened 2018-08-21
TFW-1632 - Create submenu functionality for SelectDropDown

jasonhale:
Added capability for SelectDropDown to house nested submenus.

In the process of adding submenus, the SelectDropDown return was changed from the index of the Option selected to the value to lessen the logic required for interpreting the selection.


components/selectDropDown/selectDropDown.js
```diff
@@ -157,25 +285,43 @@ export default class SelectDropDown extends PureComponent {
...
-        );
+          options.map((option, i) => {
+              const icon = (option && option.icon) ? option.icon : null;
+              const label = get(option, 'label', option);
+              const action = get(option, 'action', label);
+              const subMenuOptions = get(option, 'options', []);
+              const index = `${indexPrefix}${i + indexOffset}`;
+
+              // limiting nested submenus to 1 level deep (`&& 2 >= prevRef.length`) due to rendering position intricacies.
+              // address in future work.
```
jasonhale: Limited this to a single level deep of nested menus due to positioning (otherwise they render fine).
I'll look into it a bit more before the PR is closed.

components/table/cell.js
```diff
@@ -48,6 +47,12 @@ export default class Cell extends PureComponent {
         default:
             break;
         }
+
+        // When using keyboard navigation, selection doesn't always close the SelectDropDown.
```
troyblank: I am confused by this comment, shouldn't we always close the navigation after a selection keyboard or not keyboard? I don't think the comment is necessary but maybe I don't fully understand.

components/selectDropDown/selectDropDown.js
```diff
@@ -157,25 +281,43 @@ export default class SelectDropDown extends PureComponent {
...
+                  action={action}
+                  index={index}
+                  selectedIndex={selectedIndex}
+                  focus={itemFocus}
+                  itemSelect={onItemSelect}
+                  ariaID={ariaID}
+                  key={label + index}
+                  subMenu={subMenu}
+                  subMenuPlacement={subMenuPlacement}
+                  parentLayoutInformation={this.layoutInfo}
```
troyblank: I know this is the way I have done it in the past and this does follow other components, so you did what I would have done. But `parentLayoutInformation` is probably not needed here. @tVeratti had a great idea where the child should be able to do all these calculations. I think I update one component with this new logic but can't remember what it was.... anyway something we should talk about since I know it was a pain point prior.

components/selectDropDown/selectDropDown.js
```diff
@@ -157,25 +281,43 @@ export default class SelectDropDown extends PureComponent {
...
+          options.map((option, i) => {
+              const icon = (option && option.icon) ? option.icon : null;
+              const label = get(option, 'label', option);
+              const action = get(option, 'action', label);
+              const subMenuOptions = get(option, 'options', []);
+              const index = `${indexPrefix}${i + indexOffset}`;
+
+              // limiting nested submenus to 1 level deep (`&& 2 >= index.toString().length`) due to rendering position intricacies.
+              // address in future work.
+              const subMenu = (subMenuOptions && (0 < subMenuOptions.length) && (2 >= index.toString().length)) ? (this.renderOptions(subMenuOptions, 0, index)) : false;
```
troyblank: I am a bit confused about this; why limit it? I think it might be better just to reference only the first + second level deep and then maybe make a comment on the component about how it only supports 2 levels. Unless I am misunderstanding something this steps seems superfluous. Here I would just reference subMenu as the 2nd level and move on for now.

components/selectDropDown/selectDropDown.js
```diff
@@ -66,34 +73,112 @@ export default class SelectDropDown extends PureComponent {
...
+            itemFocus[itemFocusDepth] = 0;
+        } else if (0 > itemFocus[itemFocusDepth]) {
+            itemFocus[itemFocusDepth] = lastItemFocus;
         }
 
         if (oldItemFocus !== itemFocus) {
-            this.setItemFocus(itemFocus);
+            const itemFocusUpdate = itemFocus.join('-');
+            // storing as a string for easier storage and reliable reading.
+            this.setItemFocus(itemFocusUpdate);
```
troyblank: I am not very sure about this, I prefer declarative data; by putting this in a string we have lost all reference to what it means. We also have to do string manipulation / check length to extract the data as oppose to referencing an object. I think this would be better as an object, what do you think?

components/selectDropDown/selectDropDown.js
```diff
@@ -66,34 +73,112 @@ export default class SelectDropDown extends PureComponent {
...
+            } else if (0 < itemFocusDepth && 'left' !== subMenuPlacement) {
+                itemFocus.pop();
+                itemFocusDepth = itemFocus.length - 1;
+            }
+            break;
+
+        case 'ArrowRight':
+            if (hasASubMenu && 'right' === subMenuPlacement) {
+                itemFocus.push(0);
+                itemFocusDepth = itemFocus.length - 1;
```
troyblank: LIke I said before this seems like a strange way to figure out depth, I assume the `-1` is looking for the `-` in the string?

components/selectDropDown/selectDropDown.js
```diff
@@ -66,34 +73,112 @@ export default class SelectDropDown extends PureComponent {
...
-        if ('ArrowDown' === key) {
-            itemFocus += 1;
-        } else if ('ArrowUp' === key) {
-            itemFocus -= 1;
-        } else if ('Enter' === key && onItemSelect) {
-            onItemSelect(itemFocus);
+
+        // Break up the itemFocus into an array of indices for option lookup.
+        const oldItemFocus = [];
+        prevItemFocus.split('-').forEach((item) => {
```
troyblank: This would not be needed if `itemFocus` was an object.

components/selectDropDown/option.js
```diff
@@ -1,33 +1,118 @@
...
+            }
+        };
+    }
+
+    componentWillReceiveProps(nextProps) {
+        const { index, focus } = this.props;
+        const { selectedIndex: nextSelectedIndex, focus: nextFocus } = nextProps;
+
+        // keyboard navigation doesn't trigger a hover or click to open the submenu
+        // so, watching for that comparision here.
```
troyblank: I would say that this makes sense but is not very declarative, what about making a function that does this comparison named something like `isFocusingOptionWithSubMenu` so there is no need for explanation and future developers will understand what all this logic is for.

components/selectDropDown/selectDropDown.js
```diff
@@ -116,8 +201,47 @@ export default class SelectDropDown extends PureComponent {
...
+    }
+
+    lastItemFocus = (allOptions, itemFocusArray) => {
+        let lastItemFocus;
+
+        if (1 < itemFocusArray.length) {
+            const convertedArray = this.convertItemFocusForGet(itemFocusArray);
+            convertedArray.pop();
+            const optionCheck = get(allOptions, convertedArray);
+            lastItemFocus = optionCheck.length - 1;
```
troyblank: This along with other length lookups break autocomplete 

screen shot 2018-08-21 at 8 54 53 am

jasonhale: Tested Autocomplete but hadn't seen that.  Thanks for catching.

components/selectDropDown/selectDropDown.js
jasonhale: Agreed.  Followed this path as the component was already built that way.  Could definitely be done differently.  Had started on a way very late in the process but scrapped it due to time.

troyblank: I would be fine with leaving this as is and maybe fixing them all at one time. Just had to comment since it is something that we have recognized as no as great as we want it to be :)

components/selectDropDown/selectDropDown.js
jasonhale: Perhaps an array... which is really all that `index` string is.  One problem with the array was declaring with a "-1" as the initial value, which was a recent addition.  Not real hip on the string, but was the idea I had at the time.
Not sure an key-value pared object would work, seems like there'd be manipulation there, too.  Depends on setup.

An array, where each value is the numerical index of the option focused.  Need to know the item at each level so that the submenu can be focused when `selectedIndex` is passed value (deep reference).

What kind of object are you suggesting?

components/selectDropDown/selectDropDown.js
jasonhale: The conversion of that string to an array happened above (re: unnecessary string manipulation in another comment).  At this point, `itemFocus` is an array, and `itemFocusDepth` is just a reference to the last item in that array (Option to put into focus) for for reference at bottom of this method.

components/selectDropDown/selectDropDown.js
jasonhale: Agreed.

components/selectDropDown/selectDropDown.js
troyblank: Well, seems like focus can happen at two levels right? what about a focus object that is something like `{ depth: 1,  item: 3 }` or maybe even calling out the label of what has sub-depth in it. Anyway I think it's going to end up being a lot more complex to get it to work for every scenario like the autocomplete one.

components/table/cell.js
jasonhale: I had noticed when using the enter key to select an item as it's used in the Cell component, the SelectDropDown didn't close.  Autocomplete did, but that's because it's watching for `onClick` on itself.

It was a comment to notify a change to the initial behavior.  But, yeah, I can see where it's not completely necessary.  Removing.

PR #704, opened 2018-08-22
TFW-1786 - added base workflow demo

troyblank:
This adds a workflow demo to components. It's very basic just shows default results with no filtering, or filter bar. This is to eventually get the results labels to show, nothing fancy.

I did notice it has a lot of similarities to UniversalSearch but the results display on the table is definitely going to change so I am hesitant to make any re-usable code between the two at this point.


components/workFlow/workFlow.js
```diff
@@ -0,0 +1,52 @@
...
+                <Preloader styleVariations={['no-position']} />
+              </div>
+            );
+        }
+    }
+
+    render() {
+        return (
+          <div className={styles['work-flow']}>
+            { this.getContent() }
```
jasonhale: This is fine, but what's the thought process behind abstracting out to `getContent()` rather than an inline ternary?
```
<div className={styles['work-flow']}>
  { hasResults ?
    <ResultsGrid {...gridProps} />
    :
    <div className={styles['work-flow__pending-animation']}>
      <Preloader styleVariations={['no-position']} />
    </div>
  }
</div>
```

troyblank: It will have more things, such as no results. I generally hate coding for the future but this matches universal search and is def going to have no results so that is why the switch. Otherwise I agree switch is too much.

PR #708, opened 2018-08-23
TFW-1794 - added filter bar to work flow component

troyblank:
Since I know more than me will be working on this filter bar soon, I decided to make a small step to at least add that structure before I started re-skinning the `<Select />` for campaigns on it.

I do realize the query return is very simple right now, obviously as more filters are added this will have to be stored in `<FilterBar />`'s state.


components/workFlow/filterBar.js
```diff
@@ -0,0 +1,27 @@
+import React, { PureComponent } from 'react';
+import styles from './filterBar.scss';
+import { Select } from '../';
+
+export default class FilterBar extends PureComponent {
+    onEngagementChange = (e) => {
+        const { onQueriesChange } = this.props;
+        const query = { dimensions: 'campaign', value: e.value };
+
+        onQueriesChange([query]);
```
tVeratti: Out of curiosity, what other parts of the queries do you expect this component to manage? (Like, more than engagements? Will there be another function that funnels all query changes before calling the props `onQueriesChange`? )

troyblank: So there is a few of them in the requirements, agent and date are two for sure. I think there will definitely have to be more structure on how the queries get constructed base on several inputs changing the query array. For now though this is a small step forward.

PR #711, opened 2018-08-23
TFW-1743 - added box style variant to select box

troyblank:
This currently uses native select so it does not match design one to one; made a story to support custom options as well by creating our own select but it's a NICE TO HAVE situation and should be taken on after all other capture stuff is done.


components/form/select.scss
```diff
@@ -104,3 +104,37 @@
...
+    display: inline-flex;
+    position: relative;
+    align-items: center;
+    padding: 0;
+    border: 1px solid $color-grey;
+    border-radius: $border-radius;
+    cursor: pointer;
+
+    & > label {
+        @extend %visually-hidden;
```
tVeratti: So a11y.

PR #713, opened 2018-08-27
TFW-1758 - Add Icons to Button

jasonhale:


assets/js/components.js
```diff
@@ -56,15 +58,18 @@ renderHtml(
     <hr id="buttons" />
     <h2>Buttons</h2>
     <hr />
-    <Button text={'I am button'} onClick={e => print(e, 'Button: onClick.')} />
```
troyblank: A few things here:

- I am not sure why we need `&nbsp;`? if no reason we should look into using `margin`.
- Icons are not matching color of text for the demo, I am fine we we just want to use the filter icon for now or one that does support current color, no need to change all our icons. Just so it gets the point across that the icon should follow the text color.
- the updated spacing of the icon button in the design does not match these buttons

I realize the ticket wasn't super clear about the direction in style of the button but what we need more than anything is the ability to have a button look exactly like the one in the design. This might mean our old way of doing buttons starts to change a bit, consider making the all medium sized buttons with new spacing. Or if that won't work with the rest of the buttons consider making the icon button with the new spacing. Adding the filter icon can be part of this or we can making another ticket/sub task for it.

jasonhale: > ... `&nbsp;`
Just didn't want that space to collapse.  Removing.

> Icons...
Yeah, i was going to bring that up after - about modifying the icons to allow other colors.  Now that you mention it, i don't see a "filter" icon...

Talked to Casey - styles for the redesign are still in progress.  That 'filter' button is actually a hover state, the resting state is a blue ghost button.  The blue is a secondary color, primary being orange and orange-gradient (in another ticket).  All buttons in the new style would be of that in the ticket (extra padding on sides, single icon size, etc).

So, thoughts... make a new component, or add a style variant for the color and spacing?
As far as adding the filter icon to the library - getting that from Casey in a min.  Should be a quick add.

troyblank: The filter icon is yet to be worked into the capture work. I assumed this would be part of the button story but I should have made that more clear sorry.

I would lean towards style variation rather than new component as the functionality is the same. I also talked to Casey and all capture styles are good to go, nothing is in progress as far as us waiting to do work, but in general everything is always in progress :P

components/button/button.js
```diff
@@ -25,22 +25,28 @@ export default class Button extends PureComponent {
...
           >
+            { icon &&
+              <span className={classNames(styles.button__icon)}>{ icon }</span>
+            }
             { pending &&
               <Preloader type={'icon'} styleVariations={['light']} />
             }
-            <span>{ text }</span>
+            { text &&
+              <span>{ text }</span>
```
tVeratti: I know we talked briefly about text prop vs. children and I think we could get the best of both worlds if this were:
`{ children || text }`

jasonhale: Awesome suggestion.

components/icons/filter.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
troyblank: I am seeing difference between the design and what is done here. Seems like the one here is a bit more crowded, is this a small mistake?

screen shot 2018-08-27 at 4 44 57 pm
screen shot 2018-08-27 at 4 45 08 pm

troyblank: Also notice the round corners and top and bottom padding or off by a tad.

assets/sass/base/silents/_buttons.scss
```diff
@@ -137,3 +137,24 @@
     color: $color-black;
     font-size: $font-size;
 }
+
+// New button styles for Pattern Guide 2.0
+%button-standard {
+    @extend %base-button;
+    padding: 1.2rem 3.2rem;
```
troyblank: We should use our spacing variables here.

components/icons/filter.js
troyblank: Noted Case updated icon, so the one in components is the correct one. Still outstanding is conforming the padding and radius though from my perspective.

assets/sass/base/vars/_spacing.scss
```diff
@@ -7,3 +7,11 @@ $large-spacing: 3.3rem;
...
 $vertical-medium-spacing: 3rem;
 $vertical-large-spacing: 4rem;
+
+// Base spacing for Pattern Library 2.0 - 8pt-grid based
+$space-xs: 8px;
+$space-sm: $space-xs * 2;
+$space-md: $space-xs * 3;
+$space-lg: $space-xs * 4;
+$space-xl: $space-xs * 5;
+
```
troyblank: a couple things here that cause me pause. 1. we should probably does these in `rem`s so that the space grows with the text. and 2. instead of multiplying would you be opposed to just hard coding the value? The reason I ask is we will be referencing these spaces a lot which requires mental math. I know you were trying to keep in an 8px grid but i think that is reviewable and the benefit of having these as actual values helps the developer determine what size they want a lot easier.

assets/sass/base/silents/_buttons.scss
```diff
@@ -137,3 +140,28 @@
...
 }
+
+// New button styles for Pattern Guide 2.0
+%button-standard {
+    @extend %base-button;
+    display: inline-flex;
+    align-items: center;
+    min-height: $space-xl;
+    padding: 0 $space-lg;
+    border-radius: $button-border-radius;
```
troyblank: with `min-height` and the top and bottom padding set to `0` will this look ok if the button because two lines of text? I don't think this will be responsive but haven't tested it. What about also adding top and bottom padding?

assets/js/components.js
```diff
@@ -66,6 +67,8 @@ renderHtml(
     <br /><br />
     <Button text={'Tiny button'} styleVariations={['tiny']} onClick={e => print(e, 'Button: onClick.')} />
     <br /><br />
+    <Button text="Filter" styleVariations={['button-standard', 'tertiary']} icon={<IconFilter />} /> <Button styleVariations={['button-standard', 'tertiary']}>New Tertiary Button Style</Button>
+    <br /><br />
```
troyblank: I guess I don't understand why do we need `button-standard` and `tertiary`. I would assume this only needs `button-standard`. but I might be missing something. Also what does the word Tertiary mean here in relation to the button? I get the feeling we are over complicating this style.

assets/sass/base/vars/_spacing.scss
tVeratti: I actually kinda like the multiplication, though we could multiple rems instead.

With this, we have a very concrete pattern on values that are definitely on the same base, and adjusting it down the road will be a breeze.

components/button/button.js
```diff
@@ -25,22 +25,28 @@ export default class Button extends PureComponent {
...
-                  { [styles['button--pending']]: pending }
+                  { [styles['button--pending']]: pending },
+                  { [styles['button--with-icon']]: icon }
                 )}
             onClick={this.onClick}
             type={type}
             disabled={disabled}
           >
+            { icon &&
+              <span className={classNames(styles.button__icon)}>{ icon }</span>
```
tVeratti: You can probably just reduce this to `className={styles.button__icon}` :)

components/icons/filter.js
```diff
@@ -0,0 +1,30 @@
...
+
+    return (
+      <svg
+        xmlns={'http://www.w3.org/2000/svg'}
+        width={iconSize}
+        height={iconSize}
+        viewBox={`0 0 ${iconSize} ${iconSize}`}
+      >
+        <title>filter</title>
+        { filled ?
```
tVeratti: Alternatively, your `filled ?` could set a value for `d` and then you could use that without rewriting the rest of the path.
```
const data = filled
    ? 'M2.998 ...'
    : 'M8 7.58 ...';

...
<path fill={fill} d={data} />
```

assets/sass/base/vars/_spacing.scss
jasonhale: Yep.  Can set to rems.  Though, i'd prefer to keep the multiplication in case of adjustments - especially if using _rems_.  And adding comments for explanations.
like so:
```// Base spacing for Pattern Library 2.0 - 8pt-grid based
$space-xs: .8rem;           // 8px
$space-sm: $space-xs * 2;   // 16px
$space-md: $space-xs * 3;   // 24px
$space-lg: $space-xs * 4;   // 32px
$space-xl: $space-xs * 5;   // 40px```

File: assets/sass/base/silents/_buttons.scss (hunk 6 above)
Comment (jasonhale) [reply assets/sass/base/silents/_buttons.scss]: Buttons, explained to me, should only be one-liners.  2 lines and there could be funky vertical alignment issues.  could add a `white-space: no-wrap`... ?
Or, could add minor padding so that it won't affect the height but will allow for breathing room in the edge case of multi-line text - which should work.  I'll make that change.

File: assets/js/components.js (hunk 7 above)
Comment (jasonhale) [reply assets/js/components.js]: According to the styles that I went over with Casey, the "ghost" button that is the design used for the Filter button in the ticket is called "tertiary".  There are several - unfortunately, he does not have them all documented in the Design Patterns in invision yet... they're still in Sketch.  So this is named with those future button styles in mind.

File: components/icons/filter.js (hunk 9 above)
Comment (jasonhale) [reply components/icons/filter.js]: Yep - there's several different ways to handle it.  But followed current convention used in the other icons in this instance.
If we change it, should probably change them all in another ticket for consistency.

**Comment by jasonhale** [comment]
Updated

File: assets/sass/base/vars/_spacing.scss (hunk 5 above)
Comment (troyblank) [reply assets/sass/base/vars/_spacing.scss]: cool I have been outvoted, like the comments though that helps! I am cool with this.

File: assets/sass/base/vars/_spacing.scss [hunk 10]
```diff
@@ -7,3 +7,11 @@ $large-spacing: 3.3rem;
...
 $vertical-medium-spacing: 3rem;
 $vertical-large-spacing: 4rem;
+
+// Base spacing for Pattern Library 2.0 - 8pt-grid based
+$space-xs: .8rem;           // 8px
+$space-sm: $space-xs * 2;   // 16px
+$space-md: $space-xs * 3;   // 24px
+$space-lg: $space-xs * 4;   // 32px
+$space-xl: $space-xs * 5;   // 40px
+
```
troyblank: this works! thanks for making this more clear.

PR #716, opened 2018-08-27
TFW-1748 - add onSortChange to Workflow

tVeratti:


troyblank:
@tVeratti  i have seen this jenkins error before and not sure how to fix it. I would just merged. @mathom any ideas? details goes to 404 
screen shot 2018-08-27 at 3 55 05 pm

mathom:
@troyblank it looks like that was the job triggered on the branch, not the PR. If the branch is deleted Jenkins will remove the job, breaking the link. Try https://jenkins.topbox.io/job/Topbox/job/rci-components/view/change-requests/job/PR-715/ for an example of a PR that was tested and ran that no longer works because Jenkins has culled it. Typically the commits that are merged get picked up again and ran when they're merged into long standing branches like `develop`, which will keep the build status since the job is still in the Jenkins side of things. Github stores build status with the commit so if you find a commit without a corresponding job, you'll get a 404.

troyblank:
@mathom  cool thanks for the explanation.

PR #717, opened 2018-08-27
Datepicker skin fix

tVeratti:
It's a tiny bit hacky, but only because we don't have control over the rendered html. Can't remove that text and can't use `display: none` or `visibility: hidden` without attacking the child pseudo element.

Original problem: The `react-datepicker` changed the navigation buttons from `a` elements to `button` and added text.


assets/sass/skins/reactDatepicker.scss
```diff
@@ -60,6 +60,13 @@ $datepicker-item-size: 30px;
     top: 17px;
     width: 33px;
     height: 37px;
+    // Update introduced button el and text,
```
troyblank: I guess I don't understand this comment. Why would a future developer care that an update introduced button el and text? The rest makes sense we are hiding the text. Sorry to be a pain but this comment confuses more than helps when I read it. To be nit picky I think we should remove anything talking about the past and focus on what this css does, what do you think?

tVeratti: I added this as a "why" because someone might see this and think it's broken to have the color & background the same color (invisible text). I could see this getting changed in a refactoring moment where a well-meaning future dev thinks it seems wonky - I know I would question this if I saw it, hence the comment.

tVeratti: I can remove some of the comment info and simplify it down if that would help, but I think something is needed here to explain why the color/bg are the same.

troyblank: Sorry I don't mind the comment, I think you are right info helps here. My issue is that we are explaining to a developer about a previous state of the app when I don't think it has any bearing on the current situation. By saying "Update introduced x and y" we are adding a context that does not support why the code is there. I don't think the past matters in this situation I think the  developer just needs to know we are matching a design and the text is not supposed to be visible. Do with this what you will, it confused me it might confuse others.

PR #718, opened 2018-08-28
TFW-1516 - Don't render SelectDropDown when 'hidden'

jasonhale:


test/components/interactionEditor/actionPane/classificationForm.spec.js
```diff
@@ -55,6 +56,11 @@ describe('Classification Form', () => {
             renderAs="autocomplete"
           />);
 
+        const autoComplete = wrapper.find(`.${styles['auto-complete']}`);
+        autoComplete.forEach((ac) => {
```
troyblank: Why are we doing a `foreach` here? shouldn't the options show if the main input is focused?

test/components/selectDropDown/selectDropDown.spec.js
```diff
@@ -420,4 +420,48 @@ describe('Select Drop Down', () => {
...
+            ariaID={ariaID}
+          />
+        );
+        const instance = wrapper.instance();
+        const { itemFocus } = instance.state;
+        const optionsWithLabelsDoms = instance.renderLabelsAndOptions();
+        const labels = Object.keys(optionsWithOneLabel);
+
+        /* eslint-disable react/jsx-indent */
+        assert.deepEqual(optionsWithLabelsDoms, [
```
troyblank: Enzyme has helper functions for this, you can use `wrapper.contains` https://airbnb.io/enzyme/docs/api/ShallowWrapper/contains.html

test/components/table/table.spec.js
```diff
@@ -4,7 +4,7 @@ import { assert } from 'chai';
 import { shallow, mount } from 'enzyme';
 import styles from '../../helpers/style-fakes';
 import Table, { formatPageInfo } from '../../../components/table/table';
-import { IconArrow } from '../../../components/icons';
+// import { IconArrow } from '../../../components/icons';
```
troyblank: let's remove this import if it's no longer being used.

components/selectDropDown/selectDropDown.js
```diff
@@ -381,6 +381,8 @@ export default class SelectDropDown extends PureComponent {
               </ul>
             </div>
           </div>
+          :
+          null
         );
     }
```
troyblank: This seems odd to me, but I can't put my finger on it besides for we have not done this before. Maybe @tVeratti disagrees but it seems a little hard to read to me, I would assume we would do something like in the react documentation like:

```
const {warn} = props;
  if (!props.warn) {
    return null;
  }

  return (
    <div className="warning">
      Warning!
    </div>
  );
}```

This is not something I would  request changes for, just thoughts here and is stuck out to me as a tiny bit off our current road, not to say it's a bad thing.

jasonhale: I think both options are valid.  Depends on how we define conventions/preferences.  I prefer the single return using a ternary - more explicit to where the output is and just less code.  But the if-statement and multiple returns can be easier to read and follow.

If it's preferred the multiple return route, that's cool.  I can change it.

test/components/table/table.spec.js
jasonhale: ::facepalm::

components/selectDropDown/selectDropDown.js
troyblank: I prefer consistently; so multiple return will be more consistent with the rest of the components so I would prefer that but like I said I won't stop this PR for this nit pick.

test/components/table/table.spec.js
troyblank: lol, it's why we have reviews, we all do this stuff. A little trick I do is make the comment I will eventually remove not have a space after it so that the linter bitches to remind me to remove it.

test/components/interactionEditor/actionPane/classificationForm.spec.js
jasonhale: There are multiple `<AutoComplete/>`s.  They need to be focused to get the SelectDropDown to render to count the `Option`s.  The use of `simulate('focus')` can only be used on one node at a time, so had to iterate through the AutoCompletes to trigger them individually to render their dropdowns.

test/components/selectDropDown/selectDropDown.spec.js
jasonhale: Yep.  But was testing more the fact that they are equal, not that one contains the other somewhere within it's tree.
Prefer `contains()`?

test/components/interactionEditor/actionPane/classificationForm.spec.js
troyblank: Ah yea, sorry I see that now. Ignore me :)

test/components/selectDropDown/selectDropDown.spec.js
jasonhale: Um, @Troy, that's _your_ assert I copied, homie.  😄 
https://github.com/topbox-rci/rci-components/commit/a170a34bc7348228413636b725650a7346b7ed30#diff-5cc01d921e3451b423fba879d175d224R325

regardless, i'll change it.

troyblank: @jasonhale  well somebody should have called me out on it lol ... I didn't realize it was not new code though, so don't feel like you have to change. Update to old stuff is fine.

PR #719, opened 2018-08-28
TFW-1803 - added no results splash to work flow component

troyblank:


components/workFlow/workFlow.js
```diff
@@ -48,6 +50,14 @@ export default class WorkFlow extends PureComponent {
         switch (true) {
         case pendingResults:
             return pendingAnimation;
+        case resultsAreEmpty:
+            return (
+              <Splash
+                header={'Uh oh, there\'s nothing here'}
```
tVeratti: I have always felt that messages like this should be stored somewhere all together as constants. It's a practice that has saved my buns for globalization (not that we're doing it now, but it would give us a big advantage if we do someday). It's a nice way to keep track of all of our hard-coded messages for updating and to review / ensure there is consistency.

It would be a change outside the scope of this work, but wanted to bring it up for discussion.

test/components/workFlow/workFlow.spec.js
```diff
@@ -58,10 +76,31 @@ describe('Work Flow', () => {
...
         assert.isFalse(pendingAnimation.exists());
         assert.isTrue(filterBar.exists());
         assert.isTrue(resultsGrid.exists());
+        assert.isFalse(splash.exists());
+    });
+
+    it('should render no results splash given no results', () => {
+        const wrapper = mount(
+          <WorkFlow
+            tableData={{ results: noResults, columnNames }}
```
tVeratti: I have changed this from using column names to passing the columns fully. 
https://github.com/topbox-rci/rci-components/pull/716/files

If you have any questions/concerns let me know, but this will definitely not mesh once merged :P

components/workFlow/workFlow.js
troyblank: Per our convo, we will update that using the context api. And I will let you do the honors since you are currently the resident expert 🤙

test/components/workFlow/workFlow.spec.js
troyblank: Ah yea missed this as we were doing work at the same time. Updated!

PR #720, opened 2018-08-28
TFW-1813 - adjusted SelectButton to take new output from SelectDropDown

troyblank:
Because SelectDropDown is now outputting a label and not an index it broke SelectButton, this adjusts SelectButton to take that label and run with it.


components/selectDropDown/selectDropDown.js
```diff
@@ -72,11 +72,6 @@ export default class SelectDropDown extends PureComponent {
         }
     }
 
-    onItemSelect = (value) => {
-        const { onItemSelect } = this.props;
-        onItemSelect(value);
-    }
-
```
troyblank: This appears to no longer be used, dead code branch. But let me know if you see otherwise.

PR #722, opened 2018-08-29
TFW-1760 - New Gradient button for Capture

jasonhale:


assets/sass/base/silents/_buttons.scss
```diff
@@ -149,19 +151,49 @@ $button-border-radius: 6px;
...
     &:active,
     &:focus {
-        background: $color-blue;
+        background-color: $color-orange;
+        color: $color-white;
+    }
+}
+
+%button-style-gradient {
+    transition: background-position-y .25s;
```
troyblank: this should be `$animation-time-crisp` which is the exact same time. Part of our components should also be consistent and reusable animations. I would love to some day demo these variations of animations that we have but alas all can be found https://github.com/topbox-rci/rci-components/blob/master/assets/sass/base/vars/_animation.scss

You may also consider adding a curve on it, unless this is not intended in design.

assets/sass/base/vars/_colors.scss
```diff
@@ -54,3 +54,8 @@ $color-stripe-grey: rgba($color-black, .08);
 
 // hovers
 $color-hover-darken: rgba($color-black, .3);
+
+// Pattern Guide 2.0 Colors
+$color-orange-gradient-light: #ffac23;
+$color-orange-gradient-dark: #ff8c04;
+$color-orange-gradient-background: linear-gradient(165deg, $color-orange-gradient-light 0%, $color-orange-gradient-dark 100%);
```
troyblank: `165deg` should probably be a variable. I can see where we would want consistent angels as gradients are also used on the navigation in the future as well as also used above in this same file.

assets/sass/base/silents/_buttons.scss
jasonhale: Ah.  Adding.

assets/sass/base/vars/_colors.scss
jasonhale: the only gradient in the new designs is this orange one.  Worried about abstracting out too soon... but this is small and contained so no prob.  Changes incoming.

PR #723, opened 2018-08-29
TFW-1817 - updated output of SelectDropDown to be an object

troyblank:
Updated `SelectDropDown` to output `{ value: '', label: '' }` or `{ value: {}, label: '' }`

IF `options` param is not an array of objects `SelectDropDown` will return `{ value: label, label }` where label is the value of the array index that is selected.

I updated the three things that use `SelectDropDown` and insured they will not break.

This allows us to do more complex things with `SelectDropDown`, and offloads burden on it's parent to know what the value is for the item that was selected.


tVeratti:
So just to make sure I understand these changes...

`action` will be passed as the option's value when we use actions
`value` is always the report bit, along with `label`

Right?

troyblank:
@tVeratti  `SelectDropDown` will need to take `[ { label: 'label',  value: 'value' }, ... ]` in options or `['', '', '']` and will return as stated above. You can see in the `Table` components in this PR that I had to map `action` to `value` before giving it to `SelectDropDown`.

tVeratti:
Yep that's what I was seeing - that action is now the `value`

Thanks!

PR #724, opened 2018-08-29
TFW-1761 - add FilterSide and update Button to BEM

tVeratti:


assets/js/demo/components/workFlow.js
```diff
@@ -32,6 +33,10 @@ export default class WorkFlowDemo extends PureComponent {
         }, fakeAPIReturnSpeed);
     }
 
+    onFilterChange = (filters) => {
+        print(null, filters);
```
troyblank: you should probably at least print something to let the user know a filter changed, see my other prints, just a little context is all.

components/button/button.scss
```diff
@@ -34,46 +36,50 @@
...
     }
-}
 
-// Styles implementing Design Patterns 2.0
-.standard {
-    @extend %button;
-    @extend %button-standard;
-    box-shadow: none;
+    // Styles implementing Design Patterns 2.0
+    &--standard {
```
troyblank: ah good catch, i missed this last review.

assets/js/demo/components/workFlow.js
tVeratti: Sure, I'll add some context

components/messages/messages.en-us.js
```diff
@@ -1,9 +1,14 @@
 export default {
     // General
     CLEAR_ALL: 'Clear all',
+    CLEAR_ALL_FILTERS: 'Clear all filters',
+    APPLY_FILTERS: 'Apply Filters',
     ALL_SPEAKERS: 'All Speakers',
     CUSTOMER: 'Customer',
     AGENT: 'Agent',
+    PKEY: 'P Key',
```
troyblank: pKey label is dynamic, every client has a different pKey, this might not be a good constant, what do you think?

components/button/button.scss
```diff
@@ -34,46 +36,50 @@
...
+        @extend %button-style-gradient;
+    }
+
+    &--tertiary {
+        @extend %button-style-tertiary;
+    }
 
-.tertiary {
-    @extend %button-style-tertiary;
+    &--link {
```
jasonhale: Since '--link' is based on current styles, this should probably go above the comment about the 2.0 styles.

tVeratti: Oki doki

components/workFlow/workFlow.js
```diff
@@ -21,6 +22,14 @@ class WorkFlow extends PureComponent {
         this.setState({ sort: { direction, title } });
     };
 
+    onClearFilters = () => {
+        const { onFilterChange } = this.props;
+        const { filterKey } = this.state;
+
+        onFilterChange([]);
+        this.setState({ filterKey: !filterKey });
```
jasonhale: For toggles, should use prevState since the state changes are batched.  May get some unreliable behavior.

tVeratti: Sure I'll use that. Not super concerned since this value just needs to change for the component to re-mount, but I think that's a safer bet either way.

components/messages/messages.en-us.js
tVeratti: Not a good one at all. Didn't know that :D With the story to make it dynamic, we can remove it for good.

PR #725, opened 2018-08-30
TFW-1811 - added new select component that uses SelectDropDow

troyblank:
I figured this was a good base for the `<NewSelect />` component. It allows selection using the mouse It has some work yet to do on it though.

- still needs the ability to make a selection without mouse click (like a real select)
- still needs a bit of polish on the styles, currently the option overlay is not as pretty as in the designs
- still needs error and disable states, but those are yet to be needed for it's intended home


components/form/select/select.js
```diff
@@ -0,0 +1,138 @@
...
+                isAtBottomOfWindow: false
+            }
+        };
+    }
+
+    componentWillReceiveProps(nextProps) {
+        const { options, onUpdate } = this.props;
+
+        if (!isEqual(options, nextProps.options)) {
+            const { value, label } = nextProps.options[0];
```
tVeratti: I’m not sure that this behavior would be universally wanted. Maybe check is s selected value is provided or ..?

components/form/select/select.js
```diff
@@ -0,0 +1,138 @@
...
+        this.setState({
+            optionsShowing: true,
+            layoutInformation: getLayoutInformation(this.select)
+        });
+    }
+
+    onKeyDown = (e) => {
+        e.preventDefault();
+
+        this.selectDropDown.onKeyDown(e);
```
tVeratti: This feels a bit anti pattern compared to passing props, but at the same time, passing props would be kinda ugly and lot more work. I’m kinda torn.

components/form/select/select.js
```diff
@@ -0,0 +1,138 @@
...
+              role={'button'}
+              aria-pressed={optionsShowing}
+            >
+              { text }
+              <div className={styles.select__arrow}>
+                <IconCaret />
+              </div>
+            </button>
+            <div className={styles.select__options}>
+              <SelectDropDown
```
tVeratti: How about only rendering the options when it’s open? That means fewer unused elements rendered on load. I think the only downside might a11y but if add the right roles to the button it should be covered.

components/form/select/select.js
troyblank: @tVeratti  I am all for doing it via props, but how do you pass in the event? Since this happens in real time I think the render would cause some major weirdness, but happy to try.

The reason this is not normal is because of the event needed to be passed down without causing any render issues. Props are not great for events from what I can tell, how do you know when you get two of the same event via props?

components/form/select/select.js
troyblank: The reason this exists is because with the old select the options would sometimes change when you update things as a user, for example if you select a new engagement. What this allows is the select to be "reset" based on something that needs new options being passed in.

In this case I am not sure you would want the same thing selected if it still exists, except maybe if the select options are growing or shrinking in some way right? This is worth discussing further.

components/form/select/select.js
troyblank: Ah yes, I meant to do this and I forgot as I moved to other things. Will work on this.

components/form/select/select.js
tVeratti: It would be a super annoying method of passing like the actual info/key and then acting on it if has changed. It reminds me of this note in the docs:
https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs
But to be honest, it might be worth going about it this way simply because of the complexity of managing the response to an event.

components/form/select/select.js
tVeratti: Yeah I would at least check if they had something previously selected, and if not let it have no selection instead of picking the first option arbitrarily.

troyblank: We had a convo about this, but just for documentation this is the default behavior of a `<select />` https://codepen.io/troyblank/pen/gdLNmG so I think we are good keeping as is.

tVeratti: Yep!

components/form/select/select.js
troyblank: OK I updated this so it doesn't render when not shown, you may review my a11y button roles and aria attributes. I read some stuff and I think I got it right but you might tell my otherwise.

components/form/select/select.js
troyblank: Also discussed this, at this point there does not seem to be a better way. I agree it doesn't feel right and is not ideal. But until we can find a better way of dealing with this, it might have to stay ugly for the time being.

PR #726, opened 2018-08-30
TFW-1759 - Slide In component for Capture Tab

jasonhale:
Created a slide-in panel for the Capture tab.


assets/js/demo/components/sideDrawerDemo.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+import { SideDrawer, Button } from '../../../../components/index';
```
tVeratti: index is the default file loaded if you give a folder so you should be able to leave it as `'../../../../components`

components/sideDrawer/sideDrawer.js
```diff
@@ -0,0 +1,71 @@
...
+import { IconClose } from '../icons';
+
+import styles from './sideDrawer.scss';
+
+export default class SideDrawer extends Component {
+    static defaultProps = {
+        appearFrom: 'right'
+    }
+    state = {
+        showing: this.props.showing || false
```
tVeratti: The `||` might be a bit redundant since `this.props.showing` would have to be falsey to reach that case anyway.

components/sideDrawer/sideDrawer.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    componentWillReceiveProps(nextProps) {
```
tVeratti: I have to remind myself of this a lot, but this one of the lifecycle methods React is moving away from
https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops
That has a lot of suggestions for alternatives, but seems like `getDerivedStateFromProps` might work well for this case?

components/sideDrawer/sideDrawer.scss
```diff
@@ -0,0 +1,122 @@
+@import '../../assets/sass/base/base';
+
+$slider-width: 368px;
```
tVeratti: `36.8rem` ? Not 100% sure though, should this stay fixed size on the screen even if the font outgrows it?

assets/js/demo/components/sideDrawerDemo.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
```
troyblank: on small screens the reflow of text seems to be really strange. Any chance we can update this so that the content inside is less flow-ey. I think it will look a lot nicer there is not a reflow during animation.

components/sideDrawer/sideDrawer.scss
```diff
@@ -0,0 +1,122 @@
...
+
+    .slider--left & { right: $spacing-for-shadow; }
+
+    &__wrap {
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        width: 0;
+        transition-duration: .25s;
+        transition-property: right, left, opacity, width;
```
tVeratti: Transition a transform property instead of right/left is supposed to yield better performance.

> Now that it sits on its own layer, any 2D transform, 3D transform, or opacity changes can happen purely on the GPU which will stay extremely fast and still get us quick frame rates.

Ref: https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/

components/sideDrawer/sideDrawer.js
```diff
@@ -0,0 +1,71 @@
...
+        const { title, children, appearFrom } = this.props;
+        const { showing } = this.state;
+
+        return (
+          <div
+            className={classNames(styles['slider-container'], styles[`slider--${appearFrom}`])}
+          >
+            <CSSTransition
+              in={showing}
+              timeout={250}
```
troyblank: should this be a constant somewhere? if these could match our curve and times we are using as standards that would be amazing.

test/components/sideDrawer/sideDrawer.spec.js
```diff
@@ -0,0 +1,253 @@
...
+            title={title}
+            showing={showing}
+          >
+            {content}
+          </SideDrawer>
+        );
+        const instance = wrapper.instance();
+
+        assert.isTrue(wrapper.contains(
+          <div className={classNames(styles['slider-container'], styles['slider--right'])}>
```
tVeratti: I still think, if possible on new stuff, we should aim to test for smaller pieces at a time.
A couple `find`'s on elements to make sure they render, and a couple at a time if order/parent/child relationship is super important.

@troyblank  feel free to add your thoughts. I know this big render contains check covers a LOT of bases, but I don't think it provides a ton of value that individual finds can, and it leads to a lot of extra time fixing tests with unclear errors.

troyblank: Eh I am on the fence, I have had full render contain checks save me from making a mistake, but I also see them as being a pain in the ass. I am fine either way :)

components/sideDrawer/sideDrawer.scss
jasonhale: Yeah - talked to Casey about that.  He was specific about that size.  And 100% width on smaller mobile screens. Should probably be converted to it's computed size... essentially `$space-xs * 46` - which would scale if the base font size is changed, like you mentioned.

I don't know if it *should* scale if the font size changes.  A good Casey question.

components/sideDrawer/sideDrawer.scss
jasonhale: You're right.  I had `position` on there at one point.  Dunno where that went.  Puttin' it back.

assets/js/demo/components/sideDrawerDemo.js
troyblank: Also, sorry but is the opacity fade cool with Casey? Just asking as it seems slightly cheesy to me, but I am cool with being just a hater if it's cool with him :)

components/sideDrawer/sideDrawer.js
jasonhale: Possibly.  Where would we keep constants like this? The only other place this is being used is in QueryNode.
I'm of the "rule-of-three" line of thought... shouldn't necessarily abstract it out until it's used 3 times.  But if you feel it needs it, sure, can do.

assets/js/demo/components/sideDrawerDemo.js
jasonhale: Yep.  That was a tricky bit.  Will look back into it.

components/sideDrawer/sideDrawer.js
troyblank: Yea I guess I don't subscribe to the rule of 3, that is hard to review and tell when something hits that point with so many things like this up in the air and multiple people working on the same codebase. Also the rule of 3 is already there if you count how many animations we have using the already made standards; Demonstrated with a simple search > https://github.com/topbox-rci/rci-components/search?q=%24animation-time-subtle&unscoped_q=%24animation-time-subtle

Right now our standards are just in css https://github.com/topbox-rci/rci-components/blob/master/assets/sass/base/vars/_animation.scss

Obviously this is JS, so we will need a way to keep them in sync, it can be a comment or what not, but I would love for us to use the same times and curves for consistency throughout the app.

test/components/sideDrawer/sideDrawer.spec.js
jasonhale: I do like smaller tests.  Was following convention for most of this.  It's definitely a large bit of code to wade through in a test file and, yes, can be a pain to debug.

The initial "should render" is a good baseline for the contains, imho.  The rest could be simplified.
I'll put it through some changes and see what y'all think.

troyblank: You can use what i did in the last PR as an example to make it simpler.

components/sideDrawer/sideDrawer.js
jasonhale: Crap.  First they say they're deprecating _context_ then build out an API for it... now deprecating a lifecycle. ha!
I can change this.  Though, we'll need to eventually go through and change the other 25 files using it.

assets/js/demo/components/sideDrawerDemo.js
jasonhale: Yep.  Again, just following convention.  Changing.

troyblank: Yea there is some old `index` references in there, sorry about that... we should squash them.

components/sideDrawer/sideDrawer.scss
jasonhale: Actually - considering the new system is based on 8px, not necessarily relative font size, the fixed size of 368px (or, $space-xs * 46, or whatever) will probably be the preferred route.

components/sideDrawer/sideDrawer.js
troyblank: awww man, i am so going to forget this a lot.

components/sideDrawer/sideDrawer.js
jasonhale: Yes.  I should have spoken clearer.  Those animation timing rules aren't centralized in JS, only in CSS - of which I am familiar now (thanks for pointing out earlier).

The whole 'rule of 3' can be done when, like we're doing, reviews are seen by everyone - and at least one person has a good handle on most everything - or, as you've demonstrated - a search through the codebase can be done.

I'll start a constants file in `/components/util` for this purpose.

components/sideDrawer/sideDrawer.scss
jasonhale: I'm actually only transitioning `width` and `opacity`... not position.  Thanks for catching that.  I modified the process halfway through.

tVeratti: Right on!

components/sideDrawer/sideDrawer.js
troyblank: Still 'rule of 3' is far too human error prone for me. We should just set anything up to be re-usable and make things simple, much like the react mentality. I don't think it takes much extra effort to put variables that can be thought of as non-specific in a central location, stuff like colors, animation time, spacing etc.

troyblank: This also comes with the benefit of keeping things consistent from the start, if we use rule of 3 a situation could happen where you use two color slightly different colors for orange, you make another slightly different orange but because they are all different the 'rule of 3' never kicks in.

components/sideDrawer/sideDrawer.scss
```diff
@@ -0,0 +1,122 @@
...
+    .slider--right & { left: $spacing-for-shadow; }
+
+    .slider--left & { right: $spacing-for-shadow; }
+
+    &__wrap {
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        width: 0;
+        transition-duration: .25s;
```
troyblank: Here is another place we need to pull from our already used times, this should be `$animation-time-crisp`

jasonhale:
- Updated tests
- added a file in `/util` for housing animation constants
 - updated other component that uses animation timing (`queryNodeList`)
- component controlled by functions only, no longer via prop `showing` - removing use of in-future deprecated lifecycle event
- animation vars added to sass files
- removed opacity transition (unnecessary)

assets/js/demo/components/sideDrawerDemo.js
troyblank: This looks a lot better, thanks for the updates!

components/sideDrawer/sideDrawer.scss
```diff
@@ -0,0 +1,119 @@
...
+    .slider--left & { right: $spacing-for-shadow; }
+
+    &__wrap {
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        width: 0;
+        transition-duration: $animation-time-crisp;
+        transition-property: width;
+        transition-timing-function: $animation-curve-crisp;
```
troyblank: thanks, this is feeling a lot more hawt now!

PR #729, opened 2018-09-05
TFW-1839 - Implement SideDrawer into Workflow

jasonhale:
Added the new SideDrawer to the filters in Workflow.
Changed the SideDrawer to a completely controlled component for better integration, and modified it a bit (added a footer, to go with the header) to work better with the design.

Also, updated the demo for SideDrawer.


jasonhale:
@troyblank - PR updated.
SideDrawer moves to window edge.  Also, moved the demo from it's own page to the components page.

components/sideDrawer/sideDrawer.js
```diff
@@ -1,40 +1,46 @@
...
+    getTopOffset = () => {
+        if (this.sliderContainer && this.sliderContainer.current) {
+            const windowWidth = window.innerWidth;
+            const sliderRect = this.sliderContainer.current.getBoundingClientRect();
+            const offset = (windowWidth > this.breakpointSmallWindow) ? sliderRect.top : 0;
+            return `${offset}px`;
+        }
+
+        return 0;
     }
```
troyblank: I guess I don't understand why `getTopOffset` is needed. If you have a container that is relative `top: 0` will anchor it to the top of that container. In the demo without this it works great without this function, also if we use it for work flow it can be placed somewhere were it will make sense DOM structure wise so it has the right `top`. Is there a reason for this that I am not seeing?

components/sideDrawer/sideDrawer.js
```diff
@@ -1,40 +1,46 @@
...
+    constructor(props) {
+        super(props);
+
+        this.sliderContainer = React.createRef();
+        this.sliderWrap = React.createRef();
 
-    open = () => {
-        this.setState({ showing: true });
+        // below value referenced from ./sideDrawer.scss
+        this.breakpointSmallWindow = 378;
```
troyblank: A couple things here, why can't css handle this? Also if this is really needed we should put it in a central location to so both use the same value.

assets/js/demo/components/sideDrawerDemo.js
```diff
@@ -12,28 +12,31 @@ export default class SideDrawerDemo extends Component {
...
+    sideDrawerClose = () => {
+        this.setState({ showDrawer: false });
+    }
 
     render() {
         const { showDrawer, appearFrom } = this.state;
 
         return (
-          <div>
+          <div style={{ position: 'relative' }}>
```
troyblank: We have demo styles; we can move this into a structure similar to normal components with a sass file.

troyblank:
Here is something I am seeing with the hard coded `top`. If we get this to be css I feel like the breakpoint and top management comes for free, although it might mean adding the `ref` back.

screen shot 2018-09-06 at 11 11 23 am

components/sideDrawer/sideDrawer.js
jasonhale: Per Casey, the bottom of the drawer should be at the bottom of the window - the button for "Apply Filter" should always be visible (except for mobile, can scroll into view).
To do this, I chose to go the route of `position:fixed` instead of `absolute`.  Since `fixed` attaches to the browser window, the offset of it's container from the top is needed to be calculated so it doesn't overlap the navigation.

troyblank: But with this route you CAN get it to overlap the navigation without always listening for window resize is my point. Also it's unnecessary to calculate if you let css do the work for you. Regardless of if it's `fixed` or `absolute`. I still do not see why we need to set this with javascript.

components/sideDrawer/sideDrawer.js
jasonhale: The js needs to compare it's width to the window to set the top offset correctly.
True - could set via CSS in the media query with an `!important` to override the offset that's being placed.

What's the preference here?

troyblank: I guess I don't see why `!important` would be needed. What I would do is set the width in css and then change it to `100%` when it his our small size. 

I think `width` and `top` can all be managed with css with a slight change in placement of the drawer, and having a sibling control it rather than a parent.

As far as I see it there is two routes to remove this complexity.

I understand the navigation bar is an issue, but if you have the drawer sit below the navigation  and give it padding on top it will work great given our current navs height is static... granted we need to update the demo one to also be that way. This might be the better way since nav is in sencha and the rest is in react.

Or a more iron glad way that would roll with any top nav bar changes would be to  change in structure you could easily have a top nav bar and the drawer sit at the same level and thus always have the bottom sit at the bottom of the window.

I don't think we should not have to measure and monitor the dom to get this drawer to sit where we want it, let me know though if there is an obstacle I do not see.

troyblank:
Not fully sure why but this change added scroll bars to the bottom of this demo.

screen shot 2018-09-06 at 11 58 43 am

components/sideDrawer/sideDrawer.js
jasonhale: It wasn't overlapping the navigation when i was working on it earlier.  May I ask in what situation you're seeing that?

With absolute positioning, the design requirement wasn't able to be met because of the css being used for the containing layout of the pages.  The html containing the page contents were using 100vh with content overflowing.  Scrolling was possible but absolutely positioned elements followed the container with the height:100vh off the screen.  The decision to go with fixed position wasn't taken lightly.

This may be that the layout being used on for the demo doesn't reflect what's being used in Production.  If that's the case, then, yeah, can probably go the absolute positioning route the way it was built initially.  Which I can revert back to and just force the panel to the window edge.  However, doing so may cause issue with the bottom of the panel being at the _bottom_ of the screen vs the bottom of it's container.

troyblank: the `100vh` is on purpose as it has to have graphics anchored to the bottom on an empty or welcome splash, please see universal search for examples.

The overlapping navigation is easily seen if you do any resizing, or if you change the font sizes etc. It is easy to make this look bad because it only calculates the `top` during an interaction.

I think you are thinking of this problem only in the context of this component and I think we need to take it further out. I think we can even achieve this with flex-box positioning and transitions. I don't think this has to be the way you have it now. Regardless I see many ways we can remove the complexity of placement through JS and strongly feel we should go down that road.

components/sideDrawer/sideDrawer.js
```diff
@@ -1,40 +1,49 @@
...
+    setSideOffset = () => {
+        if (this.sliderContainer && this.sliderContainer.current) {
+            // Using the width of <html> instead of window width
+            // This keeps the drawer from going past the vert scrollbar, causing
+            // a horizontal scrollbar to appear.
+            const windowWidth = document.querySelector('html').clientWidth;
+            const sliderContainerRect = this.sliderContainer.current.getBoundingClientRect();
+            const offset = `${sliderContainerRect.left - windowWidth}px`;
+            this.setState({ offset });
+        }
```
troyblank: I suggest we remove this offset function; this shows us that we need to move this component up so we don't have to do this calculation. I am fine with merging this ticket after this change, so long as we understand the placement of this slide in needs to change in order to fix this vertical scrollbar issue.

PR #730, opened 2018-09-05
TFW-1861 - new select open width now compensates for button padding

troyblank:
Polishing width of an opened select; only other work I see to do on this is to button down the triggers to open as well as close the options.


components/form/select/select.js
```diff
@@ -81,7 +81,10 @@ export default class Select extends PureComponent {
     }
 
     getOptionsWidthCalculation = (optionsWidth) => {
-        this.setState({ optionsWidth });
+        // Because padding is in rems we must calulate padding on the fly
+        const rightPadding = parseFloat(window.getComputedStyle(this.button, null).getPropertyValue('padding-right'));
```
tVeratti: This is a pretty big string of code. Can we break it down a tiny bit for readability?

components/form/select/select.scss
```diff
@@ -35,6 +35,7 @@
     right: $space-md;
     margin-top: 1px;
     color: $color-blue;
+    z-index: 11;
```
tVeratti: This brings me back to our previous `z-index` discussions. @jasonhale mentioned the idea of storing these as variables (I think in a roundabout way with the article link) and that seems smart to me.

Maybe for local indexes where it's leveling it above its peers isn't necessary, but still worth getting on the radar since we use vars for things like spacing and color as well.

troyblank: Yea I agree I will see what I come up with, but these are meant to be above the select drop which can easily be separated into two vars.

components/form/select/select.js
troyblank: Yup! I will make this and the other change in the AM and holler when they are done.

troyblank:
@tVeratti I updated this PR with both comments. I carved out two main levels I feel like we had in our app currently, `z-index-overlay` and `z-index-modal`; Let me know if you have better names, I also provided a `z-index-base` so those components that have things that are a z-index of 1 or 2 can use this one. I put my current select work on the overlay level, and our two major notification in your face like components on the modal level. I did not change all z-indexes as i didn't want to produce a lot of testing and conflicts, if we want to do that soon I advise making a new story.

PR #731, opened 2018-09-05
TFW-1752 - reduced Workflow handlers down to filters/campaigns

tVeratti:
Also, default campaign is now passed as null so that RCI can filter it out of the queries rather than sending an empty value to the API.

Related to: https://github.com/topbox-rci/RCI/pull/908


components/workFlow/filterBar.js
```diff
@@ -21,9 +26,11 @@ export default class FilterBar extends PureComponent {
               name={'campaigns'}
               options={options}
               styleVariations={['box']}
-              onChange={this.onEngagementChange}
```
troyblank: oops, thanks.

components/workFlow/workFlow.js
```diff
@@ -23,9 +23,7 @@ class WorkFlow extends PureComponent {
     };
 
     onClearFilters = () => {
-        const { onFilterChange } = this.props;
```
troyblank: I have been trying to be consistent on how we are using props; seemed like we were always using destructuring no matter what, but it appears here since there is only one use of props we are not. This makes sense to me, but is that the rational here? Want to make sure I stay consistent going forward.

tVeratti: Here's how I make the decision:

Destructuring is a nice clean tool for basically short-handing or pulling vars out for future use. Its use case is keeping the code dry and readable. However, there are times where it is an overuse to destructure for the sake of destructuring, and that's when it doesn't actually save any duplication. So here, props are only referenced once, and same with the handler. Destructuring doesn't actually have a benefit in this case.

troyblank: Cool that's what I thought, thanks!

PR #733, opened 2018-09-06
TFW-1868 - fixed issue where focus gets lost via keyboard on select drop down

troyblank:
This smooths out keyboard focusing on select drop down options. Before this there is a a time when keying' up that totally loses focus and no options are selected, given that `-1` is a valid array option for left and right keys in this situation I decided it was best to only check bounds on up and down arrow keys.

There is already tests covering this.


components/selectDropDown/selectDropDown.js
```diff
@@ -133,7 +139,7 @@ export default class SelectDropDown extends PureComponent {
         }
 
         const lastItemFocus = this.lastItemFocus(allOptions, itemFocus);
-        if (isASubMenu) {
+        if (shouldConstrain && isASubMenu) {
```
tVeratti: So when `isASubMenu` is false it will never get constrained, right?

troyblank: Correct, sub menu's rely on a -1 to know they need to hide, so this constraint should only happen on submenu. This could probably be refactored like everything as I found it a tad confusing when I was going through all this. But this is the shortest route to fix the bug. It works great though if you give it a spin :)

tVeratti: awesome, thanks. Just making sure the intention is matching what I see :)

PR #736, opened 2018-09-11
TFW-1821 - Modify DatePicker v2 to fit container

jasonhale:


components/datepicker/datepicker.scss
```diff
@@ -1,36 +1,44 @@
 @import '../../assets/sass/base/base';
 
+// Calendars (./calendar.js) when rendered are max 294px tall due to current styles.
+// After 2.0 styles implemented, should consider implicitly setting this via vars in Calendars and imported here.
+$calendar-max-height: 294px;
```
troyblank: Curious; why do we need a max height at all? Can the content not dictate the height?

jasonhale: Standard size has two calendars side-by-side.  In smaller containers, there is only 1.  To hide the second calendar in smaller sizes, it falls below (wrapping via flex) the top calendar and gets hidden since the container has a fixed height and `overflow:hidden`.  If the height wasn't fixed, the calendars would visibly stack.
Couldn't come up with a cleaner solution to remove or hide the second calendar without using js.  (and, unfortunately, container-queries don't necessarily exist)

troyblank: Ah I guess I didn't realize that's how that worked. Do you or @tVeratti think that's an issue though? That you can select buttons that are 'off the screen'? Assuming you can keyboard select days. This is not a request for change but more of a thought to improve for the future. Anyway just food for thought.

jasonhale: Can't currently interact with the DatePicker v2 with the keyboard.  If at some point we introduce keyboard interactions for it, then yes, this solution may not work.

PR #739, opened 2018-09-12
TFW-1766 - Workflow metadata added to FilterBar

tVeratti:
Introduced a new component: `Statistic`
This can be used for any and all future rendering on presentational count/label combinations!


components/statistic/statistic.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.statistic {
+    display: inline-block;
+}
+
+.statistic__count {
+    margin-bottom: $itty-spacing;
+    color: $color-black;
+    font-size: 1.6rem;
+    font-weight: 600;
```
troyblank: We are not loading this font weight https://github.com/topbox-rci/rci-components/blob/master/assets/sass/partials/_typography.scss#L1 And I would argue for performance we should keep to 3 fonts which we are maxed out at.

tVeratti: We aren't loading the weight but it does do artificial bolding. I can check in with Casey, though I demo'd this boldness originally and he approved it.

troyblank: Yea but the faux bold is never as good as using the real font; the picky designer in me would say never to use it.

PR #746, opened 2018-09-19
TFW-1945 - allowed select to have scroll bars

troyblank:
This allows the new select to work if the options are long enough they need a scroll bar. I had to shuffle some things around so this looked o so nice.

screen shot 2018-09-19 at 10 25 28 am

There is some stuff still yet to do on this select; such as a separator if you scroll on the options so it doesn't look like the options go into nothing; which Casey and I talked about yesterday. Also if there is no default option selected the styling is weird and needs to be fixed. Past this we have new styles for categories as well as an active option.


components/selectDropDown/selectDropDown.js
```diff
@@ -238,18 +225,23 @@ export default class SelectDropDown extends PureComponent {
...
-                subMenu: -1
-            };
-            this.setItemFocus(focusReset);
+            this.select.style.display = 'inline-block';
+            this.select.style.position = 'static';
+            this.select.style.width = 'auto';
+
+            onWidthCalculation(Math.ceil(this.select.getBoundingClientRect().width));
+
+            this.select.style.display = display;
```
tVeratti: My only thought is for future us here, would it be better to capture these style values before changing them, and then setting them back to what they were? Just in case these change for variations or who knows what.

troyblank: I assume that is what I am doing with `const { display, position, width } = this.select.style;` and then setting them back. Am I missing what you are saying?

tVeratti: Oh wow for some reason my brain did not realize that! Haha, just kidding then, this is fantastic. Will approve!

PR #750, opened 2018-09-19
TFW-1928 - Table cell popover for truncated text values.

tVeratti:


tVeratti:
@troyblank  thanks for catching that - had made fixes for tests (a null check) which ended up short-circuiting the legit background color fetching.

Fixed!

components/table/cell.js
```diff
@@ -63,6 +63,41 @@ export default class Cell extends PureComponent {
...
+        /* istanbul ignore next */
+        // No reason to test up to 10 parnt levels when the stylesheet
+        // can't even be tested for confirmation.
+        while (rowNode.parentNode && 'LI' !== rowNode.tagName && iteration < maxIteration) {
+            rowNode = rowNode.parentNode;
+            iteration += 1;
+        }
+
+        /* istanbul ignore next */
+        // Cannot test stylesheets on mounted component.
```
troyblank: Although I agree; we could potentially mock the return of `window.getComputedStyle`, granted this is just a sanity check to insure this runs but possibly it might help just to insure something is returned. No change requested just a thought to maybe improve.

components/table/cell.js
```diff
@@ -71,24 +106,67 @@ export default class Cell extends PureComponent {
...
+        return showPopover && (
+          <div
+            className={popoverClassnames}
+            style={{ backgroundColor }}
+            aria-hidden={true}
+          >
+            {this.renderValue(false)}
+          </div>
+        );
+    }
```
troyblank: Why is this not a separate component? Just curious it seems ripe for one; but you might have already thought this through.

components/table/cell.js
jasonhale: Quick question...
would `const rowNode = parentNode.closest('li');` work instead of the `while... ` function?

tVeratti: I love the `closest` suggestions, I will definitely use it - thanks!

components/table/table.js
```diff
@@ -70,7 +70,7 @@ export default class Table extends Component {
               </div>
             </header>
             <div>
-              <div className={labelClasses}>
+              <li className={labelClasses}>
```
jasonhale: :(
non-semantic.  But I see the reason for it.

components/table/table.js
```diff
@@ -80,6 +80,7 @@ export default class Table extends Component {
                     onSortChange={this.onSortChange}
                     sortIcon={sort.title === title ? sort.direction : 'default'}
                     columnStyle={getCellClass(cellStyles[i])}
+                    isTitle={true}
```
tVeratti: I spent some time trying to compute what the background of the cell should be based on querying the computed style of the parent, but this not only was a costly process, it wasn't accurate with the updating background on hover. I opted to explicitly declare that it's a title because though it is an extra prop to maintain, I believe the tradeoff in perf and stability is worth it.

PR #751, opened 2018-09-21
TFW-1762 - Integrate Datepicker, Timepicker, DurationPicker into Input

jasonhale:
Integrated Datepicker, Timepicker and DurationPicker into Input via IconPopout.  Specifically done for new Capture (Workflow) Filter and use of Datepicker.

Utilized the structure of helpers from SearchBar in Universal Search.

Added 3 new Input fields for each of these options into the Demo.


components/form/input/input.scss
```diff
@@ -1,22 +1,23 @@
-@import '../../assets/sass/base/base';
+@import '../../../assets/sass/base/base';
 
 .input {
     text-align: left;
 
-    & > :first-child {
+    &__wrap {
         position: relative;
         padding-top: 1.4rem;
```
troyblank: Just a note for future us; when we conform everything will need to move this over to pattern 2.0 spacing.

components/form/input/inputHelperSwitch.js
```diff
@@ -0,0 +1,48 @@
...
+        fieldType: 'input'
+    };
+
+    state = { value: '' };
+
+    // These specific onChange functions allow us to
+    // avoid binding handlers on render. Type-specific
+    // logic should fit inside its own onChange func.
+    /* eslint-disable camelcase */
+    onChange_date = e => this.onChange(e.value, e.range);
```
troyblank: I assume these are not camelcase because they are not in universal search, and need to have some form of look up. I wonder if we could make this more of an object based lookup then and I wonder that now.... anyway just thoughts if they spark ideas. No change request.

components/universalSearch/searchBar.scss
```diff
@@ -135,7 +135,7 @@ $breakpoint-mobile: 970px;
     opacity: 0;
     z-index: 2;
 
-    input {
+    [placeholder^='Enter dimension'] {
```
troyblank: This seems strange to me, is there not a way to add a more specific class to these inputs so they are more descriptive. Such as `universal-search-form`or something?

components/form/input/inputHelperSwitch.js
jasonhale: That would make sense to me.

PR #752, opened 2018-09-24
TFW-1877 - filter pane now spans full window height

troyblank:
After talking to Casey the new nav will come with a sticky header, this will eventually need to sit always below that header. But for now we are going to sit this buddy at the very top of the window, which means no changes on RCI are needed to accommodate it.

* Allowed filter pane to take up full window height even if main content creates scroll bars.
* Allowed content of filter pane to have scroll bars if window height is too small.
* Added overlay when filter pane is open.


troyblank:
@jason a lot came out of this work though, we will need to insure we get back to a sticky button and update the design.

PR #753, opened 2018-09-24
TFW-1690 - Restrict ability to create more than 50 Cloned records

jasonhale:


components/interactionCreator/cloneForm/cloneForm.js
```diff
@@ -96,7 +96,8 @@ export default class CreateForm extends PureComponent {
 
     validateForm() {
         const formRules = [
-            { key: 'topboxID', error: 'Please set a Topbox ID', check: 'exists' }
+            { key: 'topboxID', error: 'Please set a Topbox ID', check: 'exists' },
+            { key: 'xAmount', error: 'Max 50', check: 'withinNumberRange', boundary: [0, 50] }
```
troyblank: This error is not human readable; please make the error so a user could not be potentially confused by it. I am no copy writer but something along the lines of "The maximum amount to clone is 50"

test/components/util/formValidation.spec.js
```diff
@@ -51,6 +77,20 @@ describe('Form Validation', () => {
...
+        });
+    });
+
+    it('should validate regardless of boundary order', () => {
+        assert.deepEqual(formValidation(numberValues, numberRulesInverted), {
+            [value1]: numberRulesInverted[0].error,
+            [value5]: numberRulesInverted[4].error
+        });
+    });
+
```
troyblank: What do we get if boundary does not exists or if it is a blank array?

components/util/formValidation.js
```diff
@@ -44,6 +49,17 @@ export function checkIfIsValidEmail(value) {
...
+    // allow range boundaries submit in different orders.
+    const rangeFrom = (boundary1 <= boundary2) ? boundary1 : boundary2;
+    const rangeTo = (boundary1 <= boundary2) ? boundary2 : boundary1;
+    if (value >= rangeFrom && value <= rangeTo) {
+        return true;
+    }
+
+    return false;
+}
+
```
troyblank: Question: We are already using `lodash` why not use https://lodash.com/docs/4.17.10#inRange?

Next Question why `&& value <= rangeTo`? ... it seems this code is unreachable.

components/interactionCreator/cloneForm/cloneForm.js
jasonhale: I got Casey's approval for that.  Also, longer copy will require the number field to extend to a long field - which would be awkward for requiring such a small, 2-digit number.
In this context, "Max 50" should suffice.  But if you feel it should be different we can talk to Casey or get another opinion.

components/util/formValidation.js
jasonhale: Didn't use lodash, because I'm not familiar enough with it - didn't know that 'inRange' existed.  But this simple comparison seems to work just fine.
And `&& value <= rangeTo` checks to make sure the value is between the ranges.  If we used an _or_, the value could be outside the range and still return a true.
Tests covered the instances where it was outside the range.  Though if I missed something there, please holler.

test/components/util/formValidation.spec.js
jasonhale: It'll break, since boundary would be required.  But I can definitely put in something to keep it from completely breaking.  Maybe a default to return false if a correct boundary isn't received.

components/util/formValidation.js
troyblank: That's cool, makes sense. We don't have to use lodash just curious, it's simple enough and reusable enough it makes sense to be our code.

test/components/util/formValidation.spec.js
troyblank: Yea I would at least insure there is a boundary passed in. Fine with not going crazy and checking length etc.

jasonhale: @troyblank - checks put in. give it another look over if you don't mind.

PR #754, opened 2018-09-25
TFW-1910 - update duration helper with no limit and minimums

tVeratti:
Quite a few changes to the design of the picker on top of changing the labels.

Now the maximum is constrained to the minimum, and a `no limit` checkbox disables it.


components/durationPicker/durationPicker.js
```diff
@@ -1,13 +1,16 @@
...
+import convertToSeconds from './utils';
 import TimePicker from './timePicker';
 import styles from './durationPicker.scss';
 
-export function convertToSeconds(time) {
-    return (time.hours * 3600) + (time.minutes * 60) + time.seconds;
-}
+export const ANY_DURATION = 999999;
 
 export function convertToPrettyString(time) {
```
jasonhale: This is a pure function.  Should it and the var `ANY_DURATION` be moved to `utils.js`?
We have a lot of exports in our components that aren't the component themselves.  Seems like it could lead to maintenance issues.

non-blocking comment.  just a question for my own benefit.

components/durationPicker/durationPicker.js
```diff
@@ -1,13 +1,16 @@
 import React, { PureComponent } from 'react';
 import { addLeadingZeros } from '../util/string';
+import convertToSeconds from './utils';
 import TimePicker from './timePicker';
 import styles from './durationPicker.scss';
 
-export function convertToSeconds(time) {
-    return (time.hours * 3600) + (time.minutes * 60) + time.seconds;
-}
+export const ANY_DURATION = 999999;
```
troyblank: The significance of this number should probably be commented. I know we are bending to the API, but that should be apparent to someone reading this code.

components/durationPicker/numberPicker.js
```diff
@@ -10,6 +10,13 @@ export default class NumberPicker extends PureComponent {
         value: 0
     }
 
+    componentWillReceiveProps(nextProps) {
```
jasonhale: We've talked about not using `componentWillReceiveProps` as it's on a path to be deprecated.
I'd be cool with leaving it if we can come back to change it later.  Or consider using `componentDidUpdate` for this instead.

components/durationPicker/utils.js
```diff
@@ -0,0 +1,7 @@
+
+const HOUR_SECONDS = 3600;
+const MINUTE_SECONDS = 60;
+
```
troyblank: These vars took made my brain wonder a tad instead of getting to the bizznass. Would you be opposed to making them more descriptive such as `SECONDS_IN_HOUR` ? Super nit picky I know.

components/durationPicker/durationPicker.js
tVeratti: Possibly, though it's only export to help the tests match up. I'm not against moving it.

components/durationPicker/utils.js
tVeratti: Sure, that's a better name!

components/durationPicker/durationPicker.js
tVeratti: Will add an explanation.

components/durationPicker/durationPicker.js
troyblank: considering it's super specific to the API and this I don't mind it staying here.

components/durationPicker/utils.js
tVeratti: I had an entire extra function here that convert seconds back to the a time object, which became unnecessary after a better approach. These const are kinda overkill now without it, but might be useful someday.

components/durationPicker/durationPicker.js
```diff
@@ -1,13 +1,21 @@
...
 import styles from './durationPicker.scss';
 
-export function convertToSeconds(time) {
-    return (time.hours * 3600) + (time.minutes * 60) + time.seconds;
-}
+// The API currently will not accept a range without
+// a top-end value. Therefore, this value will be passed
+// when the user wants no end limit. Until we can pass
+// a single beginning value for the range, this is how
+// we will trick the API into hevaing.
```
troyblank: is 'hevaing' a typo... behaving right? other than that all looks good

PR #760, opened 2018-09-27
TFW-1879 - Close SideDrawer with outside click

jasonhale:
This modifies the SideDrawer to close when clicking outside the slider, but within the overlay.  Note: There's some conversation on whether this is the correct user experience - waiting on final decision, PR may change.

Changes:
- SideDrawer is now controlled by functions accessed by it's ref, instead of passing in a boolean. Went this route so it could close itself.
  - Left `onClose` so a parent could be notified of closing.
- Updated the styles for SideDrawer to go with 2.0 Design updates: https://projects.invisionapp.com/d/#/console/15152772/315092349/preview
- Slightly updated `Input`'s 'box-style' to me more in line with 2.0 Design expectations.

Also modified Webpack settings for the demo to output only partially-hashed classnames to help with debugging when building new components.


components/form/input/input.scss
```diff
@@ -1,5 +1,8 @@
 @import '../../../assets/sass/base/base';
 
+$input-border-width: 1px;
+$input-top-label-spacing: $space-md; // 2.0 style
```
troyblank: do we have a plan to remove these comments? I just see a day where a developer is like, "what in the world is 2.0 style?". I am not sure it helps unless we do it to everything that is being transferred over like an inventory and have a plan to remove. But we all need to be on board for that.

jasonhale: Good point.  I thought it'd be helpful at the moment to denote which styles are implemented as part of that move.  If they're not helpful, or could cause confusion, I'll remove and won't add others unless deemed necessary.

components/sideDrawer/sideDrawer.js
```diff
@@ -1,55 +1,80 @@
...
 
-              </div>
+        this.setState({ showing: false }, () => {
+            if (onClose) onClose();
+        });
+    };
+    toggle = () => this.setState(prevState => ({ showing: !prevState.showing }));
+
+    render() {
+        const { header, footer, children, showCloseButton = true } = this.props;
```
troyblank: Do we have an instance where we do not show close button? I don't think we really need this param until it's necessary.

components/overlay/overlay.js
```diff
@@ -15,7 +15,7 @@ export default function Overlay({ showing }) {
             exitDone: styles['overlay--exit-done']
         }}
       >
-        <div className={styles.overlay} />
+        <div className={styles.overlay} onClick={onOverlayClick} />
```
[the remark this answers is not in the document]
jasonhale: Sorry, I don't follow.  `onClose` is a prop passed to SideDrawer (not Overlay), and just notify's the parent when it closes (and, as pointed out in another comment, is something not used anymore and could be removed).
Granted, `onOverlayClick` is a bit verbose and could be just `onClick`, but it's here to notify it's parent of whether or not there is a click.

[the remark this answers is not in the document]
tVeratti: It looks like it technically does tap in to the overall `onClose` if that's what you're looking for Troy. Though I'd agree with changing it to `onClick` :)

PR #761, opened 2018-10-01
TFW-1999 - check option categories on prop change in new Select (HOTFIX)

jasonhale:
New Select component breaks when new options come in via props with categories.  This is a hotfix for use in RCI 4.31 release.


jasonhale:
Test case added that covered the issue.

PR #762, opened 2018-10-01
TFW-1989 - update duration picker to use messages

tVeratti:
Have had a bit of strangeness with tests and onChange values, but it smoothed out after a few runs. Please let me know if you see anything similar on this or future stories that you work on.


tVeratti:
As a note, the errors were caused by the duration picker tests passing `chance` integers which were then being limited to the min/max values, which led to a mismatch on the reported hours/minutes/seconds afterward.

Clamped test values to be 0/0/0 (min) and 99/59/59 (max) respectively, so they don't exceed duration min/max.

PR #765, opened 2018-10-02
TFW-2029 - Set up new layout

jasonhale:
This is to create the very base of the app layout for the new design and navigation.
The idea at this point is to keep this super simple, to keep main styles and layout within rci-components, and allow the Navigation, Page Header and Page Content to be added inside RCI so that global state and routing can be utilized.

If approved, can create another ticket to import and implement this into RCI.


components/baseLayout/baseLayout.scss
```diff
@@ -0,0 +1,43 @@
...
+    grid-template-columns: auto 1fr;
+
+
+    @media (max-width: $size-tablet - 1) {
+        grid-template-areas: 'header' 'main' 'nav';
+        grid-template-rows: auto 1fr auto;
+        grid-template-columns: 1fr;
+    }
+
+    &__header {
```
tVeratti: I like this for modifiers, but I think it can be a tad confusing for nested elements. I'd suggest keeping these out as `.base__header` just to avoid confusion.

tVeratti: Of course if that seems totally sensible to you guys, I'm willing to get used to that style.

tVeratti: Also - if this is a syntax we want to keep using for nested elements, that seems to promote more scss nesting when one of the benefits of BEM is that the classnames are specific enough to prevent the need for nesting.

components/baseLayout/baseLayout.js
```diff
@@ -0,0 +1,18 @@
...
+export default function BaseLayout({ nav, header, children }) {
+    return (
+      <div className={styles.base}>
+        <div className={styles.base__main}>
+          { children }
+        </div>
+        <div className={styles.base__header}>
+          { header }
+        </div>
+        <div className={styles.base__nav}>
```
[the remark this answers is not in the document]
troyblank: Yea I agree with this, we def should use semantic elements.

components/baseLayout/baseLayout.scss
troyblank: Naw I personally don't like deep nested stuff; I would say only next if there is a beneficial reason to do so. I am with @tVeratti

assets/js/demo/components/baseLayoutDemo.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import BaseLayout from '../../../../components/baseLayout/baseLayout';
+
+import UniversalSearchDemo from './universalSearch';
+
+export default function BaseLayoutDemo() {
+    return (
+      <BaseLayout
+        nav={'Navigation'}
+        header={'Header here.  Below is an example of page content.'}
```
troyblank: Two spaces here is bothering the OCD in me but I know it's placeholder so I won't demand a change :P

components/baseLayout/baseLayout.js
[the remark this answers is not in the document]
jasonhale: I elected not to do that here, so that those elements could be used inside the specific components themselves.  Though, `main` could be done pretty safely.

If we can agree to not use those specific elements in the upcoming components, and can use them here, then I'll make the changes.  Sound good?

components/baseLayout/baseLayout.scss
jasonhale: I prefer nesting as it's quicker/easier for me to write, and allows the block to be renamed easily (without find & replace).  But it's all just opinion... so, whatever the consensus is, i'll change it. .. and looks like i'm changing it.  ;)

Though the comment _"when one of the benefits of BEM is that the classnames are specific enough to prevent the need for nesting"_ i think is misused here.  There's no nested selectors being created, only reuse of the base classname... it compiles only to 4 selectors and nothing overly specific:
```
.base {
  display: grid;
  width: 100vw;
  height: 100vh;
  overflow: hidden;
  grid-template-areas: 'nav header' 'nav main';
  grid-template-rows: auto 1fr;
  grid-template-columns: auto 1fr;
}
@media (max-width: 767px) {
  .base {
    grid-template-areas: 'header' 'main' 'nav';
    grid-template-rows: auto 1fr auto;
    grid-template-columns: 1fr;
  }
}
.base__header {
  box-shadow: 0 0 17px rgba(0, 0, 0, 0.2);
  z-index: 10;
  grid-area: header;
}
.base__main {
  overflow: auto;
  grid-area: main;
}
.base__nav {
  box-shadow: 0 0 17px rgba(0, 0, 0, 0.2);
  z-index: 10;
  grid-area: nav;
}
.base__header, .base__main, .base__nav {
  position: relative;
  background-color: #fff;
}
```

jasonhale:
> The content you used also does not work well with the current layout; For example in first screen shot the graphic does not go to the bottom, also in second screen shot the content gets cropped without scroll bars. I assume this is because the content needs to be updated but wanted to confirm you do not see an issue with this going forward.

Ah, current content container uses `flex`.  Will add that.  Though, there is still a slight vertical scroll due to some negative margins - which i've seen elsewhere to get around page padding.  We may have to make some minor modifications to some of these pages when their parent doesn't have padding.  We could introduce a `:global` selector to utilize for that if necessary.

I hid the horz overflow, but can revert that.

components/baseLayout/baseLayout.js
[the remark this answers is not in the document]
troyblank: You raise a good point here should the semantic element be with the specific component or with the base layout? I say def <Main /> but I take it back I think to have stand alone components they should probably contain the semantic element. I am cool so long as we get <Main /> in there.

troyblank:
@jasonhale  sounds good, yea we might have to have to variations of the components style so they fit in old and new layout. My comment is just to make sure we are thinking about that and it won't be hard to fit it in.

jasonhale:
@troyblank @tVeratti - y'all want to give this another look-through?  :D

components/baseLayout/baseLayout.js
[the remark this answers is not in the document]
tVeratti: In that case, we're adding quite a few extra layers of divs for very little gain.
```
<div className={styles.base}>
  <div className={styles.base__header}>
    <header className={styles.header}>
      {stuff}
```

That's at least 3 levels of wrappers before we get to any content. Maybe this is absolutely necessary - but if anything stands out to me as an area of imporovement to our HTML, it's how many layers we have to dig through (or up) to get to relevant content.

PR #766, opened 2018-10-03
TFW-2043 - added card layout toggle component

troyblank:
screen shot 2018-10-03 at 11 43 04 am

Component to be used on new Qlik pages.


components/iconButton/iconButton.js
```diff
@@ -0,0 +1,25 @@
...
+    render() {
+        const { icon, active } = this.props;
+        const variation = active ? '' : 'hollow';
+
+        return (
+          <button
+            className={styles['icon-button']}
+            onClick={this.onClick}
+          >
+            {React.cloneElement(icon, { variation })}
```
tVeratti: Interesting solution. My only question is this - does this trigger a re-render every time if the cloned element is consider `!==` ? (Similar to bound click handlers)

components/cardLayoutToggle/cardLayoutToggle.scss
```diff
@@ -0,0 +1,59 @@
...
+    justify-content: flex-end;
+    margin-bottom: $space-sm;
+
+    @include breakpoint(600px) {
+        display: none;
+    }
+
+    button {
+        position: relative;
+        margin: 0 $space-sm;
```
jasonhale: Casey mentioned in our design channel that this spacing should be 8px on either side of that dividing bar.  This calcs to 16px;  Should be `$space-xs';

components/cardLayoutToggle/cardLayoutToggle.scss
```diff
@@ -0,0 +1,59 @@
+@import '../../assets/sass/base/base';
+
+.card-layout-toggle-nav {
+    display: flex;
+    justify-content: flex-end;
+    margin-bottom: $space-sm;
+
+    @include breakpoint(600px) {
```
jasonhale: is '600px' specific to CardLayoutToggle?  Should it be a var?

components/cardLayoutToggle/cardLayoutToggle.scss
```diff
@@ -0,0 +1,59 @@
...
+
+    button {
+        position: relative;
+        margin: 0 $space-sm;
+
+        &:after {
+            position: absolute;
+            right: -$space-sm;
+            width: 1px;
+            height: 24px;
```
jasonhale: Another question if this should be a var?  (non-blocking question)

components/iconButton/iconButton.js
```diff
@@ -0,0 +1,25 @@
+import React, { PureComponent } from 'react';
+import styles from './iconButton.scss';
+
+export default class IconButton extends PureComponent {
```
jasonhale: Now that we have a component, we can work on hover states per new design.  👍

components/iconButton/iconButton.js
troyblank: After doing some reading I determine my implementation does trigger a re-render every time; so I changed it to `{React.cloneElement(icon, iconProps)}`. Which from my understanding will help with the re-renders a lot. Thanks for pointing this out.

Without `cloneElement` here the other solution is to put the burden on the consumer to supply new icons based on if the button is in an active state. I am trying to avoid this burden by using `cloneElement`.

components/cardLayoutToggle/cardLayoutToggle.js
```diff
@@ -0,0 +1,51 @@
...
+
+    render() {
+        const { layoutStyle } = this.state;
+        const { children } = this.props;
+        const isGridStyle = 'grid' === layoutStyle;
+        const isStackStyle = 'stack' === layoutStyle;
+
+        return (
+          <div>
+            <nav className={styles['card-layout-toggle-nav']}>
```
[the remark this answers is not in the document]
troyblank: @tVeratti  I updated these names as well as renamed all instances of `CardLayoutToggle` to `LayoutToggle`. I think this covers all of your concerns let me know if it does not.

components/cardLayoutToggle/cardLayoutToggle.scss
troyblank: I don't think so personally; because the size of the spacer is arbitrary from what I gather. It's variable name would probably use be `height-of-spacer`. Which seem redundant to me.

components/cardLayoutToggle/cardLayoutToggle.scss
troyblank: Yup totally, that is a mistake. Fixed!

components/cardLayoutToggle/cardLayoutToggle.scss
troyblank: Yea def should be a var, thanks will add it specific to this scss file. I also updated those in `qlikNavigationCard.scss`. Thanks!

components/iconButton/iconButton.js
troyblank: yea just did default over white right now but def can build out variations.

components/layoutToggle/layoutToggle.scss
```diff
@@ -0,0 +1,61 @@
...
+            margin-right: 0;
+        }
+
+        &:last-child:after {
+            display: none;
+        }
+    }
+}
+
+.card-layout-toggle-nav__button {
```
jasonhale: @troyblank - Straggler?  I don't see this one in the markup.

tVeratti:
Approved pending Jason's comments/questions :)

components/layoutToggle/layoutToggle.scss
troyblank: Great catch! took it out.

PR #768, opened 2018-10-03
TFW-2066 - Webpack, target removal of css-modules

jasonhale:
Modified Webpack config to allow files with `[...].beta.scss` to be compiled without hashing via css-modules.

Also, added the NavPane component into the demo to showcase this implementation.


components/navPane/navPane.js
```diff
@@ -1,12 +1,12 @@
...
-import styles from './navPane.scss';
+import './navPane.beta.scss';
 
 export default class NavPane extends PureComponent {
     state = {};
 
     render() {
         return (
-          <div className={styles.navpane}>
+          <div className="navpane">
```
troyblank: can we pretty please use `<div className={'navpane'}>` here for consistency?

test/components/navPane/navPane.spec.js
```diff
@@ -1,15 +1,14 @@
...
 import { NavPane } from '../../../components';
 
-describe('BaseLayout', () => {
+describe('NavPane', () => {
     it('should render', () => {
         const wrapper = shallow(<NavPane />);
 
         assert.isTrue(wrapper.contains(
-          <div className={styles.navpane}>
+          <div className="navpane">
```
troyblank: same here, use a js string for consistency, probably need a lint for this. `div className={'navpane'}>`

components/navPane/navPane.js
jasonhale: You're killin' me, Smalls.
This is one of the conventions in the codebase that really bothers me.  But, yeah, I'll get that in there for consistency's sake.

PR #769, opened 2018-10-04
TFW-2048 - add logo to new NavPane

jasonhale:
Added the TopBox logo icon (with the check) and a smaller version of the TopBox text logo into the component library.

Added these both to the new NavPane.

Updated colors in colors.js to match with recent changes to the sass colors variable file.


assets/sass/base/vars/_colors.scss
```diff
@@ -5,6 +5,8 @@ $color-white: #fff;
 
 // SECONDARY
 $color-blue-2: #e2f1f6;
+$color-blue-grey-1: #ced6dd;
+$color-blue-grey-2: #e6ebef;
```
troyblank: Can we add those to our rendered colors?

screen shot 2018-10-05 at 11 03 51 am

components/navPane/navPane.beta.scss
```diff
@@ -11,9 +11,31 @@ $nav-width-vert-min: 7 * $space-xs; // 56px
...
+    align-items: flex-start;
+    height: $space-xxxl;
+    padding: $space-sm $space-md 0;
+
+    .logo {
+        margin-left: $space-sm;
+    }
+}
+
 @media (max-width: $size-tablet - 1) {
```
troyblank: Sorry I didn't catch this before but we have been using `@include breakpoint` mixin for breakpoints. For consistency we should use that OR if you feel this makes things more flexible going forward let's talk about moving away from that practice. I see both ways as it starts to make sense as we are doing component specific break points. Secondly we are using `$size-tablet` which is not very component specific, do you see problems using that going forward?

assets/sass/base/vars/_colors.scss
jasonhale: fo sho'.  will do.

components/navPane/navPane.beta.scss
jasonhale: Good call.  I should've used the mixin to be consisent.

> OR if you feel this makes things more flexible
Eh.  Not so much "more flexible" as much as the breakpoint mixin doesn't really save us much repeated code.  It's pretty straightforward.  I'd be OK with not using it if you are.
There may be circumstances where the breakpoint mixin wouldn't help ... like a media query with both a min-width and max-width parameter.

Sometimes i do think components should have specific breakpoints for themselves as needed for content.  Though, I feel this navpane component doesn't fall in that category.

> very component specific, do you see problems using that going forward?
Nah - `$size-tablet` being non-component specific is ok - we should have some globally defined breakpoints.  Though, the name is misleading.  It's more of a "small screen size" since devices run the gamut of sizes.  I'd think it would almost be more like large/medium/small, like you've got set with the icons.  Though that route would involve hashing it all out with Casey and designs.

troyblank: Sounds good, yea use breakpoint mixin for now just to stay consistent, and we can talk about not using it in the future if it no longer makes sense. Other than that all is good, thanks for the thoughts.

assets/sass/base/vars/_colors.scss
jasonhale: Done.
rci_components___index

PR #772, opened 2018-10-05
TFW-2029 - Set up new layout

jasonhale:
Here is rework of setting up the base of the new layout.
- BaseLayout are no longer using css-modules.
- NavPane updated to accept 'className' as a prop.

Working with the idea to reduce _div-itis_, the base of the app would need the class name "base" added to it, with BaseLayout being a direct decendant.  Example, in RCI/nav_base: src/server/components/html.react.js:line 31: `<div id={'root'}>` would be `<div id={'root'} className={'base'}`;
This would get the main portions of the layout (header/nav/content) as high up in the DOM tree as possible without extraneous `div`s.  But still using the BaseLayout component to keep these sections in a specific order.


assets/js/demo/components/baseLayoutDemo.js
```diff
@@ -0,0 +1,18 @@
...
+
+import UniversalSearchDemo from './universalSearch';
+
+export default function BaseLayoutDemo() {
+    return (
+      <div className="base">
+        <BaseLayout
+          nav={<NavPane />}
+          header={<div>Header here. Below is an example of page content.</div>}
+        >
```
troyblank: This seems odd to me at first; I am going to give it some time to soak in. Curious what @tVeratti thinks. I am not sure of the pros and cons here of passing these going to be giant sections as props as opposed to using them as children 🤔

jasonhale: Yeah, def looks odd.  The main reason was to enforce dom order... so they can all have the same z-index but keep drop-shadows falling in the correct places.  Also, for wrapping page content in the `<main>` element, and giving everything the correct class names.  Possibly a bit convoluted.  Definitely think about it.

Otherwise could do without the BaseLayout component altogether and just use the styles... adding classnames to elements/components at the root level.

assets/js/demo/components/baseLayoutDemo.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import BaseLayout from '../../../../components/baseLayout/baseLayout';
+import NavPane from '../../../../components/navPane/navPane';
+
+import UniversalSearchDemo from './universalSearch';
+
+export default function BaseLayoutDemo() {
+    return (
+      <div className="base">
```
jasonhale: ideally, this `<div className="base">`  would be the root of the app.  for the demo, though, i thought this would describe it.

assets/js/demo/components/baseLayoutDemo.js
troyblank: After pondering this I am still on the fence, I am leaning towards just using styles and simplifying things but I see the advantage here for sure. I am going to let @tVeratti be the tie breaker since I still don't have a confident response.

tVeratti: I like it the way it is. I actually built the router section assuming BaseLayout would just render children, so this is actually important. Header/Nav as props, content (routable) as children.

PR #774, opened 2018-10-08
TFW-2090 - updated label copy on workflow

troyblank:


components/messages/messages.en-us.js
```diff
@@ -7,7 +7,7 @@ export default {
     CUSTOMER: 'Customer',
     AGENT: 'Agent',
     TOPBOXID: 'Topbox ID',
-    DATE: 'Date',
+    DATE: 'Interaction Start Date',
```
jasonhale: Question on the expectation of messages.  "Date" seems generic.  This label is specific to _interactions_... should it be it's own deal, instead?  Like: `INTERACTION_START_DATE: 'Interaction Start Date'`

Or we can elect to change it later when/if a conflict arises.

troyblank: @jasonhale good call, changed it to `INTERACTION_DATE`. Did not use "start" in the key name as there is no "end" date really and the start date is what we use to identify the interaction date. Honestly I am not sure why we use "start" in the copy at all.

PR #776, opened 2018-10-09
TFW-2097 - added basic header tab nav to base layout

troyblank:
OK, so this gets us a pretty good start for the basic tab sub navigation for our new navigation. I figured making the `<HeaderPane />` take different children is going to make sense as the design has it having to bend to many different controls. The one thing I have not super solidified is the communication between this component on the content so if you see a better way of doing it that could be part of this first step let me know.

What it is missing and have made stories for is this

- icons - https://topbox.atlassian.net/browse/TFW-2097
- selectable small size nav - https://topbox.atlassian.net/browse/TFW-2099

screen shot 2018-10-09 at 3 51 12 pm
screen shot 2018-10-09 at 3 51 21 pm


components/headerPane/tabButton.beta.scss
```diff
@@ -0,0 +1,40 @@
+@import '../../assets/sass/base/base';
+
+$size-small: 700px;
```
jasonhale: Noticed this `700px` is used in a couple spots.  Is 700 specific to the content?
Thought about moving this to the __breakpoints.scss_ so they can all be modified together?  I've done the same for the NavPane which will be in a PR shortly.

troyblank: Yea good call, but i will make it local to the headerPane folder; as we want to keep breakpoints specific to each component after my talk with @tVeratti  way back. I will def consolidate these and let you know when I have updated it.

components/headerPane/tabButton.js
```diff
@@ -0,0 +1,26 @@
...
+        const { id, onClick } = this.props;
+
+        onClick(id);
+    }
+
+    render() {
+        const { title, id, activeMenuId } = this.props;
+
+        return (
+          <button
```
jasonhale: Question on integration: Since these are in the header, outside the main content, should these be `<Link/>` instead of buttons (with page content handled via Routes)?  Or, as implemented here, an HOC would redirect on click?

troyblank: Yea I am not fully sure, I assume they may be `<Link />`s at some point.  But for now I think they are fine till we do that integration.

PR #777, opened 2018-10-10
TFW-2045 - New Nav, main section links

jasonhale:
Added new *NavItem* component.
Updated BaseLayout Demo to showcase main nav items.

Added 'color-black-2' from new design into styles and demo colors.


assets/sass/base/vars/_breakpoints.scss
```diff
@@ -3,3 +3,5 @@ $size-tablet: 768px;
 $size-tablet-small: 725px;
 $size-mobile: 600px;
 $size-mobile-small: 350px;
+
+$base-layout-breakpoint: $size-tablet;
```
troyblank: This should be put in the component css right? nothing outside of it will use this I assume. @tVeratti  and I had a convo in the past; we want to make breakpoints specific to the component and managed there so there is almost no need for this file outside of supporting older stuff.

components/navItem/navItem.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
troyblank: I don't see this component ever being used outside of the `<NavPane />`. Correct me if I am wrong but shouldn't this file be in `components/navPane` ?

components/navPane/navPane.js
```diff
@@ -1,13 +1,13 @@
-import React, { PureComponent } from 'react';
+import React, { Component } from 'react';
 import classNames from 'classnames';
-import { Logo, LogoCheck } from '../index';
+import { Logo, LogoCheck, NavItem } from '../index';
 import './navPane.beta.scss';
 
-export default class NavPane extends PureComponent {
+export default class NavPane extends Component {
```
troyblank: Why does this have to be a `Component` as opposed to a `PureComponent` ? If it's due to rendering issues; can we add a check for re-render and get the cost savings of a `PureComponent`?

jasonhale: NavPane needs to implement `shouldComponentUpdate` in to keep routing from being blocked.
Could also look into wrapping in a HOC.
https://reacttraining.com/react-router/web/guides/dealing-with-update-blocking

components/navItem/navItem.js
jasonhale: Yup.  Will move.

components/navPane/navPane.js
troyblank: yup that was what I was referring to `shouldComponentUpdate`. It would be awesome if we could keep this as a `PureComponent` so we do not get unnecessary renders. I am not familiar with the HOC way of doing things but whatever seems best.

assets/sass/base/vars/_breakpoints.scss
jasonhale: The Base Layout styles break at a specific point.  This is the same breakpoint used in NavPane and NavItem.  Seeing as how all 3 items break at the same point - and _should_ break at the same point - seemed a good idea to pull that out to somewhere central and easily updated.

If you y'all would rather we not do this, I'll remove this change and keep the breakpoints individual within each component.

components/navPane/navPane.js
jasonhale: Yeah, PureComponent doesn't utilize `shouldComponentUpdate`.  I can take a look at the HOC path recommended in the react-router docs.

assets/sass/base/vars/_breakpoints.scss
troyblank: Naw that is fair; if it's used by all of those carry on!

assets/js/demo/components/baseLayoutDemo.js
```diff
@@ -35,14 +71,23 @@ export default class BaseLayoutDemo extends PureComponent {
...
-              <UniversalSearchDemo />
-            </BaseLayout>
-          </div>
+          <BrowserRouter>
+            <div className="base">
+              <BaseLayout
+                nav={<NavPane mainConfig={navConfig} />}
+                header={header}
+              >
+                <Switch>
```
tVeratti: This is great, and the way you've done it here will translate really accurately over to RCI. Currently `betaRoutes` is a placeholder for when BaseLayout is ready.

Nice work!

PR #778, opened 2018-10-10
TFW-2101 - added create and new icons for create app subnav

troyblank:


components/icons/clone.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export default function IconClone({ filled = false }) {
+    let path;
+
+    if (filled) {
+        path = 'M12 12v.91A1.1 1.1 0 0 1 10.89 14H10v.91A1.1 1.1 0 0 1 8.89 16H1.11A1.1 1.1 0 0 1 0 14.91V5.09A1.1 1.1 0 0 1 1.11 4H2v-.91A1.1 1.1 0 0 1 3.11 2H4V1a1 1 0 0 1 1-1h8a1 1 0 0 1 1 1v10a1 1 0 0 1-1 1zm-9 2a1 1 0 0 1-1-1V5h-.89a.1.1 0 0 0-.11.09v9.82a.1.1 0 0 0 .11.09h7.78a.1.1 0 0 0 .11-.09V14zm2-2a1 1 0 0 1-1-1V3h-.89a.1.1 0 0 0-.11.09v9.82a.1.1 0 0 0 .11.09h7.78a.1.1 0 0 0 .11-.09V12z';
```
tVeratti: What do you think about using ternary statement here instead?

```
const path = filled
    ? 'M12 12v.91A1.1...'
    : 'M12 12v.91A1.1...';
```

troyblank: @tVeratti I updated all icons to use this with a new commit to this PR; makes a lot of sense to me. Mind approving if all looks well?

PR #784, opened 2018-10-11
TFW-2099 - added small screen tab nav styles

troyblank:
This adds collapsible navigation on smaller screens.

I did think of using `aria-expanded` but thought differently as the following are true 1. the buttons are always showing to a screen reader as their height is the only thing animated and 2. expanded will be false on larger screens unless I use width dom measuring to determine what media query the current design in which seems icky.

I did however make sure the buttons were never really hidden for a screen reader; it's only a visual benefit at this point to tuck away the nav and expose more main content.


components/headerPane/tabNavigation.js
```diff
@@ -1,29 +1,56 @@
...
+            >
+              <button
+                className={'tab-navigation__toggle-button'}
+                onClick={this.onToggleChange}
+              >
+                <div className={'tab-navigation__caret'}><IconCaret direction={'down'} /></div>
+              </button>
+              <label className={'tab-navigation__select-label'}>Select</label>
+              {menu.map(nav => (
+                <TabButton
```
tVeratti:
```
<TabButton
  {...nav}
  activeMenuId={activeMenuId}
  //etc..
/>
```

Spreading the nav item gets all of its goodies with less work.

components/headerPane/tabButton.beta.scss
```diff
@@ -1,22 +1,32 @@
...
-    height: $space-xxl;
+    height: $button-height;
     margin: 0 $space-sm;
+    transition: height $animation-time-crisp $animation-curve-crisp;
+    text-align: left;
 
     @include breakpoint($size-small) {
         display: block;
+        width: 100%;
+        height: 0;
```
tVeratti: Back to your comment about a11y. If this is visually hidden, is it still focusable through the keyboard? If yes, does the user have any indication of what they've got focused?

tVeratti: And I mean a11y for sighted keyboard users, to be clear.

troyblank: Right now no, we don't have any visual indication of focus. This is something I did not see in the design and need to ask Casey about.

tVeratti: Yeah I didn't think we had made any progress on that yet - just want to also point out cases like this where even if we did have a focus border color (or something) for the button, this wouldn't give any indication.

troyblank: Agreed, I left a note on the design for Casey to consider and I am going to consider this a thing down the road sorta thing. But great thoughts and I totally agree we should not stop thinking about this.

components/headerPane/tabNavigation.js
troyblank: I switched this to using spread. Not sure why but I always forget I can do that with props. Thanks for keeping me cool :)

troyblank:
@jasonhale  Yea the demo can be see in the new base layout page as well + in the header nav component section of the main components page.

PR #785, opened 2018-10-12
change open sans to var font inter

troyblank:


assets/sass/partials/_typography.scss
```diff
@@ -1,13 +1,27 @@
-@import url('//fonts.googleapis.com/css?family=Dosis|Open+Sans:400,700');
+// sass-lint:disable-block no-url-protocols
```
troyblank: S3 does not like the relative URL, so has to be `https`

assets/sass/base/silents/_typography.scss
```diff
@@ -26,6 +26,12 @@
     line-height: 1.6;
 }
 
+%italic {
```
tVeratti: I'm not overly concerned but it's worth a quick question - does this make the most sense as something we extend, instead of a mixin? Might be more of a general question to how we currently use these silents for extends.
https://csswizardry.com/2014/11/when-to-use-extend-when-to-use-a-mixin/

troyblank: @tVeratti I have always seen using mixins as really bad for performance as it duplicates a lot of redundant code and makes the file size bigger... but yea I agree with that article as well... what to do what to do.

tVeratti: Might be something where BIG chunks of styles are extended, but 3 lines or less are mixin? Idk :D Tough situation.

jasonhale: AH.  Nice.  I haven't come across this question before.  The article does make a good point - and the resultant code would lead me to prefer _@mixin_ in this situation.

Though, to the point you've both made... such a small amount of code.  Not going to hurt anything by this small bit.  If it were anything larger, maybe.

My vote is mixin.  But really not gonna lose any sleep if it doesn't change.  :)

troyblank: OK @jasonhale with the tie breaker  @tVeratti  moved to using a mixin, long live the mixin :)

PR #786, opened 2018-10-15
TFW-2044 - Base nav buttons for channel, support and user

jasonhale:
Modified NavItem to vary between links and buttons, added an active state to handle icon changes.  Set up base items on bottom of navigation.  No mobile navigation began here.


components/navPane/navItem.js
```diff
@@ -1,9 +1,21 @@
...
+export default class NavItem extends PureComponent {
+    static defaultProps = {
+        route: '',
+        type: 'button',
+        visiblyActive: false
+    };
+
+    state = { isActive: false };
+
+    onFocus = () => this.setState({ isActive: true });
```
troyblank: why not handle this with css `:active` ? is there a reason why there is an advantage using js?

components/navPane/navItem.js
```diff
@@ -15,22 +27,42 @@ export default function NavItem({ route = '', icon, label }) {
...
+              <button className={`navitem__button ${activeButtonClass}`} onFocus={this.onFocus} onBlur={this.onBlur} type={'button'}>
+                <span className={'navitem__content'}>
+                  { icon &&
+                    <span className={'navitem__icon'}>{ React.cloneElement(icon, { filled: isActive, ...icon.props }) }</span>
+                  }
+                  { label &&
+                    <span className={'navitem__label'}>{ label }</span>
+                  }
+                </span>
+              </button>
```
troyblank: There are two pretty big pieces of jsx, would it make more sense to split them into two components local to the navPane directory?

components/navPane/navItem.js
jasonhale: We can't (currently) change the icons filled/unfilled state via CSS.  Otherwise, yeah, `:active` would be the path of choice.

components/navPane/navItem.js
jasonhale: I didn't feel like they should necessarily be 2 separate things.  Mainly because they have the same structure and styles aside from some minor differences.  Though, the content could be centralized to a function... like  a `getContent()` that returns the `<span className={'navitem__content'}>...</span>`.

I can start there, see how everyone feels.  If strongly opposed, i can break this out into 2 components.

components/navPane/navItem.js
troyblank: ah this makes sense I didn't think about that. Cool I am down with this then.

components/navPane/navItem.js
troyblank: Yea doesn't have to be 2 components per say my main reaction was it just seemed like a giant ternary. so maybe just something to make it a little more readable like a `getContent`  and broken apart. Thanks!

jasonhale: Updated.  Centralized the repeated bit.  Now the ternary is strictly choosing between the `button` and the `NavLink`.

PR #788, opened 2018-10-16
TFW-2042 - added sub level navigation to header pan tab nav

troyblank:
With this I DID NOT mess with anything inside the `<SelectDropDown />` Because I didn't want to change a bunch of stuff all at once and Jason was doing some work on it.

I made a story to do this work next > https://topbox.atlassian.net/browse/TFW-2131

Also note mobile styles are a separate story > https://topbox.atlassian.net/browse/TFW-2050 (this may take duplicated sub menu, but I tried to plan a spot for that as much as possible)

screen shot 2018-10-16 at 4 35 14 pm


components/headerPane/tabButton.beta.scss
```diff
@@ -56,3 +65,20 @@
     top: 3px;
     margin-right: $space-xs;
 }
+
+.tab-button__caret {
+    display: inline-block;
+    position: relative;
+    top: 3px;
```
tVeratti: what is the significance of 3px? Just visual perfection? If the font gets zoomed, will 3px still be the right distance?

components/headerPane/tabButton.js
```diff
@@ -1,33 +1,101 @@
...
+        e.preventDefault();
+
+        this.selectDropDown.onKeyDown(e);
+    }
+
+    onSubMenuSelect = (e) => {
+        const { value, onMenuSelect } = this.props;
+
+        this.setState({ showingSubMenu: false });
+        onMenuSelect({ topLevelMenuValue: value, subLevelMenuValue: e.value });
```
tVeratti: I'm glad you passed the `topLevelMenuValue` again for consistency.

components/headerPane/tabButton.beta.scss
```diff
@@ -56,3 +65,20 @@
...
+    margin-left: $space-sm;
+    color: $color-blue;
+}
+
+.tab-button--expanded {
+    font-weight: 700;
```
tVeratti: So the changing of the weight changes the width of the tab which in my opinion is a little sketchy. What do you think? Do you notice it?

troyblank: Yea I def do and I agree with you, but this is what the design does.... Maybe we can refactor with casey. I would hate to make an explicit width.

components/headerPane/tabButton.beta.scss
troyblank: Correct; but I after talking to you I tried rems and they do a little better job keeping this visual perfection. So I updated to use them. Thanks for the nudge.

components/headerPane/tabButton.js
troyblank: It's also needed; when you select something from the sub menu you also have to set the main level as active.

tVeratti: Yeah - I could just see a scenario where the value is reused from the parent (since both are from props and therefore exist up there) but that could get stale. So I think you have the right setup here :)

PR #789, opened 2018-10-17
TFW-2055 - Add Sign out submenu to user option on nav

tVeratti:
The function of actually logging out (on click handler).
TFW-2007 for the followup.


components/navPane/navItem.beta.scss
```diff
@@ -89,6 +89,19 @@ $navitem-size: $space-xxxl; // 56px
...
 }
 
+.navitem__label {
+    flex: 1;
+    text-align: left;
+}
+
+.navitem__dropdown {
+    > div {
+        left: 90%;
```
troyblank: SO I was talking to @jasonhale  about this. I feel like `<SelectDrop />` should allow better control of position. Right now it seems like it's all over the place. Sometimes we are passing in a style variant this is using the wrapper to move it. Let's talk about a possible solution that will work for all cases we need to position this better. Also this does not work on small screens.

screen shot 2018-10-17 at 9 12 40 am

components/navPane/navItem.js
```diff
@@ -25,19 +35,43 @@ export default class NavItem extends PureComponent {
...
+    itemDropDown = () => {
+        const { options } = this.props;
+        const { layoutInformation, isActive } = this.state;
+
+        return (
+          <SelectDropDown
+            options={options}
+            showing={isActive}
+            parentLayoutInformation={layoutInformation}
+          />
```
troyblank: There is no key binding to this to allow use of keyboard navigation. Also I have been using `aria-activedescendant` as well with the `focusChanged` prop.

components/navPane/navItem.beta.scss
jasonhale: @Troy, those menu items (channel, client, support) aren't going to be formatted like that in mobile.  There is a "make it work in mobile" ticket in the backlog - but I don't think it's specific to navigation.  That portion of the navigation will most likely be handled by a wholly separate component.

troyblank: Ah OK, sounds good. I will ignore that then :) ... we still need to figure out positioning though.

components/navPane/navItem.js
```diff
@@ -25,19 +35,43 @@ export default class NavItem extends PureComponent {
...
           <span className={'navitem__icon'}>{ icon }</span>
         }
         { label &&
           <span className={'navitem__label'}>{ label }</span>
         }
+        {
+          options &&
+          <div className={'navitem__dropdown'}>
+            <IconCaret direction={'right'} />
+            {this.itemDropDown()}
```
jasonhale: I'd suggest moving the dropdown to inside the `<button>` (between lines 103/104).
Only because the button is the only version of NavItem that will have a dropdown.

components/navPane/navItem.js
troyblank: Thanks Tatiana, looks like arrows are working now.

components/navPane/navItem.beta.scss
troyblank: I still we need to wait to see what @jasonhale  comes up with positioning before we merge this in. Everything looks great otherwise though, marking as request for change until this can be done.

tVeratti: It might make the most sense to just consider this done and create a 1pt story to make the change when it's ready.

troyblank: That's cool with me so long as we don't forget about, be sure to make the story and I will go ahead and approve.

tVeratti: https://topbox.atlassian.net/browse/TFW-2150

PR #791, opened 2018-10-17
TFW-2136 - added letter spacing to new font styles

troyblank:


assets/sass/base/silents/_typography.scss
```diff
@@ -2,27 +2,31 @@
     margin: 0;
     font-size: 1.8rem;
     font-weight: 400;
+    letter-spacing: -.04rem;
     line-height: 2.4;
```
jasonhale: Can we go ahead and change these line-heights to reflect the design specs?
Here, would be either `line-height: 1.333;` or `line-height: 2.4rem;`

Or would you prefer a separate ticket?

assets/sass/base/silents/_typography.scss
```diff
@@ -2,27 +2,31 @@
...
+    letter-spacing: -.04rem;
     line-height: 1.6;
 }
 
 %subtitle-4 {
     margin: 0;
     font-size: 1.6rem;
     font-weight: 700;
+    letter-spacing: -.04rem;
     line-height: 1.6;
```
jasonhale:
```suggestion
    line-height: 1;
```

assets/sass/base/silents/_typography.scss
troyblank: cool call, updated to use `rem`s.

PR #792, opened 2018-10-18
TFW-2050 - added small screen navigation to header pane

troyblank:
This adds small screen navigation for sub menus on the header pane.

screen shot 2018-10-18 at 3 03 51 pm


components/verticalNavigation/verticalNavigationOption.js
```diff
@@ -0,0 +1,25 @@
+import React, { PureComponent } from 'react';
+import './verticalNavigationOption.beta.scss';
+
+export default class VerticalNavigation extends PureComponent {
```
jasonhale:
```suggestion
export default class VerticalNavigationOption extends PureComponent {
```

troyblank: Thanks! fixed.

PR #793, opened 2018-10-18
TFW-2131 - added link style variation to select drop

troyblank:
To do after this: handle positioning based on @jasonhale 's work. But other than that this gives the drop downs in the header pane the style needed per design.

Story for POS: https://topbox.atlassian.net/browse/TFW-2144


troyblank:
@jasonhale It should, but maybe not perfectly. If we end up using this "skin" for submenus we probably need some small tweaking.

PR #796, opened 2018-10-19
TFW-2144 - updated position of select drop on tab nav to better reflect design

troyblank:


test/components/headerPane/tabButton.spec.js
```diff
@@ -79,6 +79,7 @@ describe('TabButton', () => {
               onItemSelect={instance.onSubMenuSelect}
               onFocusChange={instance.onFocusChange}
               styleVariations={['link']}
+              position={{ top: 32 }}
```
tVeratti: So this is how we'll set it everywhere form now on? 
`position={{ top: 0, left: '100%' }}` and so on?

tVeratti: I missed the PR for this actual change xD

PR #799, opened 2018-10-22
TFW-2026 - Navigation should expand/collapse

jasonhale:


jasonhale:
@troyblank - updates made.

PR #803, opened 2018-10-23
upgraded to webpack 4

troyblank:


views/partials/head.hbs
```diff
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
 <head>
-    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta htpt-equiv="Content-Type" content="text/html; charset=utf-8" />
```
tVeratti: htpt is IN, http is OUT

troyblank: Lies; this never happened, you hacked me.

PR #804, opened 2018-10-23
TFW-2037 - add sign out modal and secondary/tertiary buttons

tVeratti:


assets/sass/base/vars/_colors.scss
```diff
@@ -36,7 +36,7 @@ $color-black-black: #000;
 $color-green: #009666;
 $color-bright-yellow: #ff0;
 $color-yellow: #cfd215;
-$color-red: #ad1919;
+$color-red: #f71e44;
```
jasonhale: I'd suggest moving this up to the top under "Secondary" since it's updated to match the new designs - that way we know it's been updated and categorized per design docs.

tVeratti: What about the blues? They are all defined above, which is what led me to just updating these/ Do you think we should move all 'updated' colors down?

tVeratti:
Updated the colors and clicking outside of the modal for closing.

components/navPane/navPane.beta.scss
```diff
@@ -31,6 +31,10 @@ $nav-width-vert-min: $space-xxxl; // 56px
     border-bottom: solid 1px $color-black-2;
 }
 
+.navpane__modal {
+    text-align: center;
+    button { margin-right: $space-sm; }
+}
```
troyblank: Shouldn't we just add this to the style of the modal instead of building a wrapper? I think we can start making all our modals have this style without issue right?

tVeratti: You don't think this would have an unplanned impact on all existing modals bvefore pattern 2.0?

troyblank: Not really; `text-align` is no biggie, and the `button margin` I assume is for the first child.... which should be more elegant anyway. It should be that last child has no margin and anything more than `1` which would not be last should have margin, or something like that. This css add `margin` when not needed. I totally can see this being done all in the current modal with very little impact. Centering text is the only huge change for current modals and I would not sweat that.

tVeratti: Alright I'll update it globally

PR #808, opened 2018-10-24
TFW-2165 - force page breaks between timeline and summary/other

tVeratti:


components/interactionEditor/timeline/timeline.scss
```diff
@@ -12,6 +12,7 @@
     position: relative;
     padding-right: 2.5rem;
     z-index: 2;
+    page-break-after: always;
```
jasonhale: This looks like it should be up in the `@media print{ ...`  like the other implementations.  Or is it intended to be for both screen and print?

tVeratti: I was actually testing it a bit with emulation, no emulation, PDF, and physical print. It didn't matter whether this was in a print block or not. I could put it in one just for clarity, but technically it didn't seem necessary.

jasonhale: Yup, just checked.  it's removed from the normal page flow anyway, wouldn't affect it on screen.

PR #810, opened 2018-10-26
TPB-1129 - changed paginator to report 0 based numbers

troyblank:


components/paginator/paginator.js
```diff
@@ -21,13 +21,14 @@ export default class Paginator extends Component {
...
-        this.reportChange(requestedPage);
+        // '- 1' is to make external communication 0 based.
+        this.reportChange((requestedPage - 1));
     };
 
     reportChange = debounce((page) => {
         const { onPageChange } = this.props;
         if (onPageChange) onPageChange(page);
-    }, 300, { leading: true });
+    }, 300, { trailing: true });
```
tVeratti: I think trailing is the default right? So we might be able to just omit this whole option.

PR #811, opened 2018-10-30
TFW-2150 - Update SelectDropDown position in nav

jasonhale:
Setting SelectDropDown position in the bottom menu items of new nav.
screen shot 2018-10-30 at 12 44 12 pm
screen shot 2018-10-30 at 12 44 22 pm


jasonhale:
@troyblank - yes, no small screen involved here.  I think TFW-2035 should cover that.

PR #814, opened 2018-10-31
 TFW-2019 - Add Heading option to SelectDropDown

jasonhale:
Added an option to pass a header into SelectDropDown and implemented it in the new nav under the user subnav.


components/navPane/navPane.js
```diff
@@ -30,6 +30,14 @@ export class NavPaneComponent extends PureComponent {
     closeModal = () => this.setState({ showSignOut: false });
     signOut = () => (window.location = '/logout');
 
+    // TODO: Add User info once user model is complete.
```
troyblank: This is the components repo, isn't this going to be mock data anyway?

components/navPane/navPane.js
```diff
@@ -30,6 +30,14 @@ export class NavPaneComponent extends PureComponent {
...
+    userHeader = () => (
+      <div>
+        <strong>LeBron James</strong><br />
+        Nike
```
troyblank: Any chance we can use a client that is not really our client? I am probably paranoid but it's good to avoid any possible conflict.

jasonhale:
>1. should the cursor be a pointer when you hover over this header?
Good catch.  No. Fixing.

>2. shouldn't be passing in mock data for what's showing like we do with our other demos?
Working on that now and will have that in my next PR.

PR #821, opened 2018-11-06
TFW-2218 - added active display on sub menu header tab nav

troyblank:
This adds back active display on header nav sub menu's which I broke when I made things a tad simpler for react router. Also a few minor style tweaks added to make things look more like they were designed.

screen shot 2018-11-06 at 10 24 17 am


package.json
```diff
@@ -74,6 +74,7 @@
         "sass-loader": "^7.1.0",
         "style-loader": "0.13.1",
         "webpack": "^4.20.2",
+        "webpack-livereload-plugin": "^2.1.1",
```
jasonhale: 👍

jasonhale:
and one of us can add the updates to the base layout demo at some point.  ;)

PR #822, opened 2018-11-07
TFW-2035 - Small screen user navigation

jasonhale:
Adds mobile navigation, specifically the user navigation items (channel, user).

This affected VerticalNavigation by adding options much like SelectDropDown.


components/baseLayout/baseLayout.beta.scss
```diff
@@ -4,6 +4,7 @@
     display: grid;
     width: 100vw;
     height: 100vh;
+    background: none $color-white;
```
troyblank: Why `none` here? Shouldn't `background: $color-white;` work?

components/navPane/navItem.beta.scss
```diff
@@ -90,6 +90,35 @@
     padding: $space-xs;
 }
 
+.navitem .vertical-navigation {
+    display: none;
+    position: fixed;
+    top: initial;
+    bottom: 0;
+    max-height: 100%;
```
troyblank: Nice, this is covers well adding scroll bar when needed.

components/baseLayout/baseLayout.beta.scss
jasonhale: :facepalm: Yeah.  It's superfluous.  Removing.

components/icons/clear.js
```diff
@@ -0,0 +1,20 @@
...
+        height={24}
+        viewBox={'0 0 24 24'}
+        xmlns={'http://www.w3.org/2000/svg'}
+        fill={'currentColor'}
+      >
+        <title>clear</title>
+        <path d={path} />
+      </svg>
+    );
+}
```
troyblank: @jasonhale this icon already exists and it's called `close`. Mind if we use that one? I don't mind updating the old icon to be more like our newer ones but just seems a little duplicate like to me right now.

components/navPane/navItem.js
```diff
@@ -120,11 +176,11 @@ export class NavItemBase extends PureComponent {
     }
 
     render() {
-        const { path, to, icon, label, type, visiblyActive, options, disabled } = this.props;
+        const { path, to, icon, label, type, visiblyActive, options, disabled = false } = this.props;
         const { isActive, value, focus } = this.state;
 
         const hasSubMenu = options && options.length;
-        const activeButtonClass = (visiblyActive || isActive) ? 'navitem__button--active' : '';
+        // const activeButtonClass = (visiblyActive || isActive) ? 'navitem__button--active' : '';
```
troyblank: Is this need anymore?

components/icons/clear.js
jasonhale: Yup. I can add the new one to the old Close icon.

components/navPane/navPane.beta.scss
```diff
@@ -87,27 +91,78 @@ $nav-width-vert-min: $space-xxxl; // 56px
...
+
 @include breakpoint($base-layout-breakpoint) {
     .navpane {
-        display: flex;
+        display: grid;
         justify-content: space-between;
         width: 100%;
         height: auto;
         transition: unset;
+        grid-template-areas: 'smlogo smmain smusertoggle' 'usermenu usermenu usermenu';
```
troyblank: is this a new language? :P .... should we use hyphen or camel case so this is readable?

jasonhale:
@troyblank - RE: mobile subnav heights and hiding main nav.

The subnavs have a minimum height of the main navigation, and each subsequent subnav has a min-height of it's parent.  This way nothing behind ever shows, and navs stay the same size or grow taller the deeper nested they go - never smaller unless closing.

components/navPane/navPane.beta.scss
jasonhale: Going camelCase so they're not confused with selectors.

troyblank:
> @troyblank - RE: mobile subnav heights and hiding main nav.
> 
> The subnavs have a minimum height of the main navigation, and each subsequent subnav has a min-height of it's parent. This way nothing behind ever shows, and navs stay the same size or grow taller the deeper nested they go - never smaller unless closing.

@jasonhale  So two things from this. 1. this differs from the way the top nav works, for consistency wouldn't we want these two animations mirroring? 2. This might mean a large white space if the main nav grows, might that be a bit odd?

troyblank:
Looking at the white space it's not too horrible.... but I am still a little bothered it does not match the other navigation.

screen shot 2018-11-07 at 5 41 48 pm

jasonhale:
I understand the concern and agree on some points.  I do think that in the case of the subnavs in the footer, they should slide to match the animation at top.  Not totally sold on the main nav sliding over because it's a little different scenario than the VerticalNavigation at top.  But, not totally opposed either. ;)

As far as subnavs, just went through a bunch of scenarios with Casey.  At the bottom of the page, it makes sense for submenus to have a min-height of their parent so that the header - a focal point - doesn't move around the page.  At top, the submenu header doesn't move since content expands downward and its current interactions make perfect sense there.  The footer, though, is different since it's tied to the bottom of the window and the header moves instead of the content (better to keep it from moving if at all possible).  The white space isn't totally disagreeable and we don't have other user-centric menu items planned anytime soon that would cause this to get too big.

Casey is even rethinking some of the small screen navigation because of it.

So...
- agreed on slide transition for submenus.
- would like to keep the min-height implemented to keep visual focus from jumping.

Next - we opposed to making a new ticket to address these changes?  That way we keep moving.

troyblank:
@jasonhale  Yea I agree with that 100%. A new ticket would be fine but we need to make sure we have to work set aside so we do it. Overall this menu is rockin' though I think our small screen menus are looking puuurdy sexy.

jasonhale:
Writing that ticket up now.

PR #824, opened 2018-11-09
TFW-2255 - updated styles of notification bar so session timeout could use

troyblank:
This is 75% refactoring to BEM, 5% style modification and 10% making a demo to get confidence we can use the current notification bar as a session timeout.

Right now only password is about to expire is using this component but it can easily be re-used for the session expire. This may need some polish after Case sees it but for now it has to be there so I am trucking on with the functional side of it.


components/notificationBar/notificationBar.beta.scss
```diff
@@ -17,63 +17,70 @@
         display: none;
     }
 
-    svg {
-        position: relative;
-        top: 3px;
-        margin-right: 1.5rem;
+    // for dynamic child content
+    a {
+        color: $color-white;
```
jasonhale: This could be `inherit` so it's always the same color as the parent - but I don't see that it would ever change, either.

PR #828, opened 2018-11-13
TFW-2277 - NavPane pass selected client

jasonhale:
Need notification when a client is selected.  Needed for TFW-2001 in RCI.


test/components/navPane/navPane.spec.js
```diff
@@ -301,4 +303,16 @@ describe('NavPane', () => {
 
         assert.equal(clientOption0.length, 0);
     });
+
+    it('should notify parent if a client is selected with onClientSelect', () => {
```
tVeratti: Might be a good idea to add a test for selecting a value that doesn't fit into the supplied list. Doesn't add coverage % but it does catch a case that could break someday with code changes.

jasonhale: Would the issue you're thinking of be covered under the test [above](https://github.com/topbox-rci/rci-components/pull/828/files#diff-1c4d0d29191acd512aec0539466ddd6bR108)?

tVeratti: Oh actually yes, that works! It was collapsed so I completely missed that. Thanks!

PR #831, opened 2018-11-14
TFW-2283 - Show client logo in subnav

jasonhale:
Shows the logo for the current client in the nav pane.  If no logo url is passed, uses the new IconUser instead.


assets/js/demo/components/baseLayoutDemo.js
```diff
@@ -100,7 +100,7 @@ export const channelList = [
 ];
 
 export const clientList = [
-    { id: '1234567890', name: 'Wally World' },
+    { id: '1234567890', name: 'Wally World', logo: 'https://signin-testing.topbox.io/logos/topbox.svg' },
```
troyblank: Can we use either a local file here or at least one in s3 for the components project itself? I am not sure we should rely on other projects to keep assets, since this is basically the base for RCI. I personally think it's best in `/assets/images/` which we are currently using for bitmaps in the project.

If you need me to set up an s3 bucket let me know.

troyblank:
@jasonhale  talked to casey and her added some sample client logos we can use > https://projects.invisionapp.com/d/#/projects/prototypes/15214437/assets

my vote would be the demo one so it's more generic.

assets/js/demo/components/baseLayoutDemo.js
jasonhale: If it's just demo purposes, we can just stash one in assets.
I used that one above since that is what is coming across in the clients object currently.  Which actually leads to other questions.

troyblank: If it helps the plan talked about was to change the logo over and soon as we move someone over to the new experience.

PR #840, opened 2018-11-28
TFW-2266 - force subMenuWrapper to position by bottom when at bottom of window

tVeratti:
Tested against the existing submenus on tables (Workflow cells, for example) and they looked good. If you know of other places where subMenus currently exist that might have weird edge cases let me know!


components/selectDropDown/subMenuWrapper.js
```diff
@@ -5,7 +5,9 @@ import styles from './subMenuWrapper.scss';
 
 export default class SubMenuWrapper extends Component {
     static defaultProps = {
-        subMenuPlacement: 'right'
+        subMenuPlacement: 'right',
+        subMenuDirection: 'top',
+        maxOptionPaneHeight: 500
```
jasonhale: Casey requested 496. :)

tVeratti: Oh snap yes you're right! Updating :)

troyblank:
I am not going to knock us for this now since we are moving fast, but i noticed it's really easy to mouse out of the list. Ideally we would have some invisible buffer on at least the bottom and top so that it's not super easy to mouse out of.

PR #841, opened 2018-11-30
TFW-2309 - update new datepicker in create form

tVeratti:


components/createForm/createForm.js
```diff
@@ -239,19 +243,21 @@ export default class CreateForm extends PureComponent {
...
                     label={'Agent'}
                     options={agents}
+                    styleVariations={['boxed']}
                     onUpdate={this.onFormChange}
                     error={this.state.errors.agent}
                     errorMessage={this.state.errors.agentName}
                   />
                 </li>
                 <li>
-                  <DatePicker
```
troyblank: We need to remove all the old 3rd party lib stuff in this component as well as the component itself.

test/components/createForm/createForm.spec.js
```diff
@@ -109,19 +110,21 @@ describe('Create Form', () => {
...
                     id={'interactionDate'}
                     name={'interactionDate'}
-                    label={'Interaction date'}
-                    error={instance.state.errors.interactionDate}
-                    errorMessage={instance.state.errors.interactionDate}
-                    onUpdate={instance.interactionStartDateChange}
+                    label={'Interaction Date'}
+                    styleVariations={['boxed']}
+                    onChange={instance.interactionStartDateChange}
+                    multi={false}
```
troyblank: this is still showing up multi for me

screen shot 2018-11-30 at 4 41 06 pm

tVeratti: Multi is about selecting a range vs. a single date, not how many calendars show :)

troyblank: OH, still seems odd there is two here right? specially when you can change one independent of another when selecting one date.

tVeratti: Maybe, but that's a subjective design discussion we can have under less crunch :D

troyblank: The old one used one though, this does change behavior. I am fine with leaving this in considering it's down to the wire but we really need to make sure this is ok with design.

tVeratti: We'll make sure to address this next sprint as we clean up.

PR #846, opened 2018-12-04
TFW-2340 - add pendo-target classes to user nav

jasonhale:


jasonhale:
Updated to use data attributes instead to bypass any style conflicts.

components/navPane/navItem.js
```diff
@@ -175,8 +175,18 @@ export class NavItemBase extends PureComponent {
...
+        keys.forEach(key => (params[`data-${key}`] = attribs[key]));
+
+        return params;
+    }
+
     render() {
-        const { path, to, icon, label, type, visiblyActive, options, disabled = false } = this.props;
+        // addition of 'data-' attributes added for use by Pendo analytics tagging.
+        // if data-attributes are no longer used at any point, consider removing this feature.
+        const { path, to, icon, label, type, visiblyActive, options, disabled = false, dataAttributes = {} } = this.props;
```
troyblank: Whoa, this is a lot of things. are these all needed by pendo? If so can we just add one big `data-pendo` target.... I would assume these could not be transitioned to something another analytics software..

Correct me if I am wrong but it seems like we went from one class pendo needed to 11 data attributes? What changed?

components/navPane/navPane.js
```diff
@@ -199,9 +199,18 @@ export class NavPaneComponent extends PureComponent {
...
+                        target: 'channelselect' // only used for Pendo tagging
+                    }}
                   />
                   <li className={'navpane__item-separator'} />
-                  <NavItem label={'Support Center'} icon={<IconSupport />} />
+                  <NavItem
+                    label={'Support Center'}
+                    icon={<IconSupport />}
+                    dataAttributes={{
+                        target: 'supportcenter' // only used for Pendo tagging
```
troyblank: `target` does not let me believe it's used for pendo tagging or even tracking. Instead of commenting every-time we use the `target` attribute, can't we just make it `pendo` so it's clear what it is used for. Something more descriptive I feel is in order like `pendo` OR `tracking` and it won't have to be spelled out.

jasonhale: Yeah.  I used "target" since that was the purpose of the attribute, to be used as a target.  Using "pendo" seemed more like using it as data.  But this suggestion is still good and makes sense.
PR udpated.

PR #847, opened 2018-12-05
TFW-2382 - Update tab navigation styles

tVeratti:
Though we won't have icons for qlik now, eventually we'll need to make sure they take a color prop for their fill/stroke.


components/icons/target.js
```diff
@@ -1,17 +1,19 @@
 import React from 'react';
+import { getColor } from '../util/colors';
 
-export default function IconTarget({ filled }) {
+export default function IconTarget({ filled, color = 'currentColor' }) {
```
troyblank: why don't we just set the fill color with css?

tVeratti: I guess we could, it would be easier. I was trying to be consistent but now that you mention it, maybe it's not worth it - or also no longer the ideal way going forward. I'll change to css.

tVeratti: @troyblank Updated to css.

troyblank: Yea unfortunately most of our icons use the color prop to make this confusing but i think we decided a bit ago to go with just currentColor css from here on out.

PR #849, opened 2018-12-05
TFW-2384 - update Side nav primary/secondary text styles

tVeratti:
Shadows approved by Casey, we talked back and forth a bit for the final updates.


components/navPane/_navPane.common.scss
```diff
@@ -5,6 +5,9 @@ $navitem-size-small: $space-xxl; // 48px
 $navitem-color: $color-white;
 $navitem-font-size: 1.4rem;
 
+$navitem-font-shadow: 0 1px 2px rgba($color-black-black, .24);
```
troyblank: I assume we only use these in the design on the nav pane?

PR #850, opened 2018-12-05
TFW-2203 - added fill attribute to icons in buttons when hovering occurs

troyblank:


components/button/button.js
```diff
@@ -35,11 +49,13 @@ export default class Button extends PureComponent {
                   { [styles['button--with-icon']]: icon }
                 )}
             onClick={this.onClick}
+            onMouseOver={this.onMouseOver}
```
tVeratti: I would change this to `onMouseEnter`/`onMouseLeave`, otherwise this will fire tons of times while they move the mouse over it.

PR #865, opened 2018-12-11
fixed build if no public dir exists

troyblank:
realized this morning if the `public` dir does not exist the build fails due to an error thrown with `rm -r`. I made this so that no error is thrown if the dir does not exist using `npm -f`.


troyblank:
@jasonhale never ran across `rm -rf` and didn't see much of when searching, gong to merge this as it works but feel free to improve.

PR #867, opened 2018-12-11
fixed error for filled attribute being boolean

troyblank:
Turns out the filled prop is giving errors for non icon props (such as using a span in the user nav item). It also turns out that this prop is not even needed and filled is set properly without it.


troyblank:
This is the error this gives before this PR

Warning: Received `false` for a non-boolean attribute `filled`.

components/navPane/navItem.js
```diff
@@ -188,14 +188,13 @@ export class NavItemBase extends PureComponent {
...
         const { isActive, value, focus } = this.state;
-
         const hasSubMenu = Boolean(options && options.length && !disabled);
         let iconLookup;
 
         if (icon) {
             iconLookup = ('link' === type) ?
                 this.getRoutedIcon(icon, path)
-                : React.cloneElement(icon, { filled: isActive, ...icon.props });
+                : React.cloneElement(icon, { ...icon.props });
```
jasonhale: That 'filled' prop was necessary for instances where there was no user logo and we had to use the IconUser instead.  In doing so, would have fill that icon based on if the navitem was active or not.

This approach removes filling that IconUser when the navitem is active.

My suggestion would be, in NavPane, where the user icon is assigned (navPane.js, 119) - return a component with the client logo that will ignore that `filled` prop.  Unless filling that icon when active is no longer a design requirement.

troyblank: Ah, that's what that prop was used for, looking into a way to solve this without applying the prop to a bunch of stuff that does not need it. thanks!

assets/js/demo/components/baseLayoutDemo.js
```diff
@@ -100,7 +100,7 @@ export const channelList = [
 ];
 
 export const clientList = [
-    { id: '1234567890', name: 'Wally World', logo: '/static/images/demo-client-logos/No_Client.png' },
```
troyblank: I removed this because it has more complex logic without that is better tested, we can always make an new demo page if needed to show logos.

components/navPane/navItem.js
troyblank: ok @jasonhale what I did here was check to see if the prop even has `filled` before setting it. That way only an icon that has filled pre-set to have it toggle based on if the icon item is being used. Let me know what you think.

assets/js/demo/components/baseLayoutDemo.js
jasonhale: There are tests to cover passing images to this in `navPane.spec.js`.
What logic is missing?

components/navPane/navItem.js
jasonhale: Looks good.  The icon for the support nav item will have to have it's `filled` prop set also.  (`<IconSupport />`)

assets/js/demo/components/baseLayoutDemo.js
troyblank: If there is tests to cover this why are they not failing? Seems odd there is test to test if the logo is passed and they are still reporting good.

I am talking about visually testing the user icon that never shows up without this data altered this way. Which is the reason I missed the point of the filled being passed in.

troyblank: Looking deeper the tests do not use this mock data. I would prefer if we have one demo we leave it with something that has a toggle rather than the static image.

components/navPane/navItem.js
troyblank: @jasonhale  cool I made sure that had the `filled` prop and pushed up new updates.

PR #871, opened 2018-12-12
TFW-2428 - adjusted universal search query node initial id to ensure no duplicates

troyblank:
The queryIdCount needs to be + 1 of whatever i there now otherwise it makes a duplicate.


components/universalSearch/searchBar.js
```diff
@@ -20,7 +20,7 @@ class SearchBar extends PureComponent {
             value: null,
             data: {},
             dimensionInputFocused: false,
-            queryIdCount: queries.length ? queries[queries.length - 1].id : 0
+            queryIdCount: queries.length ? (queries[queries.length - 1].id + 1) : 0
```
tVeratti: Should the `: 0` be updated to `: 1` then?

troyblank: @tVeratti  I guess I was wrong it's not really 1 based, it is 0 based, it's just that the queryIdCount needs to be `+ 1` of whatever i there now otherwise it makes a duplicate. If we would like it to be based starting with 1 I can update this if desired but I don't think it's needed.

troyblank: I am going to amend the commit message to be more clear.

troyblank: ok all should be updated now, holler if this is still not clear.

tVeratti: Cool, that does clear it up! Thanks!

PR #875, opened 2018-12-12
TFW-2446 - sort on original priority as num (not string label)

tVeratti:


tVeratti:
Technically there is a test for the sorting, but I can expand it to double digits campaigns :)

tVeratti:
Actually my bad, there's a test for grouping but not sorting. Adding!

PR #877, opened 2018-12-14
TFW-2452 - show message when no campaigns are available

tVeratti:


components/messages/messages.en-us.js
```diff
@@ -14,6 +14,7 @@ export default {
     MINIMUM: 'Minimum',
     MAXIMUM: 'Maximum',
     NO_LIMIT: 'No limit',
+    NO_ACTIVE_CAMPAIGNS: 'You have no active campaigns',
```
troyblank: The design says "Your selected channel has no content for this tab.", why is this different?

tVeratti: That's the message for the qlik tab. Check it out here: https://topbox.atlassian.net/secure/RapidBoard.jspa?rapidView=49&projectKey=TRW&view=planning&selectedIssue=TFW-2452

components/infoPanel/infoPanel.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { IconInfo } from '../icons';
+import './infoPanel.beta.scss';
+
+export default function infoPanel({ iconType, children }) {
+    let icon;
+
+    switch (iconType) {
```
troyblank: Why not just pass the icon in?

components/messages/messages.en-us.js
troyblank: coolness!

components/infoPanel/infoPanel.js
jasonhale: This part was me (was trying to help out).  And probably over-architected.  Since this would be used in RCI and rci-components, thought it would be beneficial to just declare the _type_ of panel (informative, warning, etc) rather than passing the icon in (requires other imports/setup to use).

Was a real quick, rough start.  Open to suggestions.

troyblank: If we only have a few types this makes sense to me. I just was thinking we didn't want to make a case statement for tons of icons. So here is my issue with this then, why not name the prop `type` so it's "theming" the component instead of declaring what icon you want.  This would make the switch statement make more sense to me. Giving the consumer a theme instead of trying to "pick" and icon makes it seem more finite to me.

PR #879, opened 2018-12-18
 TFW-2406 - fixed render lags with inputs by removing state key

troyblank:
I determined date picker lag is due to duplicate key props on the helper switch of an input. I also fixed a small padding issue that was driving me bonkers.


tVeratti:
Here's a pen I've started to try and reproduce the issue: https://codepen.io/anon/pen/LMRKPx?editors=0010

It's weird to me that keys at different scopes could have conflicts. Back in React 0.13.3 I remember specifically studying how keys were generated with the parent prefix - but they are hidden now so I can't verify anymore.

Clearly I can't argue with the results of what Troy found/changed. But I don't fully understand it, either. It worries be a bit and am hoping we can figure out exactly what's going on here.

troyblank:
I should state after what @tVeratti  said, I am not 100% sure why this works as I am not super intimate with the internal workings of react and how keys cause renders and what not. But this was night and day an improvement for me, it appeared to have fixed the issue and had no issues in itself.

troyblank:
This apparently is not the solution.

PR #887, opened 2018-12-20
TFW-2483 - Workflow, load highest priority campaign

jasonhale:
Now, when Workflow initially loads or if campaigns change, it's gets the interactions for the highest priority campaign.


components/workFlow/filterBar.js
```diff
@@ -8,7 +8,32 @@ import { NewSelect, Button, Statistic } from '../';
...
+
+        if (!campaign && initialCampaign) {
+            this.onCampaignChange(initialCampaign);
+        }
+    }
+
+    componentDidUpdate(prevProps) {
+        const { campaigns } = this.props;
+        const { campaign } = this.state;
+        const initialCampaign = this.getInitialCampaign();
```
tVeratti: Calling this every updated which builds the campaign options (which is like 6 loops over the campaign data) might be something we could optimize to only run if the given list of campaigns props has changed. You already have the check for it, but perhaps only building when that case is true would save a ton of rebuilds.

jasonhale: Truth.  Will change.

jasonhale: @tVeratti - updated

PR #896, opened 2019-01-09
added rough vertical bar chart

troyblank:
This might become a base to work off of so wanted a review. Few things here.

1. used d3 (duh).
2. used faux dom for better react to d3 management, will need to use derive state from props as charts get dynamic instead of component will mount lifecycle. See [D3 within React the right way](https://oli.me.uk/2015/09/09/d3-within-react-the-right-way/).
3. imported data from scotts example data, right now using a temporary branch but will hook to master later.

screen shot 2019-01-09 at 11 54 43 am


assets/js/demo/components/verticalBarChartsDemo.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { clone } from 'lodash';
+import { VerticalBarChart } from '../../../../components';
+import payload from '../../../../node_modules/@topbox-rci/visualizations-test-data/phoneTopicAggregations/payload.json';
+
+export function addPercentsToData(values) {
```
troyblank: I probably could have used d3 scale here, man I am rusty. Will probably improve to this later on.

PR #901, opened 2019-01-15
allowed line charts to be responsive

troyblank:
Quick update to line charts to allow them to take up 100% of their container and keep their aspect ratio. Will not work in IE 11, but if we need to there is some hacks we can go down to get it to work there.

Also added some notes to make the d3 chaining a bit more clear per @tVeratti 's always awesome suggestions.


troyblank:
@tVeratti yea I agree about the labels they will most likely be bumped once we tweak. Casey and I need to sit down and play with them at some point. Right now though you can at least see a lot of labels in one go :D

PR #902, opened 2019-01-15
TFW-2567 - refactor Question component for fewer re-renders / race conditions

tVeratti:
I have been unable to reproduce the documented bug locally, but I can reproduce it within RCI. After watching the logs and seeing these re-rendering dozens of times on initial load, I figured it's likely with some of the data fetching coming in late that certain updates could be overriding the open state of the component.

My solution is to break off the state being based off of props and read straight off of what is available where possible. When a change directly affects the state, it updates it then and there, instead of listening for the props to come back through and triggering an additional followup render.


components/interactionEditor/actionPane/formPane/question.js
```diff
@@ -102,29 +82,31 @@ export default class Question extends PureComponent {
...
+            this.onHideComment();
         }
 
         onQuestionChange(alteredQuestion, index);
     }
 
     onAutoFailChanged = (value) => {
         const { question, index, onQuestionChange } = this.props;
-        const alteredQuestion = Object.assign({}, question);
+        const answers = [...(question.answers || [])];
```
troyblank: Where were you before you worked here and told me to use the spread operator :P

tVeratti: Somewhere else telling other people to use the spread operator! xD

PR #903, opened 2019-01-15
Bar tool tip

troyblank:
Just a little interaction work on the bar charts.

Noticed we were not really loading d3 styles, so I made it global with `.beta.scss`. This is something we might need to fix on RCI as well.


components/visualizations/verticalBar.js
```diff
@@ -16,18 +18,27 @@ class VerticalBar extends PureComponent {
...
         bar.append('div')
             .text(d => d.key)
             .classed('vertical-bar__label', true);
+
+        // Bar
         barWrapper = bar.append('div')
             .classed('vertical-bar__bar-wrapper', true);
         barWrapper.append('div')
             .style('width', d => `${d.percent}%`)
+            .attr('data-tip', d => sanitizer.sanitize(`<label>${d.key}</label><div>${d.doc_count} interactions</div>`))
```
jasonhale: 👍

PR #904, opened 2019-01-15
TFW-2570 - mix onClick/onMouseDown for option events

tVeratti:
This issue only came up for Autocomplete which tipped me off that it was an order of browser event issues, since that one closes with `onBlur`

To get both types working in their own context, I've allowed each to hook into the event that works best with their interactions.

Let me know if you have any questions about the order of events I've mentioned above & in the comment.


troyblank:
@tVeratti  looks like jenkins doesn't like the new test. It works locally but i have restarted it and it fails the same way :(

PR #909, opened 2019-01-17
TFW-2542 - Check Channel Icon for variations of Channel name

jasonhale:
Checks if the channel name _contains_ certain words to assign the correct icon, instead of an exact match.


jasonhale:
@tVeratti - totally legit concern.  And I'll update to address it.  Our business rules say to support "only Chrome", and i feel IE 11 is a vulnerability concern at this point.  But, maybe, until we force users off the app if they're not on a supported browser, I think your suggestion is worth addressing.  Thanks!

jasonhale:
@tVeratti, changed to use `test`.  Better support, and does the same job.  Thanks for bringing up the issue.

troyblank:
Sorry late to the game on this but we only have data in pendo based on new users of our new app. Old app didn't even load in old browsers (to some extent). Our official support before doing new nav which is documented somewhere in confluence is all modern evergreen browsers which does not include IE. I would be fine with not even working about IE at this point considering that on our analytics.
screen shot 2019-01-18 at 1 57 19 pm

PR #910, opened 2019-01-17
added rough donut graph

troyblank:
screen shot 2019-01-17 at 1 06 50 pm


components/visualizations/donut.js
```diff
@@ -0,0 +1,131 @@
...
+
+class Donut extends PureComponent {
+
+    componentDidMount() {
+        const { data, connectFauxDOM, animateFauxDOM } = this.props;
+        const faux = connectFauxDOM('div', 'donut');
+        const segmentSize = 0.575;
+        const segmentGap = 0.01;
+        const layout = {
+            width: 188,
```
tVeratti: Should these be configurable via props? Or is there a reason not to allow it?

components/visualizations/donut.js
```diff
@@ -0,0 +1,131 @@
...
+            .attr('viewBox', `0 0 ${layout.width}, ${layout.height}`)
+            .attr('preserveAspectRatio', 'xMidYMid meet')
+            .classed('donut-chart', true);
+        const donut = svg.append('g')
+            .attr('transform', `translate(${layout.width / 2}, ${layout.height / 2})`);
+        const segments = donut.append('g');
+        const texts = donut.append('g');
+
+        // Drop shadow definition - used on text
+        svg.append('defs')
```
tVeratti: After seeing these a few times, I can't help but wonder if some kind of template system could be used for these svgs?

After a quick browse, it doesn't seem a like something supported out of the box... but since you're in research mode, maybe it's worth just looking at a solution like https://www.johannesbader.ch/2014/03/templating-in-d3js/

Might not be simpler in the end, but I can't help but see all of these js functions that look just like setting up some html.

components/visualizations/donut.js
troyblank: Good question; all these are doing is defining a ratio; with css you can change the main containers dimensions and it will follow suit. Really these could be `width: 1; height: 1`. Going forward I would love to use only css to do the layout of these graphs. The problem here is if we pass these in via props it really does no good as anything BUT a square ratio will screw up the rendering. So really it should be hardcoded but maybe simplified so they are not arbitrary units. I will do some work to make them less specific and add a bit of work to control it via css.

components/visualizations/donut.js
troyblank: 🤔 yea this is very interesting.... what I can't wrap my head around is how that deals with recursion of data. Not sure it would be helpful just to have the the template be the wrapper element and that is all....

components/visualizations/donut.js
tVeratti: Ok that makes a lot of sense! Good plan 👍

troyblank: I played with this some and realized our fonts are not jiving with the ratio scaling we are doing; meaning the font size is very much so effected by the responsiveness of the svg and the initial height and width set on the svg tag itself. This will need to be focused on and a solution needs to be figured out with another PR so that the font size looks the same at all widths of the visualization.

PR #911, opened 2019-01-18
TFW-2621 - Close channel selector after selection

jasonhale:


components/navPane/navItem.js
```diff
@@ -64,7 +64,9 @@ export class NavItemBase extends PureComponent {
         this.setState({ isActive: false });
     }
 
-    onItemSelect = ({ value }) => {
+    onItemSelect = ({ value }, event) => {
```
tVeratti: I would avoid calling it `event` just because I've run into some weirdness on I think FF where the global `event` object gets confused with these being passed around. Just in general it feels safer to go with `e` or `ev` or something similar :)

PR #914, opened 2019-01-21
TFW-2544 - Move ChannelSelect to top of navigation

jasonhale:
This moves the Channel Selector to the top of the nav pane, directly under the TopBox logo.

**Note:** new design has a different mobile nav than currently implemented. As a side effect, added ability to add extra classnames to the NavItem to display in correct position (mobile vs. desktop) until new mobile nav can be built.

Also addressed a test parameter in AutoComplete that would randomly test an incorrect param.


components/navPane/navItem.js
```diff
@@ -89,11 +90,14 @@ export class NavItemBase extends PureComponent {
...
-            const menuPosition = {
-                right: -(width - 8),
-                bottom: 0
-            };
+            const menuPosition = { right: -(width - 8) };
+            if (true === subMenuFlowUp) {
+                menuPosition.bottom = 0;
+            } else {
+                menuPosition.top = 0;
+            }
```
troyblank: I think this code may be a bit of duplication. We have a pretty good detection system of how something should flow with the auto compete via select drop down >

screen shot 2019-01-21 at 1 04 49 pm

Making this a property seems to me like adding too much burden on a developer to consider the layout of the nav which can take it's hints from where the nav is on the page. Also I noticed if the page is smaller than the nav the nav just gets cropped and items are not accessible. I would suggest checking out what select drop does and allowing it to work like it does in autocomplete. I am not sure this prop is even needed if it mimics autocomplete. If there is a reason the default select drop down layout logic can not be applied here let me know maybe you have already been down this route.

jasonhale: I get what you're saying.  Though, I wonder if the prop name `subMenuFlowUp` is too misleading - really, it's more _"pin the dropdown so it covers the button that it's contained in"_ so it adhere's to design specs.  And uses the `position` prop of SelectDropDown to do so (which, currently, overrides SelectDropDown's logic of deciding to open upward from above the parent or downward from below).

If we manually position the SelectDropDown at all, we _should_ take decision of which direction to open away from SelectDropDown - the reason being that we are explicitly deciding where the dropdown is positioned, and if the component decides for us, our manual positioning gets overridden (the direction the dropdown gets opened is controlled via positioning in SelectDropDown).

However, to your point, we could limit the `position` prop of SelectDropDown to only accept horizontal positioning.  Then, we could allow SelectDropDown to decide opening up or down like normal, and add a style variation to allow it to overlay its parent.  My only (minor) issue with that is backtracking on how we envisioned this `position` prop to work.

**Possible solution:** remove this `subMenuFlowUp` prop and utilize `getLayoutInformation()` from SelectDropDown to allow NavItem to programatically decide which direction to open the submenu.  I'll start working that in to check it out.

Let me know if I'm way off base and mis-reading your comment.  ;)

> Also I noticed if the page is smaller than the nav the nav just gets cropped and items are not accessible

Sweet, thanks for catching.  I know the bit causing that. Will fix.  👍

troyblank: I think we are on the same page, your possible solution makes sense and is what I was driving at!

jasonhale: @troyblank - updated.  Give a look-see.

troyblank: This is an improvement, but maybe we were not on the same page sorry. So when i look at the `<SelectDropDown />` component the logic to figure out if the menu goes up or down is already there > https://github.com/topbox-rci/rci-components/blob/master/components/selectDropDown/selectDropDown.js#L418 ... the problem is as soon as you select an x or y position this logic is no longer hit and then we have to duplicate it. To me if you did not provide a specific direction the select drop down should be able to figure out which way to go on it's on right?

Also this is the issue I was seeing where the items are getting cropped that does not happen on other select drop downs (end of the browser window crops the options) >

screen shot 2019-01-21 at 3 50 34 pm

jasonhale: The dropdown getting cut off should already be fixed.  You may have to delete your local branch and re-download (I rebased).
screen shot 2019-01-22 at 9 33 28 am

Going to have to make modifications to SelectDropDown, how it handles applying the values in the `position` prop and adding another style-variation so that it can vertically cover the parent button.

jasonhale: Updated by removing that location logic from NavItem and adding a style-variation to SelectDropDown.

troyblank:
@jasonhale This is looking great but one last thing I am noticing that is really strange; If I scroll in the select drop down I get this error and the screen goes all white >

```selectDropDown.js?b951:341 Uncaught TypeError: Cannot assign to read only property 'top' of object '#<Object>'
    at SelectDropDown._this.assignMenuPosition (VM452598 105:277)
    at SelectDropDown.render (VM452598 105:536)
    at finishClassComponent (VM452507 14:14534)
    at updateClassComponent (VM452507 14:14489)
    at beginWork (VM452507 14:15335)
    at performUnitOfWork (VM452507 14:18150)
    at workLoop (VM452507 14:18190)
    at HTMLUnknownElement.callCallback (VM452507 14:149)
    at Object.invokeGuardedCallbackDev (VM452507 14:199)
    at invokeGuardedCallback (VM452507 14:256)```

On master it is fine, so it seems this PR is causing this.

jasonhale:
@troyblank - PR is updated.  That sillyness shouldn't be an issue anymore.

PR #915, opened 2019-01-22
TFW-2524 - add links to infoPanel

tVeratti:
Design for unavailable pages when Omnichannel is selected is here: https://projects.invisionapp.com/d/#/console/16169512/342385075/preview


components/infoPanel/infoPanel.beta.scss
```diff
@@ -14,3 +19,38 @@
...
+    background: $color-white;
+
+    li {
+        width: 100%;
+
+        &:not(:last-child) { border-right: 1px solid $color-grey-light; }
+    }
+}
+
+.info-panel__link {
```
jasonhale: This is probably nit-picky - but the same element is being targeted in two spots: **.info-panel__links li** and **.info-panel__link**
I'd suggest consolidating all into **.info-panel__link** for ease of maintenance.

assets/js/components.js
```diff
@@ -722,6 +722,19 @@ renderHtml(
     <InfoPanel iconType={'info'}>
           Add pertinent message here;
     </InfoPanel>
+
+    <BrowserRouter>
+      <InfoPanel
+        iconType={'info'}
```
jasonhale: Not your mistake as it was already there - mind changing this so the icon shows in the component showcase (here and the example above)?
the prop is `type`, not `iconType`.  thanks!

PR #917, opened 2019-01-23
TFW-2525 - Separate Omnichannel to top of channel list

jasonhale:
This moves Omnichannel (and Composite since they're used interchangeably) to the top of the channel selector list, and separates them from the other channels by a rule.

https://projects.invisionapp.com/d/#/console/16169512/335269971/preview


components/navPane/navItem.js
```diff
@@ -183,13 +183,26 @@ export class NavItemBase extends PureComponent {
         return params;
     }
 
+    // since both Arrays and Objects are passed as 'options',
+    // check if 'options' exists and has keys/values.
```
troyblank: Have we considered changing the uses of Arrays to Objects to simplify things?

jasonhale: I had that thought at one point.  But, then changing the user settings to use an object instead of an array felt a little hollow - didn't seem to make sense in that context, and only removes the array check in exchange for more configuration (the style options to hide labels).

I'm ambivalent on this one.  Makes sense to me either way - but i've been in here a lot recently.  If it makes more immediate sense to you or @tVeratti, I'd be happy to modify.

PR #928, opened 2019-01-25
Responsive d3 text

troyblank:
This is not perfect as far as text size and cropping but builds out a way we can keep svg text from shrinking with viewBox changes.


components/visualizations/line.js
```diff
@@ -100,19 +115,21 @@ class Line extends PureComponent {
...
 
     render() {
-        const { lines } = this.props;
+        const { id } = this.props;
 
         return (
-          <React.Fragment>
-            { lines }
-          </React.Fragment>
+          <div id={id} />
```
jasonhale: 💯 
One thing to think about - 'id' is used a lot for data, and programmatically assigned identifiers.  In the past, i've seen this as `htmlId` to note it's intent: `<div id={htmlId}`.  It seems odd, but cut down on confusion.  Do you think that'd be helpful here, or not necessary?

components/visualizations/line.js
```diff
@@ -2,46 +2,61 @@ import React, { PureComponent } from 'react';
...
+    componentDidUpdate() {
+        ReactTooltip.rebuild();
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener('resize', this.updateChartText);
+    }
+
+    // eslint-disable-next-line space-before-function-paren
+    buildChart  /* istanbul ignore next: d3 chaining will not render to jsdom in react currently */ () {
```
jasonhale: This should be an arrow function, since you're accessing Line's `this`.

jasonhale: May technically not need it since it's only being called within the lifecycle method's.  Thoughts?

components/visualizations/line.js
troyblank: I guess I don't see the difference. The `id` set for the component could be used for data or id attributes in the same regard. In this case we are only using it for html but that does not mean it has to be solely used for that. It just means this component's unique identifier is this. I don't see that as being confusing from my perspective, particularly because the user of the component does not have to worry about what the id does, just that it makes it unique.

components/visualizations/line.js
troyblank: Eh I like to stay consistent, we have been using arrow functions usually. I will update. A side note though is sometimes arrow functions are harder to unit test, but I don't think that is the case here.

troyblank: @jasonhale  updated, thanks!

components/visualizations/line.js
jasonhale: 👍 cool thanks

PR #934, opened 2019-01-29
Stack chart

troyblank:
Just a baby step into the stack graph.

screen shot 2019-01-29 at 5 34 22 pm


components/visualizations/stack/stack.js
```diff
@@ -0,0 +1,129 @@
...
+            });
+            if (dTotal > maxValue) maxValue = dTotal;
+        });
+
+        return { maxValue, maxPercent: (maxValue / total) * 100 };
+    }
+
+    // eslint-disable-next-line space-before-function-paren
+    buildChart  /* istanbul ignore next: d3 chaining will not render to jsdom in react currently */ () {
+        const { id, data, keys } = this.props;
```
tVeratti: What if we generated the keys from the given data?

components/visualizations/stack/stack.js
```diff
@@ -0,0 +1,129 @@
...
+
+    // Gets max value of any date and any key and max percent of any date data based on total of values
+    getMaxValues = () => {
+        const { data, total, keys } = this.props;
+        let maxValue = 0;
+
+        data.forEach((d) => {
+            let dTotal = 0;
+
+            keys.forEach((c) => {
```
tVeratti:
```suggestion
            const dTotal = keys.reduce((total, current) => total + current);
```

components/visualizations/stack/stack.js
```diff
@@ -0,0 +1,129 @@
...
+            .attr('class', (d, i) => `stack-area color-vis-${themeA[keys.length - 1 - i]}-1`);
+
+        // Cleanup all unused d3 tick markup
+        xAxisLine.select('.domain').remove();
+        xAxisLine.selectAll('.tick line').remove();
+        yAxisLine.select('.domain').remove();
+        yAxisLine.selectAll('.tick line').remove();
+    }
+
+    updateChartText = () => {
```
tVeratti: Might consider debouncing this since it does query the dom (slow) and when resize fires, it goes nuts.

components/visualizations/stack/stack.js
troyblank: Yea I thought about this, but not fully sure what the data will look like yet. Not sure if we will have to filter out a date prop or flatten the object on our own to prepare for d3. I was sorta paralyzed with the many options the payload could be in the future. I could given the data I made up filter out the date and generate the keys on the fly. I will do that as it will most likely be the end result.

components/visualizations/stack/stack.js
troyblank: Yea very true! I will update this and incorporate the other great suggestions and holler back. Thanks.

troyblank:
@tVeratti OK thanks again, updated base on your feedback!

components/visualizations/donut/donut.js
```diff
@@ -64,17 +65,15 @@ export default class Donut extends PureComponent {
...
 
         // Draw donut segments
         segments.selectAll('path')
             .data(pie(data))
             .enter()
             .append('path')
-            .attr('class', (d, i) => `donut-chart__segment color-vis-blue-${i + 1}`)
+            .attr('class', (d, i) => `donut-chart__segment color-vis-${themeA[i]}-1`)
             .attr('d', segmentArc)
             .attr('data-tip', d => sanitizer.sanitize(`<label>${d.data.label}</label><div>${d.value}%</div>`))
```
jasonhale: These come out on the tool-tips as crazy high percentages (is the value with a '%' after), instead of a percentage of the total.  Not sure the expected output... but I think you just need to remove the `%` here.
screen shot 2019-01-30 at 12 59 41 pm

components/visualizations/line.js
```diff
@@ -115,15 +117,15 @@ export default class Line extends PureComponent {
         yAxisLine.selectAll('.tick line').remove();
     }
 
-    updateChartText = () => {
+    updateChartText = debounce(() => {
```
jasonhale: 👍

components/visualizations/donut/donut.js
troyblank: Yea totally forgot to update the tooltip there when getting the stack graph going. Easy change though; updated PR.

PR #937, opened 2019-01-31
added tool tip to stack graph

troyblank:
Added tooltip to stack graph

screen shot 2019-01-31 at 1 42 18 pm


components/visualizations/stack/stack.beta.scss
```diff
@@ -5,3 +5,23 @@
...
+
+.stack-area__vertical-bounds {
+    fill: transparent;
+}
+
+.stack-area__vertical-line {
+    fill: $color-black;
+}
+
+.tack-area__point {
```
jasonhale: should be `.stack-area__point`?

components/util/number.js
```diff
@@ -13,3 +13,36 @@ export function boundNumber(number, min, max) {
 
     return number;
 }
+
+// Turns an array of numbers into a rounded array of percentages that always equal 100
+// Currently rounds to whole number, can be modified to take precision paramenter.
+export function calculatePercent(numbers) {
```
jasonhale: This is a good solution.  Sucks that it has to be so complex, though.

troyblank: Yea I know I was shocked it was as hard as it was. Lodash doesn't even have something for it.

components/visualizations/stack/stack.beta.scss
troyblank: no way, I totally meant `tack-area` ....  🙃. Thanks turns out our black is much like the default svg fill lol. Updated PR.

components/util/number.js
jasonhale: The only way I see to simplify it would be to use `Math.floor` instead of `Math.round`.  Then you wouldn't have to check direction (over or under 100) as it'll always =< 100.

PR #938, opened 2019-01-31
Stack chart legend

troyblank:
Added legend to stack chart.

screen shot 2019-01-31 at 5 29 55 pm


components/visualizations/donutStack.beta.scss
```diff
@@ -7,3 +7,30 @@
...
+    display: flex;
+    justify-content: center;
+    margin: (-$space-lg) (-$space-md) $space-md;
+    padding: $space-sm 0;
+    border-bottom: 1px solid $color-grey-light;
+    font-size: 1rem;
+}
+
+.donut-stack-legend__item {
+    margin-right: $space-md;
```
jasonhale: If we used `margin: 0 $space-md/2;` here, we could remove the `&:last-child` bit.

PR #944, opened 2019-02-06
Donut line graph

troyblank:
Added donut line graph.

screen shot 2019-02-06 at 3 58 13 pm


components/card/card.beta.scss
```diff
@@ -7,7 +7,8 @@
     border-radius: $border-radius;
 }
 
-.card__head {
+.card-head {
```
tVeratti: So I get what you're doing here, since BEM kinda makes these things less obvious at the top level, but I don't think it's technically correct to their recommendations. I'm actually not against naming things this way, but maybe there's a third option that's even better we might explore.

In their docs I remember specifically the example of a person. They said wouldn't write `person__head__eyes` but instead `person__eyes` and I think we're running into the same issue where we want the context of the middle component.

Mostly food for thought, but I'm not gonna insist that this changes since I don't have a better suggestion atm.

components/form/select/select.js
```diff
@@ -3,9 +3,9 @@ import { isEqual, isEmpty } from 'lodash';
 import classNames from 'classnames';
 import onClickOutside from 'react-onclickoutside';
 import { SelectDropDown } from '../../index';
-import styles from './select.scss';
 import { getLayoutInformation } from '../../selectDropDown/selectDropDown';
 import { IconCaret } from '../../icons';
+import './select.beta.scss';
```
tVeratti: I love that you converted this to non-modularized css. Extra credit :D

components/visualizations/donutStack.js
```diff
@@ -2,25 +2,24 @@ import React from 'react';
 import { Card } from '../';
 import Donut from './donut/donut';
 import Stack from './stack/stack';
+import Legend from './legend/legend';
+import AxisControls from './axisControls/axisControls';
 import { themeA } from './colors';
```
tVeratti: I think React Context might be great for dealing with theming! Let me know if you'd like some more information about that :)

components/card/card.beta.scss
troyblank: Yea I run into this a lot; most of the time as a group I thought we decided on starting a new context. But yea something we probably need to think about more I am cool with either way.

components/visualizations/donutStack.js
troyblank: Yea def! I can put a PR together that converts it. I can also look up best practices but any help is always appreciated. I just scribbled it down as something to smooth out along with some other stuff on my notepad :D

components/form/select/select.js
troyblank: boom boom boom let me hear you say wayo

PR #947, opened 2019-02-08
remove motion from scroller and actionPane

tVeratti:
Please fiddle with this locally, I'm not sure if I'm forgetting to check certain workflows that may not behave normally after this change. I'm hoping that I've been able to remove motion and refer to props for the actionPane and bypass all of the animation housekeeping.

Note, this is for the action pane. Since users were seeing it for both action and timeline, this change will allow us to have a comparison of behavior when they go to test it (scroll working on one and not the other).

There were some pretty suspect things that I have removed in the process of this (like measuring form height and setting it on the parent div only on certain update conditions). If we end up wanting to re-implement the animation stuff, we might be best off redoing it from a clean perspective and not use the stuff that has now been bypassed. We can discuss that further later.

NOTE: I am having a really difficult time with the tests. They are failing but I'm not able to see what is causing the `.contains` to return false. Any thoughts there would be great :)


troyblank:
@tVeratti  sorry for the rebase issues, I cherry picked this commit and just put it on the branch `release/5.16.7-alpha`. So work from there now :)

PR #950, opened 2019-02-11
TFW-2804 - derive row count from results on Table

tVeratti:


components/table/table.js
```diff
@@ -7,9 +7,9 @@ import { injectNumberCommas } from '../util/string';
 import styles from './table.scss';
 import Cell from './cell';
 
-export function formatPageInfo(pageInfo) {
-    const pageStart = (pageInfo.size * pageInfo.offset);
-    let pageEnd = pageStart + pageInfo.size;
+export function formatPageInfo(pageInfo, pageSize) {
+    const pageStart = (pageSize * pageInfo.offset);
```
jasonhale: I haven't run this - but doesn't look like the math works out for the last page, if less than the page size is returned.
Example, of 59 found items, a page size of 10, on the last page only 9 are returned.
Since you're using the response length as a count...
```
const pageStart = (9 * 5);  // = 45.
```
... but the last page would actually start at 51... not 45.

I'd suggest keep `pageStart = pageInfo.size * pageInfo.offset;`
but do a comparison if your calc'd pageSize (returned rows) is less than the set page size, then calc pageEnd.

Let me know if my logic is off here.

components/table/table.js
```diff
@@ -69,7 +69,7 @@ export default class Table extends Component {
                 { totalInteractions && ` / ${injectNumberCommas(totalInteractions)}` }
                 {' interactions'}
               </div>
-              <div>
+              <div className={styles.table__showing}>
```
jasonhale: I don't see this style anywherez

tVeratti: It's just for the tests so that I'm not searching for base elements (fragile) like '.header > div:nth-child(2)'

troyblank: I missed this. Not sure if adding classes for unit tests only makes sense.....got me pondering.

tVeratti: Well the alternative is to query on base elements which makes the test fragile and it no longer tests that the text values are correct, but also that they are in the exact div/position as before. Which is redundant because we already have a test that checks for this. It's not something that gets repeated a lot so it doesn't clutter a ton of list items.

I'm open to discussing doing the alternative, but personally that would be making tests more fragile in favor of making one line of code shorter.

troyblank: Yea I mean it makes sense I can't think of a better way. But my initial thought was an env variable that renders classes on unit tests only so that we optimize the end result.

jasonhale: Yeah.  It's definitely a tricky situation.  I see why you added it, though.
However, since it's not in the main styles, it won't get rendered in build.  Only for tests since it's in the fakes.
A little confusing, to be sure.  But I don't have a better suggestion at the moment.

troyblank:
Oh yeah good call Jason, if it was a beta style though we might need a small util that only renders during testing maybe!

components/table/table.js
troyblank: Oh yeah good call Jason, if it was a beta style though we might need a small util that only renders during testing maybe!

PR #957, opened 2019-02-19
TFW-2856 - Set print font

jasonhale:
The InterUI font we're using has a problem printing on certain versions of Windows, where the font loads but doesn't show any characters in the resulting PDF.
This just sets the print font to a stable, widely available font.

Current state (print previews):
interui print font

After change (print previews):
changed print font


assets/sass/partials/_typography.scss
```diff
@@ -23,6 +23,10 @@ body, button, input, optgroup, select, textarea {
     letter-spacing: -.02rem;
     // For some reason safari defaults to slnt 10;
     font-variation-settings: 'slnt' 0; // 0 - 10
+
+    @media print {
+        font-family: Arial, Helvetica, sans-serif;
```
troyblank: We are using sans-serif for backup; `font-family: 'Inter UI', sans-serif;`

I assume windows is loading Inter UI but not printing it. Why are we not using `sans-serif` here?

troyblank:
@jasonhale  why is the normal (non print) font different in those screen shots?

assets/sass/partials/_typography.scss
troyblank: Also a note about the bug here would be nice, we need to move back once this is fixed, changing fonts on printing is not optimal here.

troyblank:
Nevermind about my comment, I thought the right side of each screen shot was the NON print view. It is just another print view.

assets/sass/partials/_typography.scss
jasonhale: I chose that print font stack so we could be definitive since InterUI is very similar to Arial and Helvetica (spacing, et al) and wouldn't have to debug printing issues if customer were using non-standard sans-serif fonts as their defaults.

assets/sass/partials/_typography.scss
```diff
@@ -18,11 +18,17 @@ html {
...
     color: $color-black;
-    font-family: 'Inter UI', sans-serif;
+    font-family: 'Inter UI', Arial, Helvetica, sans-serif;
     font-size: 1.4rem;
     letter-spacing: -.02rem;
     // For some reason safari defaults to slnt 10;
     font-variation-settings: 'slnt' 0; // 0 - 10
+
+    // There are issue printing with Inter in certain versions of browsers and versions of Windows.
+    // Removing Inter from the font stack for printing for reliability.
```
troyblank: This is great, any chance we can mention the exact windows version we know about and how it's fixed on newer versions of windows? this will help us look at analytics when we decide to move back and determine it's a good move. I know google analytics does his, really hoping pendo does but last time I checked I don't think it does.

PR #964, opened 2019-02-20
TFW-2886 - Modify page breaks for printing Interaction

jasonhale:
When printing an interaction, the timeline started on a new page, instead of below the header info.  Just took out an extraneous page-break.

Print out after changes:
[RCI components _ Interaction Editor - View Mode.pdf](https://github.com/topbox-rci/rci-components/files/2886634/RCI.components._.Interaction.Editor.-.View.Mode.pdf)

A pdf illustrating the issue is [attached to the ticket](https://topbox.atlassian.net/browse/TFW-2886).


jasonhale:
I stopped short of that change because the interaction viewer has been temperamental lately and we have some redesign/updates on the near horizon.

But, yeah, I can make those updates. No prob.

jasonhale:
@troyblank, BEMification complete... at the top level, at least.

components/interactionEditor/interactionEditor.beta.scss
```diff
@@ -0,0 +1,147 @@
...
+    height: 100%;
+
+    @media print {
+        display: block;
+        height: auto;
+    }
+}
+
+.interaction-editor__timeline,
+.interaction-editor__actionpane {
```
troyblank: This could be moved to it's own scss file since it has it's own component.

jasonhale: Those are the wrapping `div`s for TimelineScroller and ActionpaneScroller.  Can't move those at the moment.

troyblank: oops sorry I wrote this in the wrong spot, I meant for it to be at `.scroller--actionpane`

components/interactionEditor/interactionEditor.beta.scss
```diff
@@ -0,0 +1,147 @@
...
+    @media print {
+        display: block;
+        width: 100%;
+        height: auto;
+        margin: 0;
+        overflow: visible;
+    }
+
+    .scroller--timeline,
+    .scroller--actionpane {
```
troyblank: This could be moved to it's own scss file since it has it's own component.

jasonhale: ha!  i'm in the process of doing that right now.

components/interactionEditor/scroll/scroll.beta.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../../assets/sass/base/base';
+
+.scroll {
```
troyblank: This name is pretty generic, do you think it will conflict?

jasonhale: It is.  And may.  Had that concern, too.  "Scroller" a better, non-generic name?

PR #966, opened 2019-02-24
TPB-1527 - allowed line charts to scroll horizontally

troyblank:
This allows line charts to horizontally scroll. Also makes the points fixed width distances.

You can see this code running as a demo here > https://s3-us-west-2.amazonaws.com/rci-components/pages/searchVisualizations.html


components/visualizations/line/line.beta.scss
```diff
@@ -3,14 +3,33 @@
...
-    width: 100%;
-    max-width: 1000px;
+    display: flex;
+}
+
+.line-chart__y-axis {
+    fill: $color-black;
+}
+
+.line-chart-y-domain-seperator {
```
jasonhale: In keeping up with out BEM naming convention, shouldn't this be `line-chart__y-domain-separator`?

troyblank: I was trying to avoid `line-chart-y-domain-seperator--active` turn into `line-chart__y-domain-seperator--active` .... but that's silly. Will change.

jasonhale: Nah. I get it. Names are getting long. Thanks!

troyblank: @jasonhale ok updated! Thanks!

PR #970, opened 2019-02-26
 TFW-2951 - added expandable button wrapper component

troyblank:
screen shot 2019-02-26 at 3 10 35 pm


components/button/expandableButtons.beta.scss
```diff
@@ -0,0 +1,35 @@
...
+
+    .button__text {
+        max-width: 0;
+        margin: 0;
+        transition: max-width $animation-time-crisp $animation-curve-crisp, opacity $animation-time-crisp $animation-curve-crisp;
+        opacity: 0;
+        overflow: hidden;
+    }
+
+    .button:hover {
```
tVeratti: Do you have this split out from `.button` to avoid 4th nesting layer?

troyblank: Yes and I generally try to get css as flat as possible within reason :D

PR #971, opened 2019-02-26
TFW-2945 - update Universal Search input styles

tVeratti:
screen shot 2019-02-26 at 3 21 35 pm

This does not including moving all of the content to the header - this just changes the whole input/dimension helpers to the new style.


components/selectDropDown/selectDropDown.js
```diff
@@ -444,6 +444,7 @@ export default class SelectDropDown extends PureComponent {
             className={classNames(
                 'select-drop-down',
                 styleVariationsClasses,
+
```
troyblank: Why a blank line here?

components/universalSearch/helperSwitch.js
```diff
@@ -58,12 +58,19 @@ class HelperSwitch extends PureComponent {
 
         const helperWrapperClassName = classnames(
             styles['helper-switch__helper'],
-            styles[`helper-switch__helper--${type}`]
+            styles[`helper-switch__helper--${type}`],
```
troyblank: we can probably `.beta.scss` this mug.

components/universalSearch/helperSwitch.scss
```diff
@@ -1,19 +1,30 @@
 @import '../../assets/sass/base/base';
 
+$breakpoint-mobile: 970px;
+$helper-height: 5.2rem;
```
troyblank: Why do we need explicit heights here? Shouldn't the content + padding make up the height?

components/universalSearch/searchBar.scss
```diff
@@ -13,178 +11,53 @@ $breakpoint-mobile: 970px;
...
-        }
-
-        opacity: 1;
-
-        input {
-            cursor: text;
-        }
-    }
+    flex: 1;
+    margin: 0 $small-spacing 0 $itty-spacing;
```
troyblank: `$itty-spacing;` is legacy spacing, we should probably use the `*8` stuff here. Same on line 56

components/selectDropDown/selectDropDown.js
tVeratti: Not intentional :) Will remove.

components/universalSearch/helperSwitch.js
tVeratti: Yeah might be a good time to just get it done.

components/universalSearch/helperSwitch.scss
tVeratti: When we have proper inputs everything lines up - but some of the other helpers that are dropdowns don't actually get as tall. We had explicit heights before as well.

components/universalSearch/searchBar.scss
tVeratti: Damn actually I have referenced the legacy vars throughout this pr, will update them all.

components/universalSearch/helperSwitch.scss
troyblank: Cool makes sense we don't want stuff to shift. Would a `min-height` be better instead of `height` case of a content explosion? I worry that content becomes inaccessible with explicit heights some times.

components/universalSearch/searchBar.js
```diff
@@ -134,10 +137,10 @@ class SearchBar extends PureComponent {
     }
 
     render() {
-        const { dimension, value, dimensionInputFocused, queryIdCount } = this.state;
+        const { dimension, value, queryIdCount, dimensionInputFocused } = this.state;
         const { popularDimensions, dimensions, queries, dimensionValues, messages } = this.props;
 
-        const dimensionPickerButtonText = dimension || messages.SEARCH__DIMENSION_PLACEHOLDER;
+       // const dimensionPickerButtonText = dimension || messages.SEARCH__DIMENSION_PLACEHOLDER;
```
jasonhale: Commented for a reason?  or meant to be removed?

components/selectDropDown/selectDropDown.beta.scss
```diff
@@ -84,8 +90,10 @@ $select-drop-down-separator-color: $color-grey-4;
 }
 
 .select-drop-down__label {
-    padding: $medium-spacing $tiny-spacing $tiny-spacing;
-    color: $color-blue-1;
+    padding: $medium-spacing;
+    padding-top: $small-spacing;
+    padding-bottom: $tiny-spacing;
```
jasonhale: You're already using the shorthand above.  Why not just use that?
`padding: $small-spacing $medium-spacing $tiny-spacing;`

components/universalSearch/searchBar.js
tVeratti: Ah meant to be removed, thanks for spotting that!

components/selectDropDown/selectDropDown.beta.scss
tVeratti: Eh I personally dislike that shorthand, but will update it since technically that matches the code style guide :D

PR #972, opened 2019-02-26
TFW-2935 - updated universal search to take new incoming dimension structure

troyblank:


components/universalSearch/searchBar.js
```diff
@@ -110,27 +110,27 @@ class SearchBar extends PureComponent {
...
+        const attributes = this.getDimensionAttributes(dimension);
         const data = {};
 
-        // Set all default values for types if set in
+        // Set all default values for attributes if set in
         // helperMap. This is important especially for
         // channel which might not receive a change and
         // be submitted with default value.
-        types.forEach((type) => {
+        attributes.forEach((type) => {
```
tVeratti: Technically you might want this to be `attribute` but I'm not overly concerned if you want to leave it as-is.

troyblank: Ah, I fixed this! @jasonhale  mind re-approving when you got a moment?

PR #973, opened 2019-02-27
TFW-2919 - add PropTypes to Universal Search components

tVeratti:
I also updated webpack to run in `development` by default, which is the mode needed for PropTypes to issue their warnings.


components/universalSearch/resultsGrid.js
```diff
@@ -1,12 +1,42 @@
 import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
 import styles from './resultsGrid.scss';
 import { Table, Paginator } from '../';
 
 export default class ResultsGrid extends PureComponent {
+    static propTypes = {
+      results: PropTypes.arrayOf(PropTypes.shape({
```
jasonhale: Looks like `results` is an object.  More like:

```js
results: PropTypes.shape({
  rows: PropTypes.array,
  total: PropTypes.number.isRequired,
  offset: PropTypes.number.isRequired,
  totalInteractions: PropTypes.number.isRequired,
})
```

components/util/propTypes.js
```diff
@@ -0,0 +1,6 @@
+import PropTypes from 'prop-types';
+
+export const optionPropType = PropTypes.shape({
+    label: PropTypes.string.isRequired,
+    value: PropTypes.string
+});
```
troyblank: Surprised linter did not find these... but should have a new line at end of file.

tVeratti:
This is kind of a test-drive for prop-types. We can use them here for visualization/universal search and decide if they help us enough to enforce using them in the future, and to also learn any lessons before going global in the app with them.

tVeratti:
@troyblank @jasonhale updated with your notes!

PR #974, opened 2019-02-27
 TFW-2954 - added seperate buttons for omit and add to universal search

troyblank:
screen shot 2019-02-27 at 12 55 04 pm


jasonhale:
Are we concerned about small screens yet?  The buttons are very narrow.

screen shot 2019-02-27 at 4 07 54 pm

Aside from that, looking good.

components/universalSearch/searchBar.js
```diff
@@ -181,13 +185,28 @@ class SearchBar extends PureComponent {
...
               </div>
-              <SelectButton
-                type={'submit'}
-                options={[messages.SEARCH__ADD, messages.SEARCH__OMIT]}
-                onClick={this.onSearchQueryAppend}
-                disabled={!dimension || !value}
-                ariaID={'universalSearchAppendButton'}
-              />
+              <div className={'search-bar__buttons'}>
+                <ExpandableButtons>
```
tVeratti: I almost wonder if you could define some props on the `ExpandableButton` to control all of the buttons within like `disabled` and `styleVariations` - it might come in handy and reduce the code needed to get this consistent :) I'd say not something that has to get done with this PR, but maybe sometime soon.

test/components/universalSearch/searchBar.spec.js
```diff
@@ -32,48 +27,11 @@ describe('Universal Search - Search Bar', () => {
...
-                type={'submit'}
-                options={['Add to search', 'Omit from Search']}
-                onClick={instance.onSearchQueryAppend}
-                disabled={true}
-                ariaID={'universalSearchAppendButton'}
-              />
-            </div>
-          </form>
-        ));
+        assert.isTrue(wrapper.find('.search-bar').exists());
```
tVeratti: I know that I'm the person that originally pushed for tests like this, but I'm starting to think maybe a middle ground would be best. Like testing that it contains each section individually, mainly to test that the right props are going through. Something like:

```
assert.isTrue(wrapper.contains(<AutoComplete blah={blah} />);
assert.isTrue(wrapper.contains(<HelperSwitch blah={blah} />);
// etc, etc...
```

This way we don't have overly fragile tests, but we still confirm the right components get the right props. Thoughts?

components/universalSearch/searchBar.js
troyblank: That's a great thought, I can def do that currently s both of those props are used by both buttons. Will do when I update this PR.

test/components/universalSearch/searchBar.spec.js
troyblank: Yea that sounds good, the reason I went to this is because I am passing icon component literals to the buttons which was failing on `contains`. I will go back and make them more like your example seems like a happy mid area anyway between the two approaches.

troyblank:
@tVeratti  and @jasonhale  I updated this PR with all your great comment adjustments!

screen shot 2019-02-28 at 10 25 49 am

PR #975, opened 2019-02-27
TFW-2945 - updated height of inputs

tVeratti:
I changed it back to `height` instead of `max-height` so that its style more closely matches the inputs which it mimics. Especially since it's going to contain strange children (other than pure text) we don't want anything stretching it out.


assets/js/mockData/dimensions.js
```diff
@@ -1,90 +1,90 @@
 const dimensions = [
-    { name: 'x', label: 'CategoryAllLevels', type: ['autocomplete'] },
```
troyblank: thanks!

PR #978, opened 2019-03-01
TFW-2949 - Update search node styles

jasonhale:
This updates the search nodes to latest pattern, and adds the requisite "X" icon from the UI patterns into the library.

screen shot 2019-03-01 at 11 02 23 am

Also "beta-fied" styles for these components.


components/queryNodeList/queryNode.js
```diff
@@ -20,25 +16,21 @@ export default class QueryNode extends PureComponent {
         } = this.props;
         const prependSymbol = styleVariations.includes('exclude') ? '-' : '+';
 
+        const styleVariationClassnames = 0 < styleVariations.length ? styleVariations.map(styleVar => `query-node--${styleVar}`).join(' ') : '';
```
tVeratti: Can we simplify this to `styleVariations.length ? ...`?

jasonhale: sho' nuff. and done.

tVeratti:
One last thing to check out is the smaller sizes :)
screen shot 2019-03-01 at 1 31 45 pm

jasonhale:
Good catch.  Content won't wrap now.
screen shot 2019-03-01 at 2 19 32 pm

PR #979, opened 2019-03-01
TFW-2917 - introduce Context to Universal Search

tVeratti:
This PR is up first for discussion.

Do we like the tradeoffs of using Context? PropTypes become less usable at the higher levels to check things being passed through, but we gain the reduced code clutter of passing props through and a Context which a component can subscribe to at any level and have access to the data coming from RCI.

Will fix tests if we all agree we like this approach.


troyblank:
I think the tradeoff is a good one, I would personally would put having less of the react passing and passing and passing and passing of things than the advantages of the ease of PropType usability. It makes component declaration more easy to read for sure. 👍

jasonhale:
Demo isn't working for me at the moment.  Not sure if it's due to context or just things not wired back up.  :)

Overall, I'm torn.  Context is promising.  I see it more useful as an _app-wide_ feature.  Although, here soon, Universal Search is going to be more complex and Context would be more beneficial at that time.

Some of the benefits seen here can be accomplished to some extent with strategic destructuring.  But as US grows - especially with the upcoming visualizations - that would be a less viable solution.

While I'm still kinda torn on the whole deal, I wouldn't mind seeing it implemented and in use to learn if it will help the US feature as it grows with visualizations.  And though it kinda sucks, the lack of proptypes isn't blocker for me.

jasonhale:
All looks good to me.  Just missing a tiny bit of code coverage and the conflicts.

jasonhale:
> Also I noticed this difference from develop branch which baffles me why it's happening but the content on universal search is getting scrunched and does not take up the full content section.

Looks like `links` aren't getting passed.  Causing the IDs to be far left aligned.

tVeratti:
This has been a giant pain to maintain with conflicts and stuff over time. I had to deal with some big conflicts when rebasing so if you could mostly double check that your header pane changes (@jasonhale) still look good after the rebase, that would be awesome!

troyblank:
Everything looks good to me besides for the content "scrunch" issue is still there that I mentioned above.

troyblank:
Nevermind after talking with @jasonhale  it appears the issue I found was on the release branch and is not introduced here. I compared with develop earlier, my mistake.

PR #980, opened 2019-03-04
TFW-2941 - Move US SearchBar into Header

jasonhale:
Moves the HeaderPane out of the base layout, updates some styles accordingly, and adds the SearchBar into the Header within Universal search only.

Also added some page layout styles from the new design pattern for padding on pages.

This work is required for TFW-2941 on the RCI repo.

Will clean up the commits after PR is approved and ready to go.


assets/sass/base/_mixins.scss
```diff
@@ -50,3 +50,24 @@
...
+    position: relative;
+    width: 100%;
+    max-width: $max-content-width;
+    height: 100%;
+    margin: 0 auto;
+
+    @include breakpoint($breakpoint-size) {
+        @include page-gutters('small');
+    }
+}
```
troyblank: A couple issues I noticed with this; The header seems to have too much spacing at the bottom compared to the design. Also the rocket in mobile view is now off the bottom of the page where before it sat at the bottom.

screen shot 2019-03-05 at 10 10 09 am

screen shot 2019-03-05 at 10 11 30 am

jasonhale:
@troyblank - pulled in latest from `develop` and updated a few styles.  Also modified the "clear" button to match the design docs.  Talked to Casey, it's a one-off style.

screen shot 2019-03-05 at 11 54 19 am

PR #983, opened 2019-03-05
TFW-3082 - update universal search to pass dimension objects instead of strings

tVeratti:


jasonhale:
The demo isn't working as expected just yet... popular dimensions are still being searched for based on `name` instead of `id`.  Other than that tiny thing, all looks good.

PR #984, opened 2019-03-06
TFW-3083 - no link option for tab navigation buttons

troyblank:
This builds on PR #982

- Does not render `<Link />` if the `location` prop is not present on `<TabNavigation />`.
- Introduced a prop `activeValue` to compare active buttons when `location` is not present.


components/navigation/tabNavigation/tabButton.js
```diff
@@ -79,22 +98,12 @@ export class TabButtonComponent extends PureComponent {
...
-                  })}
-                >
-                  { icon &&
-                    <div className={'tab-button__icon'}>
-                      {React.cloneElement(icon, iconProps)}
-                    </div>
-                  }
-                  { label }
-                </div>
+                { this.renderButtonGuts(iconProps, active) }
```
jasonhale: The `<Link />` component wraps its children with an anchor element, while this `renderButtonGuts()` returns a button... resulting in `<a><button/></a>`, which is bad juju wrapping an interactive element (button) with another interactive element (anchor).

what about relgating `renderButtonGuts()` to rendering the actual guts, leaving the wrapping _Link_ or _button_ in the jsx ternaries?

troyblank: Yea I agree I actually fixed this in my current work as I came to the same conclusion, I will go back and update this and let you know when it's updated.

troyblank: @jasonhale I just updated this; I didn't use ternaries because it has to be more of a switch but it does not put the button in the link which I feel is what the main issue is.

components/navigation/tabNavigation/tabButton.js
```diff
@@ -79,22 +94,23 @@ export class TabButtonComponent extends PureComponent {
...
-                      'tab-button--active': active
-                  })}
-                >
-                  { icon &&
-                    <div className={'tab-button__icon'}>
-                      {React.cloneElement(icon, iconProps)}
-                    </div>
-                  }
-                  { label }
+                <div className={classNames('tab-button', { 'tab-button--active': active })}>
```
jasonhale: Could take this even further... this `<div>` isn't really necessary, as the styles could just be added to the `Link` itself... cleaning up the markup a little more.  Whatcha think?
Like:
```
<Link to={value} className={classNames('tab-button', { 'tab-button--active': active })} onClick={this.onClick}>
  { this.renderButtonGuts(iconProps) }
```

troyblank: Ah good call, updated!

PR #986, opened 2019-03-07
TFW-2942 - Update Interaction Count style on Universal Search

jasonhale:
Separated the header from Table since the US layout won't be using it.
Then added in the updated interaction total and label.

Bonus items:
- updated the showing label from "Showing" to "Results"
- updated typography for subtitles to match new design pattern. https://projects.invisionapp.com/d/#/console/15214437/316347201/preview
  - required and update to Cards and will require an update to the omni error page in RCI.


components/card/card.beta.scss
```diff
@@ -9,7 +9,7 @@
 
 .card-head {
     display: flex;
-    padding: $space-sm $space-md;
+    padding: ($space-md / 2) $space-md;
```
troyblank: I thought all our stuff was supposed to fit in a grid of 8.... did we ask casey why we had to do this?

troyblank: Another thought; are we sure this doesn't need to be delegated to `line-height` since it is not divisible by 8?

jasonhale: It was because padding is being used to create the height.  Yeah, can go back and modify.  Will double-check the effect of this change on items added to the bar via children/props.

jasonhale:
That spacing was just missed because i was focused on the labels and what they immediately related to.  Good catch, though.  Will add.

jasonhale:
@troyblank - updated.

PR #987, opened 2019-03-11
TFW-3122 - fix to US page scroll

jasonhale:
Introduced a vertical scrollbar placement problem with TFW-2941.  This fixes that issue.

Screenshot of initial prob:
Scrollbar on US


assets/sass/base/_mixins.scss
```diff
@@ -64,7 +64,7 @@
     position: relative;
     width: 100%;
     max-width: $max-content-width;
-    height: 100%;
```
troyblank: This causes the issue that the welcome graphic won't sit at the bottom of the screen.

Screen Shot 2019-03-11 at 12 28 15 PM

jasonhale: Awesome, thanks

jasonhale:
@troyblank - ok, updated.  Thanks for catching.  I reverted the style changes introduced into the splash page component (also, it would've modified the splash as it is on Workflow), and then added a container in US to get it layed out correctly.

troyblank:
@jasonhale sorry man just one more small area it's not flush to the bottom.

Screen Shot 2019-03-11 at 2 21 23 PM

jasonhale:
No prob.  Updated.

PR #992, opened 2019-03-12
TFW-3032 - added tooltip to target score

troyblank:
Added tooltip to target score number. Also updated styles to match design.

Screen Shot 2019-03-12 at 10 16 36 AM


troyblank:
@tVeratti taking to Mr. Casey he thinks pointer is best. So I guess we should stick with it as it's the best option out of not a lot of great ones.

PR #996, opened 2019-03-13
TFW-2946 - Create expand/collapse for Search header

jasonhale:
- added UI icons for expand and collapse
- added expand/collapse button to HeaderPane
- animated opening and closing the child content of HeaderPane


jasonhale:
@troyblank - Looks like I removed the overflow during cleanup.  Thanks for catching that.

Regarding the button icons... i just don't know at the moment.  I can't get it to happen with any real consistency.  When it does happen, mousing over it causes an update and fixes the missing icon. I'll spend a little more time on it.  May have to come back to it.

jasonhale:
Boom.  Thanks @troyblank for the fix to the repaint issue.  All ready to go here.

jasonhale:
To address awkward search field auto focus - https://topbox.atlassian.net/browse/TFW-3140

PR #997, opened 2019-03-13
TFW-3038 - add sub-routing to Universal Search

tVeratti:
This work is purely to make sure sub-routing will work nicely with the toggling of views within Universal Search. The story for doing the styling/work on the navigation buttons is TFW-2958.


tVeratti:
You are correct!

PR #1001, opened 2019-03-18
TFW-2958 - Create toggle button component styles

tVeratti:
This prepares a meta section where we will need to move the results count / interaction total into.


components/navigation/tabNavigation/tabNavigation.beta.scss
```diff
@@ -7,22 +7,41 @@
...
 
-.tab-navigation__main {
-    @include breakpoint($size-small) {
-        @include stand-out-shadow;
-        position: absolute;
-        left: 0;
-        width: 100%;
-        transition: left $animation-time-crisp $animation-curve-crisp;
-        background-color: $color-white;
+    &--header {
```
troyblank: I assumed this was normal tab navigation with a card container around it. Have you looked into doing that first?

components/universalSearch/universalSearch.js
```diff
@@ -118,10 +119,11 @@ class UniversalSearch extends PureComponent {
               <div className={'universal-search__content'}>
                 <TabNavigation
```
jasonhale: TabNavigation has a required prop `onMenuSelection`.  I'd suggest to add that in here, or put in a check on tabNavigation.js to make sure it exists.
Otherwise, it breaks anytime I click on it.

components/universalSearch/universalSearch.js
```diff
@@ -118,10 +119,11 @@ class UniversalSearch extends PureComponent {
...
                   menu={[
-                  { label: 'Visualizations', value: '/universal-search/visualizations' },
-                  { label: 'Interactions', value: '/universal-search' }
+                  { label: 'Visualizations', value: '/universal-search/visualizations', icon: <IconChart /> },
```
jasonhale: I imagine it's on your list of things to do, but the path to universal search should be passed in from RCI and used here instead of hardcoding the paths.

components/universalSearch/universalSearch.js
tVeratti: Interesting, I thought that was only true for the non-link buttons? I had run into that originally but after some tweaks and making sure it was set up for href it stopped throwing. Will have to take another look.

components/universalSearch/universalSearch.js
tVeratti: It seemed a bit like overkill to me when I first went through this. The paths are hard-coded in RCI, so what is the tradeoff in maintainability? If we add it as a prop, that includes new validation, new dependency from RCI. If we leave it here, that means it needs to be updated if the `universal-search` changes as a pathname... do we realistically think that would change?

I'm up for changing it, but I'm not sure it's a winning tradeoff.

jasonhale: in RCI, the root path is saved in `BASE_PATH`, and Universal Search's page is at '/search', not '/universal-search'.
So, if you opted to hard code it, you'll have to change these paths to */cx/search* and */cx/search/visualizations*.

As far as being a dependency, i do believe it'd be a decent tradeoff.  The less need to synchronize between the repos, the less maintenance.

jasonhale: also, just realized those changes were made in a previous PR.  Sorry I didn't catch before.

tVeratti: No worries - you're right and I will fix it here!

components/navigation/tabNavigation/tabNavigation.beta.scss
tVeratti: Well it needs extra consideration for mobile styles that need to be overridden as well. Also, is there even a card style that fits? Sizes of the buttons etc need to be adjust to match the overall design so it's more than just a wrapper for now.

troyblank: Sounds good. I obviously didn't explore or test it out just raising the question.

tVeratti: Yeah it's a good question, I definitely didn't think of it first. After you asked I went and checked it out a bit and that's what led to my followup questions. I think we're okey dokey. I don't SUPER love the style variation, but there's a lot of really really relevant/good stuff in TabNavigation that I want to use and would end up just recreating for a style change.

PR #1004, opened 2019-03-18
TFW-3045 - added y axis controls to stack chart

troyblank:
This adds control to the Y axis of the line and donut stack graphs. It changes count to percent and vise versa. What this didn't do was add separate data flow for donut charts which is needed to separate `value` from `percent` and so they behave more like the line and stack graph. This was work figured out in the last meeting about the api but is something I might be able to tackle this week.

I know this has a lot of commits but there is a lot of conflicts on `release/4.41.0` to `develop` right now I can't resolve on my own. So that might need to be done first before this PR is sane.


components/visualizations/stack/stack.js
```diff
@@ -120,74 +140,90 @@ export default class Stack extends PureComponent {
...
+            yAxis.tickFormat(d => smallifyNumber(d));
+        }
+
+        // Append axis
+        if (this.ySVG) this.ySVG.remove();
+        this.ySVG = d3.select(`#yAxis${id}`).append('svg')
+            .attr('width', this.yAxisWidth)
+            .attr('height', height + marginTop + marginBottom);
+        const yAxisLine = this.ySVG.append('g')
+            .attr('transform', `translate(${marginLeft}, ${marginTop})`)
```
jasonhale: the `marginLeft` and `marginRight` threw me off initially.  I expected them to be used as inline styles.

troyblank: Yea it's actually sort of a d3 using svg idiom that is also used in their docs. It's not real margin but theoretical margin to keep the plotting right for your axis lines.

PR #1012, opened 2019-03-22
TFW-3037 - updated line graph x values when month or week is selected

troyblank:
Selecting from the temporal date drop down on the line graph now filters/groups data for the line graph.


components/util/date.js
```diff
@@ -33,4 +33,8 @@ export function formatDateRangeToFrom(range) {
     };
 }
 
+export function isLastDayOfTheMonth(date) {
+    return 1 === new Date(date.getTime() + 86400000).getDate();
```
tVeratti: What exactly is this doing to get the last date? Does it work for all months? The previous way I've done this is go up a month and then down a date, if I recall.

components/visualizations/line/line.js
```diff
@@ -4,12 +4,14 @@ import sanitizer from 'sanitizer';
...
-import { smallifyNumber, RESIZE_DEBOUNCE_DELAY } from '../util';
+import { smallifyNumber, getDataWithRepeatedPoints, RESIZE_DEBOUNCE_DELAY } from '../util';
 import { injectNumberCommas } from '../../util/string';
 import './line.beta.scss';
 
 export default class Line extends PureComponent {
     componentDidMount() {
+        const { data } = this.props;
+
         this.layout = {
```
tVeratti: I wonder if it would make sense to move more of these things into `defaultProps` - we might not touch them for now, but if we did want to make subtle tweaks to some, it would already be possible without major refactoring.

components/util/date.js
tVeratti: Nevermind I see, you're adding 24hrs to the current time and seeing if it's then the first date. Seems solid to me.

components/visualizations/line/line.js
troyblank: The problem with moving them to default props is they are stylistic in nature and are pretty tightly coupled with the css (such as font size). There is not really any changing of this object you can do without also re-tweaking the layout in css. So I don't see any value of exposing these outside the component personally. Unfortunately thought these values need to be in js to layout the svg though.

PR #1013, opened 2019-03-25
TFW-3016 - allowed line graph in donut line component to react to date filter changes

troyblank:
This builds on TFW-3037 to add date filtering to the other line graph we are using in phase 1.


components/visualizations/util.js
```diff
@@ -1,8 +1,91 @@
...
-};
+}
+
+export function addPercentsToData(data, yKey) {
+    const percents = calculatePercent(data.map(v => v[[yKey]]));
+    const newData = data.map((d, i) => ({ ...d, percent: percents[i] }));
+
+    return newData;
+}
+
```
troyblank: `filterDataByWeek` and `filterDataByMonth` are pretty similar. I am combining them into one function in the current work I am doing that also allows `yKey` to be an array for stack graphs. So if you do have concerns about duplication here I can show you the work I am doing to remedy this. But this is a good step to get to that refactor in the meantime.

jasonhale: No concerns from me at the moment.  I like these separate functions here - they're specific and easy to read.  Thanks!

PR #1014, opened 2019-03-25
TFW-3205 - Protect against null options in SelectDropDown

jasonhale:
Sometimes options in AutoComplete come across as `null`.  This prevents throwing exceptions when that occurs.


jasonhale:
Dimensions sometimes come through with `null` - according to Josh, apparently it's a buggy issue in either Admin or db.

components/form/autoComplete/autoComplete.js
```diff
@@ -10,7 +10,7 @@ export function filterOptions(value, options) {
     let regex;
 
     options.forEach((option) => {
-        const optionValue = option.label || option;
+        const optionValue = (option && option.label) || option;
```
troyblank: To me lodash's `get` is easier to read, but I know we have been avoiding it. https://lodash.com/docs/4.17.11#get ... what do you think?

jasonhale: `get` is already being used in this component.  Can easily modify it.  How it's currently coded makes sense to me because it's obvious what the logic is doing.  But, it's all opinion at this point and I'm good either way.

troyblank: I personally prefer `get` because the conditional logic is handled for you. But I think @tVeratti  likes using conditionals....  so yea I agree it's opinion but not one I would die on... if it was a sword or something so whatever. Note that as the lookup gets deeper in an object so does your conditional logic; but with `get` it never grows.

PR #1017, opened 2019-03-26
TFW-3202 - Move Interaction/Results Count to Universal Search meta section

tVeratti:
Screen Shot 2019-03-26 at 1 01 39 PM


components/universalSearch/metaData.beta.scss
```diff
@@ -0,0 +1,37 @@
+@import '../../assets/sass/base/base';
+
+.search-meta {
+    display: flex;
+    align-items: center;
+    padding: $space-md $space-xl;
+}
+
+.search-meta__divider {
```
troyblank: Why not save a DOM element and just make this the left border of the `search-meta__count` stuff?

tVeratti: The size (height) wouldn't be reliable in that case - it has to be a wee bit shorter than the rest of the content. A pseudo element could also get the job done in hindsight, maybe that's the best of both worlds.

troyblank: Yea pseudo element sounds like a solid plan to me :D

tVeratti:
Screen Shot 2019-03-27 at 10 40 57 AM

Updated the tab button to fill out the container when it goes to stacked mobile view.

components/navigation/tabNavigation/tabNavigation.js
```diff
@@ -14,7 +14,7 @@ export class TabNavigationComponent extends PureComponent {
     onMenuSelect = (values) => {
         const { onMenuSelection } = this.props;
 
-        onMenuSelection(values);
+        if (onMenuSelection) onMenuSelection(values);
```
jasonhale: 👍

troyblank:
@tVeratti  the problem seems to be still there for me above the smaller view. Sorry :(

Screen Shot 2019-03-26 at 2 45 59 PM

components/universalSearch/metaData.beta.scss
```diff
@@ -0,0 +1,62 @@
...
+    @include breakpoint-below($breakpoint-tiny) {
+        margin-right: 0;
+        padding-right: 0;
+    }
+
+    width: 0;
+    height: $space-lg;
+    margin: 0 $space-md;
+    padding-right: 1px;
+    background: $color-grey-3;
```
jasonhale: From the designs, this should be $color-grey-4 (#e5e5e5).  Scanned the design docs, separators are pretty consistently $color-grey-4;

components/universalSearch/resultsGrid.beta.scss
```diff
@@ -33,7 +33,7 @@
 }
 
 .results-grid__showing {
```
jasonhale: `.results-grid__header`, `.results-grid__count`, `.results-grid__count-label` and `.results-grid__showing` can all be deleted since they don't reference anything anymore.

tVeratti:
Removed old css, added breakpoint for mobile size to make sure the tab button has a full width between existing breakpoints.

components/universalSearch/universalSearch.js
```diff
@@ -78,7 +78,7 @@ class UniversalSearch extends PureComponent {
             const paths = this.getPaths();
             return (
               <React.Fragment>
-                {this.renderMetaData()}
+                <MetaData paths={paths} pathname={history.location.pathname} />
```
jasonhale: React Router docs recommend against using history.location as it's mutable.
https://reacttraining.com/react-router/web/api/history/history-is-mutable

PR #1024, opened 2019-03-28
TFW-2921 - Add Error Boundary to Universal Search

tVeratti:
`SearchError` will catch render-killing errors and display an InfoPanel with some suggested solutions. I'm sure once Casey is back he will have some ideas but this is my current suggestion/proof of concept.

In order to see it locally, add a failing line somewhere in the render cycle in `SearchBar` or `HelperSwitch`. I did it by adding `if (queries.length) queries.goError()` so that I could see it clearing/resetting by removing the queries array, simulating if we `RESET` the query data from the `universalSearch` reducer.

Screen Shot 2019-03-28 at 2 16 35 PM


components/universalSearch/searchError.js
```diff
@@ -0,0 +1,44 @@
...
+                  links={[
+                    { label: 'Clear Queries', onClick: this.onResolve },
+                    { label: 'Reload Page', onClick: this.onReload }
+                  ]}
+                >
+                  <h2>An error occurred while building your query.</h2>
+                  You can try clearing your queries or reloading the page.
+                </InfoPanel>
+              </div>
+
```
troyblank: Weird blank line here. Not a huge deal though.

tVeratti: Haha yeah will remove!

troyblank:
I like this a lot, we have used this style for pages that are not ready to load for some reason or not with the qlik stuff if I remember right. Informing the user something went boom is clutch; the only thing maybe we are missing for the future that I can think of is tracking when a user sees this and why, maybe we should talk about bringing pros and or cons of sentry to the front end side of things.

tVeratti:
After playing a bit with Sentry it definitely seems like something we can easily set up in RCI proper. We could easily add some meta data to the global exception scope including user/channel/whatever. I'd really like to do it sooner rather than later - not a lot of work to get really nice error logging.

PR #1025, opened 2019-03-28
TFW-3027 - updated channel donut text on percent change

troyblank:
This builds on a PR open for TFW-3039.

This allows the donut on the donut stack graph to update with y axis mode changes.

Also some fixed fixed issue where precent was showing up in stack tooltip that came out of this work.

The animation of the stack graph is jacked up and I plan on making a `yAxisMode` much like there is a `xAxisMode` instead of just changing the key for the y axis. This allows graphs like the stack graph to maintain a total outside of the data itself and display percentages as a line graph without it looking bad.


components/visualizations/stack/stack.js
```diff
@@ -47,22 +51,33 @@ export default class Stack extends PureComponent {
...
-        const { data } = this.props;
-        const { width, marginRight } = this.layout;
-        const minDate = new Date(data[0].date);
-        const maxDate = new Date(data[data.length - 1].date);
+        const { data, metaData } = this.props;
+        const { xKey } = metaData;
+        const { marginRight } = this.layout;
+        const dates = data.map(d => new Date(d[xKey]));
+        const minDate = dates[0];
+        const maxDate = dates[data.length - 1];
```
tVeratti: Just to be safe, should this be `dates.length - 1` ? It's a 1:1 for now but if any other manipulations come in it'd blow up.

components/visualizations/stack/stack.js
```diff
@@ -47,22 +51,33 @@ export default class Stack extends PureComponent {
...
 
     setYScale() {
-        const { data, metaData } = this.props;
+        const { metaData } = this.props;
         const { yKey } = metaData;
         const { height, marginBottom } = this.layout;
-        const keys = Object.keys(data[0]).filter(key => 'date' !== key).reverse();
+        const keys = this.getKeys();
         const { maxValue, maxPercent } = this.getMaxValues(keys);
         const max = 'percent' === yKey ? maxPercent : maxValue;
```
tVeratti: Should we consider having some constants defined for these keys? I've seen them referenced as literals in a few places and I'm wondering if we had `const KEY_PERCENT = 'percent'` we might protect ourselves from the api changing key names on us.

components/visualizations/stack/stack.js
```diff
@@ -243,45 +280,91 @@ export default class Stack extends PureComponent {
...
+
     updateYValues() {
         this.setYScale();
         this.buildYAxis();
         this.updateLines();
         this.updateGraphScale();
     }
 
-    updateLines() {
+    updateLines() /* istanbul ignore next: d3 chaining will not render to jsdom in react currently */ {
```
tVeratti: I really like how you have separated the logic/calculations from the presentation which can't be tested. Good thinking ahead for actually being able to test the heavier stuff.

test/components/visualizations/donutStackGraph.spec.js
```diff
@@ -13,32 +14,54 @@ describe('Donut Stack Graph', () => {
...
-              <Stack id={`${id}Stack`} data={data.values} total={total} metaData={metaData} />
+              <Donut id={`${id}Donut`} data={donutData} colors={themeA} metaData={donutMetaData} />
+              <Stack id={`${id}Stack`} data={instance.state.filteredStackData} metaData={stackMeta} />
             </div>
           </Card>
         ));
     });
 
+    it('should update x axis', () => {
+        const setState = sinon.spy(DonutStackGraph.prototype, 'setState');
```
tVeratti: I'd spy on the instance instead. Just have to do it after the instance has been created :)

components/visualizations/stack/stack.js
troyblank: Ah yes! Updated this thanks.

test/components/visualizations/donutStackGraph.spec.js
troyblank: I wish you could spy from the instance but alas you can not for some reason.

```instance.setState = sinon.spy();``` <-- fails with `setState` is undefined.

File: components/visualizations/stack/stack.js (hunk 2 above)
Comment (troyblank) [reply components/visualizations/stack/stack.js]: I agree there is two many places where it is made and it needs to be in a higher place. I will do some work to make it all reference in one location and let you know when it's done.

File: test/components/visualizations/donutStackGraph.spec.js (hunk 4 above)
Comment (jasonhale) [reply test/components/visualizations/donutStackGraph.spec.js]: I've seen the prototype spy before, and it works fine.
Can also set up the spy a little different and go:
```
...
const wrapper.instance():
sinon.spy(instance, 'setState');
...
```

This way, you can spy on `setState` using the instance instead of the prototype.

components/visualizations/stack/stack.js
troyblank: OK @tVeratti  updated to use a constant, much improved thank you!

test/components/visualizations/donutStackGraph.spec.js
troyblank: OOOOPS yea @jasonhale  is right, why the hell did `instance.setState` blow up on me before? oh well updated to use `instance` over  `prototype`.

PR #1026, opened 2019-03-29
 TFW-2997 - changed all graphs to use yAxisMode instead of changing their yKeys

troyblank:
This builds on two PRs out currently FYI.

I was using `yKey` prop to change which object key to point the y axis on when the user toggled between 'volume' and 'percent'. But after doing the stack graph in this manner I realized this way was flawed in order to get the design to match up to the front end.

The problem is the stack graph in 'percent mode' does not fully use percents to build the stack, it uses a little of volume and a little of percent as shown here >

Screen Shot 2019-03-29 at 8 59 31 AM

This PR changes the use of `yKey` to change modes to 'yAxisMode' much like the 'xAxisMode' used before. This way each graph can decide what that means for them as with the stack graph it's a lot more complicated than just pointing the data at a new key.


components/visualizations/line/line.js
```diff
@@ -104,7 +107,7 @@ export default class Line extends PureComponent {
     buildToolTip(d) {
         const { metaData } = this.props;
         const { xKey, yKey } = metaData;
-        const perecentText = d.percent ? `<div>${d.percent}% of interactions</div>` : '';
+        const perecentText = 'number' === typeof d.percent ? `<div>${d.percent}% of interactions</div>` : '';
```
jasonhale: Super minor misspelling... `perecentText` -> `percentText`

troyblank: Story of my life, updated ty ty ty.

troyblank:
@jasonhale  and @tVeratti  I addressed all commits but had to deal with merge conflicts so re-requesting a review as a formality.

PR #1028, opened 2019-04-01
TFW-3290 -  allowed line charts to be centered

troyblank:
this builds on all 3 PRs i have up right now, 2997, 3027, and 3039

This centers line charts so they are no longer left aligned if the data is too small to fill the container.

Screen Shot 2019-04-01 at 11 17 58 AM


components/visualizations/line/line.js
```diff
@@ -356,15 +353,33 @@ export default class Line extends PureComponent {
     }
 
     updateGraphScale = debounce(() => {
-        const xSvg = this.xSVG.node();
+        const { id } = this.props;
+        const contentWidth = this.getContentWidth();
         const ySvg = this.ySVG.node();
+        const xAxisContainer = document.querySelector(`#xAxis${id}`);
```
tVeratti: Should this be a ref? Could select other graphs with the same id if we had two line graphs, no?

tVeratti: If not a ref due to d3 stuff, then `querySelector` off of the root of the component (as a ref)

jasonhale: Ideally, the IDs would be unique when they get assigned to the visualization initially.  And every nested viz within would inherit and append, creating a new unique id to the page.

components/visualizations/line/line.js
```diff
@@ -356,15 +353,33 @@ export default class Line extends PureComponent {
...
-        const xWidth = xSvg ? xSvg.getBoundingClientRect().width : 0;
+        const xWidth = xAxisContainer ? xAxisContainer.getBoundingClientRect().width : 0;
         /* istanbul ignore next: d3 dom does not exist in jsdom */
         const yWidth = ySvg ? ySvg.getBoundingClientRect().width : 0;
         const scale = this.lineChart.offsetWidth / (xWidth + yWidth);
         const hasHorizontalScrolling = 1 > scale;
 
+        // Centering data
+        /* istanbul ignore next: d3 dom does not exist in jsdom */
+        if (xWidth > contentWidth) {
```
jasonhale: Not blocking, but could reduce a little redundancy here, unless you think any of these methods may change?

Suggesting:
```
let translateX = 0;
if (xWidth > contentWidth) translateX = (xWidth / 2) - (contentWidth / 2);

this.xSVG
    .transition()
    .duration(750)
    .attr('width', contentWidth)
    .attr('transform', `translate(${translateX}, 0)`);
```

troyblank: Updated this this suggestion thanks.

components/visualizations/line/line.js
troyblank: Updated to use `ref` in this case, makes more sense. Thanks.

test/components/visualizations/line/line.spec.js
```diff
@@ -132,5 +132,20 @@ describe('Line', () => {
...
     });
+
+    it('should be able to center a the line graph if its not big enough to fill container', () => {
+        const updateXValues = sinon.spy(Line.prototype, 'updateXValues');
+        const wrapper = mount(<Line data={data} id={id} metaData={metaData} />);
+        const instance = wrapper.instance();
+
+        instance.xAxis.getBoundingClientRect = () => ({ width: chance.natural({ min: instance.getContentWidth() + 1 }) });
+
+        // instance.updateGraphScale();
```
jasonhale: Got a comment left out there.  :)

troyblank: oops, ok fixed!

PR #1033, opened 2019-04-03
TFW-3299 - Fix spacing and icon placement on Search Splash for Universal Search

jasonhale:
Bugfix for RCI 4.41.
Search Splash page needed some minor updates due to some changes since it was created.
Fixed spacing on the sides and set the icon to the bottom of the window.

Also added a way to show the "No Results" error in the Universal Search demo to more easily view the search error message.

Ticket created (TPB-1748) to address solutions for the scrolling issue mentioned in TFW-3299.

Screen Shot 2019-04-03 at 9 33 54 AM
Screen Shot 2019-04-03 at 9 33 26 AM


assets/js/demo/components/universalSearch.js
```diff
@@ -16,8 +16,15 @@ export default class UniversalSearchDemo extends Component {
         print(null, queries);
         this.setState({ pendingResults: true, queries });
 
+        const queryResultsReturned = { ...queryResults };
+
+        // Allow demoing of "no results"
```
troyblank: Awesome, thanks for doing that!

tVeratti:
It seems to me that the height issue is probably separate, so we should create an issue and give it its own estimate, etc. :)

troyblank:
> It seems to me that the height issue is probably separate, so we should create an issue and give it its own estimate, etc. :)

Yea I would be cool with that, just need to make a new ticket and update the criteria of the current one since it's not a new issue.

jasonhale:
Ticket already created (TPB-1748) and sent to Maroua to track the scrolling problem. It's definitely big enough to not be addressed in a bugfix. 👍  Also updated TFW-3299 notes.

PR #1035, opened 2019-04-03
TFW-3304 - Map visualization data to React components

tVeratti:
Screen Shot 2019-04-03 at 11 38 39 AM

This story DOES:
- Get initial structure in place between API response and React components.
- Add Error Boundaries around charts to allow isolated failures.
- Build an initial mapping between string types and component classes (this will need updating as the mongodb types evolve)

This story DOES NOT:
- Massage all data as each graph will need.
- Bring in proper mock data.
- style the dashboards.


components/universalSearch/visualizations/chart.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import * as Charts from '../../visualizations';
+
+export const CHART_TYPES = {
+    line: Charts.LineGraph,
+    pie: Charts.DonutLine
+};
+
+export default function Chart({ type, encoding, values, ...props }) {
```
troyblank: But what if it's a graph? :P .... graphs and charts are different things but they are so fuzzy that I am fine with using it generically. But just food for thought. Could be more general of a name.

tVeratti: Well Visualization was kinda overused. I'm up for any ideas xD

troyblank: I am cool with chart, but really my main idea is `VizzyNizzy` pleazy?

components/universalSearch/visualizations/chart.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import * as Charts from '../../visualizations';
+
+export const CHART_TYPES = {
+    line: Charts.LineGraph,
+    pie: Charts.DonutLine
+};
+
+export default function VizzyNizzy({ type, encoding, values, ...props }) {
```
jasonhale: 😆

troyblank: it's gucci

PR #1037, opened 2019-04-05
TFW-3036 - Update design for Viz Axis Controls

jasonhale:
Fixin' styles on these axis controls so content is centered and the dropdown is overlapping the req'd 8px.

Also updated the styles to remove that border and the colon after the label.

Screen Shot 2019-04-04 at 7 20 04 PM


components/visualizations/axisControls/axisControls.js
```diff
@@ -4,6 +4,8 @@ import { KEY_PERCENT } from '../util';
 import './axisControls.beta.scss';
 
 export default function AxisControls({ id, onXUpdate, onYUpdate }) {
+    const top = 'calc(100% - .8rem)';
```
troyblank: Explain why this can't be done in css again? I know we did this before for `<Select />`'s right? Just making sure this is necessary in js.

jasonhale: Sure.  Technically, we could do it with CSS, it'd just be with some `!important` tags.

SelectDropDown uses js to absolutely place itself.  We came up with the way to pass a position to it to override that feature (mostly for the main and header navigation).  The one caveat to passing a position is that it disables SelectDropDown's auto placement when it's too close to the bottom of the screen.

So, yeah, we can
- do via CSS, and retain the feature of the dropdown flowing up when too close to the bottom of the screen.
- do via js, using the allowed ability of passing position.

And, honestly, a portion of the issue is supporting old and new styles.  If we could consolidate/update to use nothing but new design, we could alleviate this manual placement in a few spots.

jasonhale: ... or, to promote easier and more visible styling (instead of the masking that can happen when overriding with `!important`), we could allow turning off SelectDropDown's self-positioning altogether so a parent can style it more explicitly without going nuclear and overriding styles.

troyblank: Hmmm yea we had this convo before, sorry I forgot. But yea what you say makes sense and it seems like what you did is the less convoluted until we can re-do all styles to the new ones. Thanks for the explanation.

PR #1038, opened 2019-04-08
TFW-3282 - added base word cloud component

troyblank:
This is mainly just to get everything hooked up for word cloud with a "rough" prototype. Turned out to not be so rough though...

Screen Shot 2019-04-08 at 11 35 11 AM


components/visualizations/wordCloud/wordCloud.js
```diff
@@ -0,0 +1,83 @@
...
+import './wordCloud.beta.scss';
+
+export default class WordCloud extends PureComponent {
+    componentDidMount() {
+        this.layout = {
+            width: 960,
+            height: 500
+        };
+
+        this.textSizeRatio = 1.2;
```
tVeratti: Instead of having these in `didMount` you could add them at the component level so they'd be defined even for the initial render?
```
... extends PureComponent {
    textSizeRatio = 1.2;
    textPadding = 5;
    // .. etc
```

troyblank: OK, this is a great idea and I had to do a bit of linting updating to get it to work. So there is now two commits on this PR, one with the work and one with this update plus a ton of linting tweaks.

Also right now the `eslint-config` package `v1.1.1` does not exist yet, so this PR is not really testable until the PR in the config is approved and good. So you can ignore this for now and I will ask for a re-review once that tag is ready.

package.json
```diff
@@ -14,7 +14,8 @@
     ],
     "scripts": {
         "start": "grunt build && node server",
-        "test": "grunt test",
+        "test": "npm run lint && grunt test",
+        "lint": "eslint components/**/*.js && eslint assets/js/**/*.js && eslint test/**/*.js",
```
jasonhale: 💯

troyblank:
@tVeratti OK the new rules are ready to `npm i` so this is ready for a re-review thank you!

components/visualizations/wordCloud/wordCloud.js
```diff
@@ -0,0 +1,82 @@
...
+        const { id, data, metaData } = this.props;
+        const { yKey } = metaData;
+        const { width, height } = this.layout;
+
+        this.setColorScale();
+
+        this.SVG = d3.select(`#${id}`).append('svg')
+            .attr('width', width)
+            .attr('height', '100%')
+            .attr('viewBox', `0 0 ${width}, ${height}`)
```
jasonhale: I noticed there's a weird inconsistency in the viewBox values where there's a comma.  It's not hurting anything... just looks off.  Usually it's all or nothing as far as commas are concerned.

tVeratti:
Do we have any plan/design for dealing with text overlapping? Prob not important for this story since you'd said it's more of a prototype than a finished design, but I want to get it out front early.

troyblank:
@tVeratti  yea next PR solidifies that overlap, but maybe you are seeing what jason is... a bug I can't reproduce.

components/visualizations/wordCloud/wordCloud.js
troyblank: I am fixing this in my current work, so thanks for the point out.

PR #1040, opened 2019-04-09
TFW-3283 - made sure d3 cloud uses a percent size

troyblank:
so I learned d3-cloud expects the `size` prop to do it's magic so I made sure it is added no matter want and is also in the range from 0 to 100. With this code the `yKey` of wordClouds can be anything and still produce a good word cloud. Displaying the hover might have a conflict in later work if the origin data also uses `size` as a `yKey` but I will cross that bridge when I get there.


jasonhale:
Not sure what it is, but occassionally when refreshing the page, the collision detection doesn't seem to kick in...
Screen Shot 2019-04-08 at 11 45 07 PM

troyblank:
@jasonhale so @tVeratti I think saw the overlap as well, I can seem to get it sometimes when hitting the back button to the word cloud......so maybe it will be an issue :(

PR #1041, opened 2019-04-09
TFW-3327 - Add content spacing around visualization dashboard / vizzies

tVeratti:
I used this story to also introduce the idea of creating helper css classes. This could be applied to a lot of places that are including the `content-wrap` mixin, but I would wait to refactor those until we get a chance to see this live in action and agree we like the direction.

Screen Shot 2019-04-09 at 1 38 32 PM


components/helpers.beta.scss
```diff
@@ -0,0 +1,11 @@
+@import '../assets/sass/base/base';
+
+// IMPORTANT!
+// ----------------------------------------
+// This is not a partial style sheet!
+// These classes will be loaded into the sheet every time it is imported,
+// therefore it must only be loaded once as part of the main export `index.js`.
+
+.h__content-wrap {
```
troyblank: Why not actually make this a `mixin` and then you won't have to make a global class? Making the component `@include` it when it's needed. This strikes me as backwards a tad from how reusable css is thought of in terms of `mixins`.

troyblank: ooor a `silent` can be used as well but I think we don't like those as a team maybe?

tVeratti: It already is an existing mixin, but here's the issue:

Every time we declare a component that wants that `content-wrap` style, we're including the mixin code. A helper gives us that reusable style without bloating the stylesheet by re-including the mixin every single time it's used. I think they have two different use cases.

- Mixin: When we want to use parameters to affect the included code and when specificity is really complicated
- Helper: When we want a chunk of reusable style that doesn't change much from place to place where it's used.

Mixins are good, but remember that everytime we use it, we're adding all of those lines of css when you could alternatively have a single helper that gives it a base set of styles that should be easy to style alongside.

tVeratti: In fact the entire point of use a global class like this was to *not* use a mixin :)

troyblank: So what you are talking about i a `silent` which does not bloat your css.

```%red {
	color: red;
}

.thing1 {
	@extend %red;
	background: blue;
}

.thing2 {
	@extend %red;
	background: green;
}``

which turns into

```.thing1, 
.thing2 {
	color: red;
}

.thing1 {
	background: blue;
}

.ting2 {
	background: green;
}```

which eliminates the use of global css glasses but does not copy the css in multiple places. Does this seem like something that would work?

Comment (troyblank) [reply components/helpers.beta.scss]: Yea sorry the only reason I suggested `mixin` is because I think I got a lot of criticism for using `silents` in the past :P

Comment (jasonhale) [reply components/helpers.beta.scss]: The problem with `@extend`-ing things in my experience has been the issue of inheritance.  Sass will group those things that are extended and move them high up in the cascade - which can cause inheritance issues.  It's not overly prevalent but it happens often enough and bites hard when it does.

Mixins are good in that the selector has all the styles right there in the cascade where it's supposed to be.

There's definitely pros and cons to both approaches.

Comment (tVeratti) [reply components/helpers.beta.scss]: Using `extend` we're still going to end up with:
```
.workflow,
.results-grid,
.visualizations,
.qlik,
// etc
// etc
// so on, for every page we make on forward
```
If you consider a helper with ~5 lines of CSS, you'll double your code after only ~5 other styles extend it. Not a big deal for one or two, but when that's how all of our reusable styles work, you're talking about doubling (or more) a lot of the code.

I don't really understand why you see a few helper classes as a step backward. They would allow us to add/remove little chunks of styles without recreating them everywhere. I also think it's a bit easier to read in dev tools where certain styles are coming from when they're grouped under a helper class.

Comment (tVeratti) [reply components/helpers.beta.scss]: And to be clear, I don't mean to suggest we should always use helper classes over mixins or extends. Only for cases like this of frequently used bits of style that don't care where they are (no params).

Comment (troyblank) [reply components/helpers.beta.scss]: I am just trying to avoid doing two things the same way in the code base. I still don't understand the arguments for universal classes over extend as each one I have heard does not seem right in my experience.

So 1. @jasonhale  you say that `@extend` moves stuff higher up but so does using global css classes, both have issues with specificity.

Screen Shot 2019-04-10 at 7 32 51 AM


and 2. @tVeratti  adding `extend` does add a chunk to the css for each extend yes but also adding a class for each dom element adds add chunk to the js as well, and both get gzipped in the same fashion so I don't understand how that is bloat, you are just shifting the weight of css to js from what I can tell.

It's not fair I said step backward but it's more like I feel like we are fragmenting. We had one way of doing things before now we have two. I feel `extend` is more powerful because of the pros that A. you can use it on pseudo elements, which I personally do a lot and B. your markup is not as class heavy (less style declarative). If we take all our silents today and convert them into global css classes we will have a lot of new classes all over the place. So my argument is consistency, are we going to mix match ways of using reusable css or do we want to use the global class names over the silents?

I realize this topic has pros and cons both ways and there is no right or wrong, and I am fine either way but would love to talk about a way to be consistent when it comes to using reusable css code, possibly moving all our silents to global class names and if it's fine that we lose the pros I listed in the last paragraph.

Comment (tVeratti) [reply components/helpers.beta.scss]: At the end of the day, the difference here is that helpers move the logic to javascript. Doing that same work in just css means adding 1-2 more classes that are component specific when sometimes it's just wrapping a style with another classname.

I think there will be times when a helper makes sense, and times when mixins/extend make sense. There are times when we create a bunch of block/modifier classes scoped to a component that ends up being just a wrapper around a mixin. In those cases, I see helpers as the superior solution.

Just have another way to do this isn't necessarily fragmenting or throwing consistency out of the window. It's just acknowledging that there is a third way to add styles for certain scenarios that merit it. Again, it's not a replace-all solution. I personally wouldn't want to over-use helpers since that turns into a bootstrap-esque situation where classnames become huge unreadable monsters. I think peppering it in with purpose is the right move. This can remove needs for certain classes (and even stylesheets) entirely when they are defined just to call a mixin in.

Comment (troyblank) [reply components/helpers.beta.scss]: Can you expand on how this same work will mean adding 1-2 more classes? I would assume all you would have to do is this > `.search-dashboard { @extends %hor-wrap-or-whatever; ... }`

Comment (tVeratti) [reply components/helpers.beta.scss]: Yeah that's an example of one block level class added. That would be on and then further modifiers or child blocks might continue the trend.

Comment (troyblank) [reply components/helpers.beta.scss]: But same if you had to add a class to each of those children right?

File: components/universalSearch/visualizations/dashboards.js [hunk 2]
```diff
@@ -11,7 +11,7 @@ export default function Dashboards({ index = 0 }) {
     const activeDashboard = dashboards[index];
 
     return (
-      <div className={'search-dashboard'}>
+      <div className={'search-dashboard h__content-wrap'}>
```
troyblank: The class name `search-dashboard`  is not needed anymore right?

components/universalSearch/resultsGrid.js
```diff
@@ -19,7 +17,7 @@ const ResultsGrid = ({ titles, sort, onSortChange }) => {
     };
 
     return (
-      <div className={'results-grid'}>
+      <div className={'results-grid h__content-wrap'}>
```
troyblank: hate to ask for this after the long convo but does `h__content-wrap` stand for `horizontal-content-wrap`? I am assuming h === horizontal, but I am a huge fan of being explicit so that there is no confusion what something is. Any chance we would entertain making this more verbose? Also if that assumption is right the two underscores usually mean a new section in BEM, so I am confused why horizontal has a section. Could we talk about making using something so we know it's...... wait! it stands for helper right? can we spell out helper.... if that is right? so `helper__content-wrap` ... I thought about editing this but it shows my confusion a little better if I do not :P

components/universalSearch/visualizations/dashboards.js
tVeratti: Technically we could remove it, yup.

components/universalSearch/resultsGrid.js
tVeratti: It is helper, yeah. I can update it if we prefer!

troyblank: I do! it helps more quickly understand what it is imo.

PR #1042, opened 2019-04-09
TFW-3001 - Data should be loaded in to WordBars

tVeratti:
WHEW LAD!

Made a lot of changes to `lineGraph` and especially `wordBar` after I got new updated json data samples.


tVeratti:
I've also started a new way of demo'ing the visualizations. I am mimicking what our reducer will eventually look like with dashboards/visualizations lining up by ids. That's what you'll see in demo with merging data. As I continue to connect data to different graphs, their sample data will also get merged in there.

assets/js/mockData/wordbars.js
```diff
@@ -0,0 +1,102 @@
...
+        encoding: {
+            y: {
+                field: 'key',
+                type: 'ordinal'
+            },
+            x: {
+                field: 'doc_count',
+                type: 'quantitative'
+            },
+            sentiment: {
```
troyblank: not sure if this will be something the API will give us just FYI. I kinda got a little creative here originally.

components/visualizations/wordBar.js
```diff
@@ -1,17 +1,45 @@
 import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import memoize from 'memoize-one';
 import { Card, TabNavigation } from '../';
 import VerticalBar from './verticalBar/verticalBar';
 import './wordBar.beta.scss';
 
+export const addPercentsToData = memoize((values) => {
```
troyblank: Couple things here; there is already a util class that is called `addPercentsToData` in the `components/visualizations` folder that can do this now. Also we probably should not be hard coding `encoding.x.field` in this scenario as we can't always trust that axis will always be `doc_count`. Most visualizations do this but word bar chart is old and has yet to be converted.

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -45,7 +45,7 @@ export default class VerticalBar extends PureComponent {
         barWrapper.append('div')
             .style('width', d => `${d[KEY_PERCENT]}%`)
             .attr('class', d => `vertical-bar__bar color-vis-${d.sentiment}`)
-            .attr('data-tip', d => sanitizer.sanitize(`<label>${d.key}</label><div>${d.doc_count} interactions</div>`))
+            .attr('data-tip', d => sanitizer.sanitize(`<label>${d.key || d.label}</label><div>${d.doc_count} interactions</div>`))
```
troyblank: I don't understand the need for `d.key || d.label` we should probably use one or the other right? Will the labels in the data arrays be variable? Also consider possibly making an `encoding` for this in the data possibly?

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -28,15 +28,15 @@ export default class VerticalBar extends PureComponent {
         this.div = d3.select(`#${id}`).append('div')
             .attr('class', 'vertical-bar');
         const bar = this.div.selectAll('div')
-            .data(data)
+            .data(data.values || data)
```
troyblank: Shouldn't we just use one or the other here? I am not seeing the use case where this could be different. If we just need to update the other demos outside vizzynizzy we should just do that.

components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -2,30 +2,51 @@ import React from 'react';
...
+        visProps = { charts: results };
+        break;
+    }
+    case LINE:
+    default: {
+        const { x, y } = results.encoding;
+        const metaData = { xKey: x.field, yKey: y.field };
+        visProps = { data: results.data.values, metaData };
+        break;
+    }
```
troyblank: The LINE case seems very similar to most visualizations this might become a `default` in the future but hopefully there will be less exceptions so this switch doesn't grow too big.

components/visualizations/verticalBar/verticalBar.js
tVeratti: In the raw json that I was given from the api, it had both in the same data set. Not ideal, but it does keep us from having to change it if they don't know what they're doing with it yet.

components/visualizations/verticalBar/verticalBar.js
tVeratti: This was my attempt to make a more passive change since I'm not fully aware of the plan for the difference use cases of vertical bar. If you think it's only the demo, then I will remove it

components/universalSearch/visualizations/vizzyNizzy.js
tVeratti: It is actually - falls through to default. I kept it there purely to help us when reading it, in case we specifically wanted to for LINE and didn't have to work to make the logical leap to "it's just handled by default"

components/visualizations/wordBar.js
tVeratti: Ok I'll shift to use the existing util, this is the one that was coded into the demo and I moved it down to the component (after changing the values expected type from dict to array)

components/visualizations/verticalBar/verticalBar.js
troyblank: ahhhh, i see. Ok sound fun, this is good then :D

components/visualizations/verticalBar/verticalBar.js
troyblank: Yea, I think i made the mistake if passing in data with the full object in tack when it should be just the array. So I assume we can simplify and not have to check for `data.values`. But let me know if I am missing something.

components/universalSearch/visualizations/vizzyNizzy.js
troyblank: OH yea, I need my eyes checked I see that now :) You are always two steps ahead of me.

components/visualizations/wordBar.js
troyblank: Yea that was my fault I should have gone back and swapped that out, so sorry for the confusion, but let me know if it has issues using the util. From looking at it, it seems to be doing exactly what the util is doing.

jasonhale:
So far so good.  I'm good with these changes once Troy's concerns are taken care of.

assets/js/mockData/wordbars.js
tVeratti: This is data that I got straight for T.A. so it seems consistent.

components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -2,30 +2,51 @@ import React from 'react';
 
 import * as Vizzies from '../../visualizations';
 
+
```
troyblank: the new linter updates might not allow this double line...just a heads up, no a prob with me.

components/visualizations/verticalBar/verticalBar.js
troyblank: Sorry noticed one small thing here after your changes `doc_count` can be dynamic based on `yKey` here. Although you didn't really change it so I won't block you with it.

PR #1043, opened 2019-04-11
TFW-3288 - added a "too few" notification for word cloud

troyblank:
This builds on another PR out there now. This might need some eyes on it for standardizing it to the way we do loading or errors possibly. But i did what I think was best considering them, but might need feedback in that regard.

Screen Shot 2019-04-11 at 9 52 16 AM


tVeratti:
This makes the workflow and qlik error message fullscreen. We might need to add a way to distinguish a fullscreen info panel and an auto-sized one.

troyblank:
@tVeratti ok made `style-variation` prop for info panel, thanks for the heads up missed all the use cases.

PR #1044, opened 2019-04-11
TFW-3332 - Visualization Error

tVeratti:


components/icons/alert.js
```diff
@@ -24,7 +23,7 @@ export default function IconAlert({ size = 'large', color = 'black', filled = fa
         { isLarge &&
           <g>
             <path
-              fill={'currentColor'}
+              fill={fill}
```
jasonhale: Why this change here?  Most, if not all of the new icons that have filled versions have been implemented with a `currentColor` fill (or at least defaulting to it).  This will require specifying the color for all instances of this component since it defaults to a black fill.

There's still a couple spots where the IconAlert color isn't passed as a prop.

components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import * as Vizzies from '../../visualizations';
+import './vizzyNizzy.beta.scss';
 
 import './vizzyNizzy.beta.scss';
```
jasonhale: d'oh - imported twice.  I imagine this was just a goof while getting latest.

components/util/colors.js
```diff
@@ -25,6 +25,7 @@ export const subtleYellow = '#fffde0';
 export const red = '#ad1919';
 export const mutedRed = '#752b2b';
 export const subtleRed = '#ffe3e3';
+export const warningRed = '#f71e44';
```
jasonhale: In the design specs, that's noted "red-1", not warning red.  I think it'd be helpful to stick to that naming convention so we all know the reference.

components/icons/alert.js
troyblank: Yea, I thought we planned on changing all icons with css from the outside from here on out.

components/util/colors.js
troyblank: It's called `red` right now as the design docs changed. But yea the color is already there and this line is not needed but I would be up for changing it to match design docs for sure.

Screen Shot 2019-04-12 at 9 56 10 AM

components/icons/alert.js
tVeratti: Good call. For some reason I had it in my head that `colors` default was `currentColor` so it wouldn't have changed the result. However after double checking you're right and it's `black`. Will change to use css.

components/universalSearch/visualizations/vizzyNizzy.js
tVeratti: ah merge conflict fail :(

components/util/colors.js
tVeratti: I'm going to remove it altogether and use css as suggested above.

PR #1046, opened 2019-04-11
TFW-2998 - added loading animation to visualizations

jasonhale:
This adds the loading animation to visualizations.
Also introduced some memoization from @tVeratti 's suggestions.

The WordBar may need some work depending on some other in-progress tickets.


components/visualizations/donut/donut.js
```diff
@@ -25,6 +25,12 @@ export function animateText(text, isHidden) {
 }
 
 export default class Donut extends PureComponent {
+    static baseLayout = {
```
troyblank: I this just to make sure the layout props don't get mutated?

components/visualizations/donutStack.js
```diff
@@ -11,40 +12,64 @@ import './donutStack.beta.scss';
...
-            headerChildren={<AxisControls onXUpdate={this.onXUpdate} onYUpdate={this.onYUpdate} />}
+            headerChildren={this.getAxisControls()}
           >
-            <Legend id={id} labels={channels} colors={themeA} />
-            <div className={'donut-stack__graphs'}>
-              <Donut id={`${id}Donut`} data={donutData} metaData={donutMeta} colors={themeA} />
-              <Stack id={`${id}Stack`} data={filteredStackData} metaData={stackMeta} />
-            </div>
+            { isPending ?
+              <div style={{ height }}>
```
troyblank: I am not sure this approach is best as I have talked about with you prior to this work I think the svgs can make the container the height you want without doing the inline style. You can see where this is a bit flawed when you use the smaller screen view where the svgs become stacked on the Donut + Line and there is a lot of shifting as data is loading. You can avoid this shifting if you just let the svgs get rendered without data much like I did with the "too few data" info panel work. 

The idea is render the svgs even when the component is pending, and only write data to them after the pending is done. Pending then would be placed OVER all content in the vizzy component.

troyblank: Granted horizontal scroll bars do cause shifting in all views but that's sorta unavoidable.

assets/js/demo/components/universalSearch.js
```diff
@@ -8,6 +8,8 @@ import visualizations from '../../mockData/visualizations';
 export default class UniversalSearchDemo extends Component {
     state = { queries: [] };
 
+    onSplashClick = () => this.getResults([]);
```
tVeratti: Nice :)

components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -18,16 +18,21 @@ export const VIZZY_TYPES = {
...
 
     if (!Component) return <div>Vizzy not found for {props.name} ({type})</div>;
 
-    // Prepare MetaData for the vizzy before passing props.
-    const { x, y } = encoding;
-    const metaData = { xKey: x.field, yKey: y.field };
+    if (encoding) {
+        // Prepare MetaData for the vizzy before passing props.
+        const { x, y } = encoding;
+        metaData = { xKey: x.field, yKey: y.field };
```
tVeratti: Assuming you will have conflicts here, lines 29/30 belong in the bottom case of the new switch.

tVeratti:
I was expecting the loading check/display to be handled a bit higher up like in `vizzy-nizzy` but you've got it worked out so I won't complain. It gives flexibility between charts but adds some cost to maintaining them. Seems like a fine tradeoff. Also nice hooks :D

jasonhale:
@tVeratti - the check for pending data is in the `Vizzy-Nizzy` component.  But, due to design, the loader display has to be a little lower in the tree.

components/visualizations/wordCloud/wordCloud.js
```diff
@@ -75,7 +92,7 @@ export default class WordCloud extends PureComponent {
             .size([width, height])
             .words(data)
             .font('Inter UI')
-            .fontSize(d => { return this.setFontSize(d.size); })
+            .fontSize(d => this.setFontSize(d.size))
```
troyblank: oooops I did that, thanks! :D

PR #1050, opened 2019-04-15
hotfix fields type changing to array in encoding

tVeratti:


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -45,7 +45,9 @@ export default function VizzyNizzy({ type, results, ...props }) {
     default: {
         if (results.encoding) {
             const { x, y } = results.encoding;
-            metaData = { xKey: x.field, yKey: y.field };
+            const xKey = x.fields ? x.fields[0] : x.field;
```
jasonhale: This will work, as long as `x.fields` is _always_ an array.  Otherwise we'll be getting substrings.

PR #1053, opened 2019-04-16
Fix viz demos

troyblank:
There is a few detail issues here but for the most part this fixes the demos from blowing up.

Also merged master on this branch so that's why extra commits.... but it's a good thing right? 😓


assets/js/demo/components/visualizations/wordCloud.js
```diff
@@ -4,8 +4,8 @@ import { WordCloudChart } from '../../../../../components';
 import payload from '../../../mockData/wordCloud';
 
 export default function WordCloudDemo() {
-    const [vizdata, setVizData] = useState({ values: null, metaData: null });
-    const [vizdataSmall, setVizDataSmall] = useState({ values: null, metaData: null });
+    const [vizData, setVizData] = useState({ values: null, metaData: null });
+    const [vizDataSmall, setvizDataSmall] = useState({ values: null, metaData: null });
```
jasonhale: fixing the casing and goofed this one.  no big deal.  just noticed it.

PR #1061, opened 2019-04-17
TFW-3355 - Allow visualization groups

jasonhale:
Adds ability to group visualizations in Dashboard


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -40,6 +53,57 @@ export default function VizzyNizzy({ type, results, ...props }) {
         visProps = { charts };
         break;
     }
+
+    case DONUTLINE: {
+        let donutMetaData, lineMetaData;
+
+        const donutResults = results[DONUT];
```
tVeratti: I think we could look through `Object.keys(results)`, since these all seem to have the same setup?

tVeratti: Or make a little function that transforms encodings to this really common metadata setup.

components/visualizations/stack/stack.js
```diff
@@ -20,8 +33,11 @@ export default class Stack extends PureComponent {
         marginBottom: 24,
         marginLeft: 48
     };
+
     pointSpacing = 60;
+
     yAxisWidth = 50;
+
```
troyblank: ummmm @jasonhale .... it's baaaaack

tVeratti: No that was me - it was autoformatted with eslint.

jasonhale: HA!  That was @tVeratti this time.  My linter has autofix turned off.

tVeratti: guilty :D

components/universalSearch/visualizations/vizzyNizzy.js
jasonhale: Didn't completely roll through it with a `Object,keys(results)` since the prop names are specific for the components and allows the isPending check.

PR #1064, opened 2019-04-17
Allow Donut-LineAvg viz to show

jasonhale:


jasonhale:
just renaming the branch.  brb

PR #1068, opened 2019-04-19
TFW-3392 - change calendar display for create

tVeratti:


assets/js/demo/components/datepicker.js
```diff
@@ -3,6 +3,10 @@ import { Datepicker } from '../../../../components/index';
 
 export default function DatepickerDemo() {
     return (
-      <Datepicker month={0} />
+      <React.Fragment>
+        <Datepicker month={0} />
+        <Datepicker backward={true} />
```
troyblank: I think this should be `canSelectFuture` or something similar. As adding ```backward={true}``` makes it seem like it by default can not go backwards in time and you are allowing it to do so. Which is not what this is doing so it confused me.

PR #1070, opened 2019-04-22
TFW-3415 - added decimal precision of 1 to donut line graph

troyblank:
This adds precision of 1 decimal to the line of donut line graph. I didn't pass this as a prop as I am not fully sure if all donut lines will be a 1 or not. It could easily be made into a prop to pass in though but that brings up a lot of questions about if the donut needs configuration etc so I just made it semi-hard coded at this juncture.


components/visualizations/util.js
```diff
@@ -4,6 +4,16 @@ import { isLastDayOfTheMonth } from '../util/date';
 export const KEY_PERCENT = 'percent';
 export const RESIZE_DEBOUNCE_DELAY = 500;
 
+export function updateDataDecimalPrecision(data, yKey, precision) {
+    const percisceData = [];
```
jasonhale: I know it's not affecting code - but "precise" is misspelled.

troyblank: oooops, ok updated.

PR #1071, opened 2019-04-22
TFW-3391 - Bar lengths match yKey percent values

tVeratti:


assets/js/demo/components/visualizations/wordBarCharts.js
```diff
@@ -21,6 +22,8 @@ export default function WordBarChartsDemo() {
         getData();
     });
 
+    console.log(chartData);
```
jasonhale: Left a log in here.  Accident or necessary?

tVeratti:
Updated demo data

PR #1074, opened 2019-04-23
TFW-3429 - Consistent colors in DonutStack graph

jasonhale:
Addresses keeping label colors consistent for the Donut and StackedArea graphs in a DonutStack graph by introducing `colorMap`.  Keys don't always come back in the same order, so the colors need to be mapped and shared between related graphs.

Also removed an extra (and confusing) donut stack demo component.


components/visualizations/donut/donut.js
```diff
@@ -25,12 +26,31 @@ export function animateText(text, isHidden) {
 }
 
 export default class Donut extends PureComponent {
+    static propTypes = {
```
tVeratti: \o/

PR #1076, opened 2019-04-23
TFW-3426 - updated mode labels for visualizations

troyblank:
changed "Volume" to be "Count" ... gave these labels a more proper home and gave them some home cooking.


components/visualizations/axisControls/axisControls.js
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 import { NewSelect } from '../../';
 import { KEY_PERCENT } from '../util';
+import { withMessages } from '../../messages/MessageContext';
```
tVeratti: Omg... yay! I had honestly kinda forgotten about these despite being the person to ask us to use them, LOL. Thanks for the reminder here.

components/visualizations/axisControls/axisControls.js
```diff
@@ -1,9 +1,10 @@
...
 import './axisControls.beta.scss';
 
-export default function AxisControls({ id, onXUpdate, onYUpdate }) {
+function AxisControls({ id, onXUpdate, onYUpdate, messages }) {
```
tVeratti: Since this is a functional component, you could actually use a hook for the messages!! Which would also save the need for the `dive` in the tests, yay!

```
import React, { useContext } from 'react';
/...
const messages = useContext(MessageContext);
```

Just need to import the actual context itself, not the HOC component (from the same file).

troyblank: @tVeratti  I did like it! Thanks a lot I updated this PR.

PR #1077, opened 2019-04-23
TFW-3374 - On first load, set line scroll to end

tVeratti:
I spent a while trying to cover the `else` case with some trickery and timeouts, but decided that trying to get d3 results in this one case when we've ignore the rest, just isn't worth it.


components/visualizations/line/line.js
```diff
@@ -57,6 +57,7 @@ export default class Line extends PureComponent {
                 this.applyXAxisData();
                 this.applyYAxisData();
                 this.buildChart();
+                this.setScrollEnd();
```
jasonhale: In the case there is data when the component mounts, you may consider adding this line in the same list up in `componentDidMount`.

tVeratti: Sure, will add it there too. Will need to reset the start time, but that's no problem.

jasonhale: Nice!  I was going to ask about that next.

PR #1078, opened 2019-04-23
TFW-3421 - Extend Line y-axis for data extending past

jasonhale:
Some data would be higher than the highest point on the y-axis of the Line graph.  This extends the y-axis one more tick higher to keep all data within the defined y-axis ticks.


components/visualizations/line/line.js
```diff
@@ -285,6 +292,8 @@ export default class Line extends PureComponent {
             .attr('height', 20);
     }
 
+    tickCheck
```
tVeratti: What is this?

jasonhale: That's a tickCheck. Y'know, like what you do after a hike through the woods.

or... it could be a start of a function that wasn't needed and didn't get removed. <sigh>

tVeratti: Hah, thought it might be that :D

components/visualizations/line/line.js
```diff
@@ -82,12 +82,8 @@ export default class Line extends PureComponent {
     }
 
     setXScale() {
-        const { data, metaData } = this.props;
```
troyblank: thanks and sorry :D

components/visualizations/line/line.js
```diff
@@ -101,15 +97,26 @@ export default class Line extends PureComponent {
...
             .range([height - marginTop, marginBottom]);
+
+        // Add one more tick if data extends past the last one generated by d3
+        const ticks = this.yScale.ticks();
+        const lastTick = ticks[ticks.length - 1];
+        const newLastTick = lastTick + (ticks[1] - ticks[0]);
+        if (lastTick < domainMax) {
+            ticks.push(newLastTick);
+            this.yScale.domain([domainMin, newLastTick]);
+        }
```
troyblank: awesome! glad this was relatively easier than I thought it would be :D

PR #1081, opened 2019-04-24
TFW-3422 - added auto width rules to points that fit in line graph container

troyblank:
Based on the new-ish design rules, the points of a line and stack graph take up the full graph if they are not larger than the container itself. You can see this by selecting the month view in the line graph.

Screen Shot 2019-04-24 at 9 50 19 AM

Note that I did not change the spacing of the points from 50px to 40px apart per the design as this would mean we also need to change the date formatting which is a much bigger task that has yet to be scoped out.


components/visualizations/line/line.js
```diff
@@ -74,21 +74,26 @@ export default class Line extends PureComponent {
         window.removeEventListener('resize', this.updateGraphScale);
     }
 
-    getContentWidth() {
+    setContentWidth() {
```
jasonhale: @tVeratti 's changes for setting scroll to end of graph used `getContentWidth()`.  This would conflict with work in PR 1077 (https://github.com/topbox-rci/rci-components/pull/1077)

Looks like you need to get latest from develop.

troyblank: poooop, ok!

troyblank: @jasonhale  good call! weirdly there was no git conflicts about it :) ok updated this PR to work with that stuffffff.

components/visualizations/line/line.js
```diff
@@ -27,6 +27,7 @@ export default class Line extends PureComponent {
     componentDidMount() {
         const { isPending, data } = this.props;
 
+        this.setContentWidth();
```
jasonhale: 👍

PR #1082, opened 2019-04-24
TFW-3420 - Fix labels on Line Avg chart

jasonhale:
There is currently no other chart plans to use the Line Avg for anything other than sentiment.  Updated the labels on the line chart to reflect "sentiment" data if a `target` is passed.


tVeratti:
Ideally these could use `messages` context, but that's something we can tackle more broadly next time things are less high-pressure.

jasonhale:
oooh, good point.  Yeah, we should tackle all those viz's at once with label updates to use messages.

PR #1084, opened 2019-04-24
TFW-3427 - Area tooltip should toggle to counts

tVeratti:
Screen Shot 2019-04-24 at 2 49 26 PM


components/visualizations/stack/toolTip.js
```diff
@@ -4,7 +4,9 @@ import { injectNumberCommas } from '../../util/string';
 import { KEY_PERCENT } from '../util';
 import './toolTip.beta.scss';
 
-export default function ToolTip({ date, totalInteractions, interactions }) {
+export default function ToolTip({ date, totalInteractions, interactions, metaData }) {
+    const { yAxisMode } = metaData;
+    const isPrecentage = KEY_PERCENT === yAxisMode;
```
jasonhale: Got a misspelling there.  :P

components/visualizations/stack/toolTip.js
```diff
@@ -4,7 +4,9 @@ import { injectNumberCommas } from '../../util/string';
 import { KEY_PERCENT } from '../util';
 import './toolTip.beta.scss';
 
-export default function ToolTip({ date, totalInteractions, interactions }) {
+export default function ToolTip({ date, totalInteractions, interactions, metaData }) {
+    const { yAxisMode } = metaData;
```
jasonhale: For the sake of simplification: better to pass the single `yAxisMode` as a prop, instead of `metaData` that you have to use only the one value out of to keep the function explicit.

tVeratti:
@jasonhale updated based on your comments

PR #1085, opened 2019-04-24
TFW-3417 - Fix missing labels on Donut Chart

jasonhale:
And some prop type cleanup as bonus.


tVeratti:
Thanks for updating the proptypes as we go :)

PR #1088, opened 2019-04-24
TFW-3418 - Fix missing counts on Donut tooltips

jasonhale:


jasonhale:
Yep!  Already left a comment in that ticket and told Dave.  :)

PR #1090, opened 2019-04-25
TFW-3394 - Don't allow non-matching entries in SearchBar AutoComplete

jasonhale:
In order to optimize ES searches, we're being asked to not submit search queries from Universal Search autocomplete entries unless it matches one of the available dimensions.

This disables the add/omit buttons until an allowed value is entered.


jasonhale:
Yes.  However, doing so will not allow autocomplete to send changes up to the parent to make the call to get matching dimensionValues (it'd be calling the api for matches against an empty string every time - or until it matches one of the initial 10 dimensionValues supplied).

But, yeah, if we got _all_ the dimensionValues up front, `allowAnyValue: false` would work.

troyblank:
@jasonhale  cool! just making sure we are not re-building the wheel. 🎸

PR #1091, opened 2019-04-25
TFW-3287 - added tooltip to wordcloud

troyblank:
This is pretty straightforward besides for the `count` prop added the meta data and encoding. I feel like this will be the best way going forward but feel free to push back if you see a better route. It's easily changeable though if need be down the road.

Screen Shot 2019-04-25 at 2 56 54 PM


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -41,10 +41,13 @@ export function setMetaData(encoding) {
     let metaData;
 
     if (encoding) {
-        const { x, y } = encoding;
+        const { x, y, count } = encoding;
         const xKey = x.fields[0];
         const yKey = y.fields[0];
-        metaData = { xKey, yKey };
+        metaData = Object.assign(
```
jasonhale: Only a suggestion... could be simplified a _tiny_ bit by:
```js
metaData = {
  xKey,
  yKey,
  ...(count && { count: count.fields[0] })
);
```

jasonhale: Only a suggestion, could be simplified a _tiny_ bit by:
```js
metaData = {
  xKey,
  yKey,
  ...(count && { count: count.fields[0] })
);
```

troyblank: @jasonhale  updated, thanks!

troyblank:
@tVeratti  because `KEY_PERCENT` is made on the front end. This count is actually data coming from the api so I feel like it's different and part of the metadata.

troyblank:
@tVeratti  I used `KEY_COUNT` which ended up being a bit different than `KEY_PERCENT` as the count is just in the encoding/meta. But there is great power in having this all in one place. Holler if it looks weird in any place as it felt a tiny bit odd but it's an improvement for sure. thanks!

PR #1096, opened 2019-04-26
Fix assets linting

troyblank:
same thing as last but for assets


assets/js/demo/components/navPaneDemo.js
```diff
@@ -25,7 +25,6 @@ export default function NavPaneDemo() {
         <NavPane
           mainConfig={mainConfig}
           userName={userName}
-          activeClient={activeClient}
```
jasonhale: Oof.  This shouldn't be deleted.  Should be `defaultActiveClient`

troyblank: ah! see linters are our friends and find stuff like this :) fixed thanks!

PR #1097, opened 2019-04-26
Fix linting

troyblank:
last one!


components/visualizations/line/line.js
```diff
@@ -432,7 +432,7 @@ export class LineComponent extends PureComponent {
     }
 
     /* eslint-disable-next-line react/sort-comp */
-    updateGraphScale = debounce(() => {;
```
jasonhale: oh.  damn.

troyblank: yuuuuuuupppp :) so sad our linters were not working for a bit. Barfs.

troyblank:
@jasonhale if you don't mind checking my last commit; I had to push one more as new lint issue got into develop.

PR #1098, opened 2019-04-26
TFW-3446 - Hook DonutGraph to Vizzy

jasonhale:
Allows DonutGraph to display via Dashboard


components/visualizations/donutGraph.js
```diff
@@ -57,7 +57,7 @@ export default class DonutGraph extends PureComponent {
             <Legend id={id} labels={labels} colors={themeA} />
             <div className={'donut-card'}>
               <Donut
-                id={`${id}Donut`}
+                id={`Donut${id}`}
```
troyblank: Ah thanks, I thought I got this one but apparently not!

jasonhale: Oh, thanks for the reminder.  There were a few others I updated after you mentioned this.

PR #1101, opened 2019-04-30
TFW-3468 - Allow strings as x-axis labels in LineGraph

jasonhale:
This introduces the _type_ of encoding for the x-axis into LineGraph so it can determine how to render the labels - as "strings" or dates if type is "temporal".

(Bonus: fixed missing test coverage on Datepicker.)


components/visualizations/lineGraph.js
```diff
@@ -15,23 +15,29 @@ export default class LineGraph extends PureComponent {
...
-            xKey: PropTypes.string.isRequired
+            xKey: PropTypes.string.isRequired,
+            xType: PropTypes.string.isRequired
         })
     };
 
     state = {
         xAxisMode: 'day',
-        yAxisMode: 'count'
+        yAxisMode: KEY_COUNT
```
troyblank: thanks! i think I missed this.

components/visualizations/lineGraph.js
```diff
@@ -15,23 +15,29 @@ export default class LineGraph extends PureComponent {
...
     }
 
     /* eslint-disable-next-line react/sort-comp */
     updateData = (data, xAxisMode, xKey, yKey) => {
-        const byDate = filterDataByDate(xAxisMode, data, xKey, yKey);
-        return addPercentsToData(byDate, yKey);
+        let filteredData = data;
+        if (this.canUpdateXAxis()) {
+            filteredData = filterDataByDate(xAxisMode, data, xKey, yKey);
+        }
```
troyblank: Strings can not be updated... this seemed weird to be at first as it's not really named as "x axis is a date" ... "then do this" but rather it can be updated because it's 'temporal'. I think this is fine just leaving this note as it made me have to think a bit and possibly could have better naming but not sure what that would be fully.

jasonhale: I was thinking more of extending if there would be another reason to update the x-axis.  Y'think this is a little too far-future coding? Renaming to be more date oriented could keep confusion down.

troyblank: Yea that was my gut to make it more date oriented as it caused my pause. But it's a very small concern.

assets/js/demo/components/visualizations/lineGraph.js
```diff
@@ -16,18 +17,31 @@ export default function LineGraphDemo() {
...
+      <React.Fragment>
+        <LineGraph
+          id={'lineChartDemo'}
+          label={'Total Contacts'}
+          data={vizdata.values}
+          metaData={vizdata.metaData}
+          isPending={isPending}
+        />
+        <LineGraph
+          id={'lineChartDemoPartDeux'}
```
tVeratti: :D PartDeux

components/visualizations/lineGraph.js
```diff
@@ -15,23 +15,29 @@ export default class LineGraph extends PureComponent {
...
         metaData: PropTypes.shape({
             yKey: PropTypes.string.isRequired,
-            xKey: PropTypes.string.isRequired
+            xKey: PropTypes.string.isRequired,
+            xType: PropTypes.string.isRequired
         })
     };
 
     state = {
         xAxisMode: 'day',
```
tVeratti: Should `day` also be a key constant?

troyblank: I agreeeee with this

jasonhale:
Updated to add "day", "week", "month" as constants.  Named `FILTER_DAY`, `FILTER_WEEK`, `FILTER_MONTH`... since that was their use.  Open to suggestions if this doesn't make sense.

PR #1102, opened 2019-04-30
TFW-3447 - Allow visualization legends to wrap

jasonhale:
Normal:
Screen Shot 2019-04-30 at 1 52 57 PM

Wrapping:
Screen Shot 2019-04-30 at 1 52 39 PM


components/visualizations/legend/legend.beta.scss
```diff
@@ -2,19 +2,17 @@
 
 .graph-legend {
     display: flex;
+    flex-flow: row wrap;
     justify-content: center;
     margin: (-$space-lg) (-$space-md) $space-md;
-    padding: $space-sm 0;
+    padding: $space-xs ($space-md / 2);
```
tVeratti: This 12px spacing seems to be coming up more and more. Is this something we should ask to be addressed by our overall style guide? It's either an outlier that doesn't match the 8px rule, or we need to add 12px as a common spacing?

jasonhale: Not really sure it should be added for the simple fact that the times the 12px is needed is when it's only half of the actual visual spacing of 24px.

Like, the style-guide directed end product is 24px -  but due to how things are built it has to be implemented in pieces.

But, you're right.  It has come up a lot.  I'm not totally opposed to adding.  (Naming will be awkward ;) )

troyblank: Yea i saw this implementation vs concept sorta thing. To me it seems normal but maybe a var with a name that is indicative that it's half would be convenient.

PR #1103, opened 2019-04-30
TFW-3451 - added base combo bar line chart without z axis

troyblank:
This adds anything for combo bar line chart not included as feature enhancements. Which is basically a bar chart using the line chart demo data :) 

Screen Shot 2019-04-30 at 3 34 59 PM

There is a small coverage issue here that is fixed with the PR jason is doing > https://github.com/topbox-rci/rci-components/pull/1101/commits/a74fc16e5c7d80ce072a2294ebe16a2fd3e96750

so please ignore that.


components/visualizations/comboBarLine/comboBarLine.js
```diff
@@ -0,0 +1,260 @@
...
+        const { isPending } = this.props;
+
+        this.setContentWidth();
+        this.buildXAxis();
+        this.buildYAxis();
+        this.buildSVG();
+
+        if (!isPending) {
+            // apply data to graph if data should be available
+            // this.maxDataLength = data.length;
```
jasonhale: Commented code.  Needed?

components/visualizations/comboBarLineChart.js
```diff
@@ -0,0 +1,72 @@
...
+            PropTypes.object
+        ),
+        metaData: PropTypes.shape({
+            yKey: PropTypes.string,
+            xKey: PropTypes.string
+        })
+    };
+
+    state = {
+        xAxisMode: 'day'
```
jasonhale: we have a constant for that now.  :)

components/visualizations/comboBarLineChart.js
```diff
@@ -0,0 +1,72 @@
...
+        return null;
+    }
+
+    render() {
+        const { label, id, metaData, isPending = false, data } = this.props;
+        const { xAxisMode } = this.state;
+        let filteredData, meta;
+
+        if (!isPending) {
+            filteredData = data;
```
jasonhale: This line doesn't look necessary.

troyblank: nope it's not, removed!

components/visualizations/comboBarLineChart.js
troyblank: haha, no fair no fair, updated!

components/visualizations/comboBarLine/comboBarLine.js
troyblank: noooope, removed thanks.

PR #1106, opened 2019-05-02
TFW-3454 - added z axis to combo chart

troyblank:
This just adds the axis for the Z side. Data plotting will come in another PR.

Screen Shot 2019-05-01 at 11 16 47 PM


components/visualizations/comboBarLine/comboBarLine.js
```diff
@@ -92,32 +101,50 @@ export class ComboBarLineComponent extends PureComponent {
...
         conformTicksToCoverLargestDataPoint(this.yScale, domainMin, domainMax);
     }
 
-    buildSVG() {
-        const { marginTop } = this.layout;
+    setZScale() {
+        const { data, metaData } = this.props;
+        const { height, marginTop } = this.layout;
+        const { zKeys } = metaData;
+        const domainMin = 0;
```
jasonhale: The design shows a non-zero domain minimum on the z axis.  Should this be set here, or is that a future-state thing?

troyblank: ah yea... good call, i need to calculate minimum as well. Let me do that. Most of them are zero in the design so this was just an oversight on my part.

troyblank: ok updated to accommodate this now thanks!

assets/js/demo/components/visualizations/comboBarLineChart.js
```diff
@@ -15,8 +15,8 @@ export default function ComboBarLineChartDemo() {
 
     useEffect(() => {
         const { values, encoding } = payload;
-        const { x, y } = encoding;
-        const metaData = { xKey: x.fields[0], yKey: y.fields[0] };
+        const { x, y, z } = encoding;
+        const metaData = { xKey: x.fields[0], yKey: y.fields[0], zKeys: z.fields };
```
tVeratti: 👍

components/visualizations/comboBarLine/comboBarLine.js
```diff
@@ -23,18 +25,25 @@ export class ComboBarLineComponent extends PureComponent {
     pointSpacing = 50;
     barWidth = 16;
     yAxisWidth = 50;
+    zAxisWidth = 50;
+    axisColorDotPadding = 8;
+    axisTextPadding = 3;
+    yColor = 'blue-5'
```
tVeratti: I definitely feel like these magic strings should be constants, but we can make a tech story to tackle that since I'm sure this is an existing thing elsewhere too.

PR #1112, opened 2019-05-02
TFW-3452 - Visualizations should be default on search if available

tVeratti:


components/universalSearch/universalSearch.js
```diff
@@ -108,8 +116,8 @@ class UniversalSearch extends PureComponent {
     getPaths = () => {
         const { basePath } = this.props;
         return {
-            results: `${basePath}`,
-            visualizations: `${basePath}/visualizations`
+            results: `${basePath}/results`,
```
jasonhale: I would think this should be "/interactions" - to match the label on the button that switches views.  "results" seems overly broad.  Is "results" the preferred routing?

troyblank: I agree with this, since the user will see both the button label and the path I also think they should be consistent.

PR #1113, opened 2019-05-03
TFW-3453 - added plotted lines to combo chart

troyblank:
This adds plotted lines on z index on combo.

Screen Shot 2019-05-03 at 10 32 57 AM


components/visualizations/colors.beta.scss
```diff
@@ -27,6 +31,16 @@
 .color-vis-blue-8 { @include chart-color($color-vis-blue-8); }
 .color-vis-blue-9 { @include chart-color($color-vis-blue-9); }
 
+.stroke-vis-blue-1 { @include stroke-color($color-vis-blue-1); }
```
jasonhale: Would it help or hinder using Sass's `@for` loop here?
```@for $i from 1 through 9 {
  .stroke-vis-blue-#{$i} {
    @include stroke-color($color-vis-blue-#{$i});
  }
```

Just wondering if that'd be helpful, or even beneficial, now that there's 2 instances of sequential colors.

assets/js/mockData/comboBarLine.js
```diff
@@ -15,8 +15,8 @@ const comboBarLineData = {
         },
         z: {
             fields: [
-                'silentTime',
-                'duration'
+                'duration',
```
tVeratti: Is the order important?

components/visualizations/colors.beta.scss
tVeratti: Nice idea, we could really get sass to do some of this work for us :)

components/visualizations/comboBarLine/comboBarLine.beta.scss
```diff
@@ -34,3 +34,8 @@
 .combo-bar-line .tick text {
     transform: translate(3px, 0);
 }
+
+.combo-bar-line__line {
```
tVeratti: combobarline line? :P

components/visualizations/colors.beta.scss
troyblank: This is a great idea; but it appears making dynamic vars on run time is not supported by sass from my toying around. I could be wrong but it doesn't seem to be possible.

assets/js/mockData/comboBarLine.js
troyblank: It is sorta, it matches up the colors in which in design silent time is less important (lighter blue)... we could make a color map that loosely connects the props though but order seemed more dynamic and iron glad to me. So yea it matters in which one gets the light or dark blue but it does not mater a lot. I talked to API peeps to make sure this order is like this.

components/visualizations/comboBarLine/comboBarLine.beta.scss
troyblank: david?

PR #1115, opened 2019-05-03
TFW-3549 - Update to Wordcloud tooltip

jasonhale:
Removed the "% of interactions" and interactions count, updated the label for score, and fixed the score decimal.


components/visualizations/wordCloud/wordCloud.js
```diff
@@ -66,11 +64,9 @@ export class WordCloudComponent extends PureComponent {
     buildToolTip(d) {
         const { metaData, messages } = this.props;
         const { xKey, yKey } = metaData;
-        const percentText = `<div>${d[KEY_PERCENT]}% of ${messages.DESCRIPTION_INTERACTIONS}</div>`;
-        const scoreText = `<div>${d[yKey]} ${messages.DESCRIPTION_SIGNIFICANT_TERM_SCORE}</div>`;
-        const countText = `<div>${injectNumberCommas(d[metaData[KEY_COUNT]])} ${messages.DESCRIPTION_INTERACTIONS}</div>`;
+        const scoreText = `<div>${messages.DESCRIPTION_SIGNIFICANT_TERM_SCORE} ${Number.parseFloat(d[yKey].toFixed(4))}</div>`;
```
tVeratti: Should this use the `precision` prop from vizzy nizzy? It would be the first one set up with a different value (`4`) from others, but would serve as a good first example for overwriting it to that value.

jasonhale: True.  This is less data, more tooltip... which was an exception i was fighting over internally when adding that precision prop.

I can add and see what we think.

jasonhale: @tVeratti - added.

components/visualizations/wordCloud/wordCloud.js
```diff
@@ -64,7 +64,8 @@ export class WordCloudComponent extends PureComponent {
     buildToolTip(d) {
         const { metaData, messages } = this.props;
         const { xKey, yKey } = metaData;
-        const scoreText = `<div>${messages.DESCRIPTION_SIGNIFICANT_TERM_SCORE} ${Number.parseFloat(d[yKey].toFixed(4))}</div>`;
+        // const scoreText = `<div>${messages.DESCRIPTION_SIGNIFICANT_TERM_SCORE} ${Number.parseFloat(d[yKey].toFixed(4))}</div>`;
```
troyblank: Are we going to add this again soon? Why just commented out?

jasonhale: Nah, just neglected to remove it while testing things out.  Thanks for catching.

components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -132,15 +132,24 @@ export default function VizzyNizzy({ type, results, ...props }) {
...
         isPending = Boolean(!metaData && !data);
         visProps = { data, metaData, isPending, decimalPrecision };
         break;
     }
+
+    case WORDCLOUD: {
+        data = results.values;
+        metaData = setMetaData(results.encoding);
+        isPending = Boolean(!metaData && !data);
+        visProps = { data, metaData, isPending, decimalPrecision: 4 };
```
tVeratti: Cool - I just think it fits well here so we can see it clearly has a different value than the rest of the vizzies. Thanks for updating!

PR #1125, opened 2019-05-08
add debug check for vizzy error boundaries

tVeratti:


components/universalSearch/visualizations/error.js
```diff
@@ -11,6 +11,14 @@ export default class Error extends PureComponent {
 
     state = { hasError: false };
 
+    componentDidCatch(error, info) {
```
troyblank: should there be a unit test for this with a console stub?

tVeratti: I don't think so, it's not typical app/business functionality, just a debug helper. That might be a bit like adding a test for the sake of adding a test :)

troyblank: My thought was that this not working correctly potentially effect other business functionality if it errors out (weirder things have happened). But not a huge deal cool with this as is.

PR #1128, opened 2019-05-09
TFW-3583 - allowed visualization date filtering to have flexible average keys

troyblank:
Treating this sorta like a hotfix (going to master).

Issue on test; we need the average of the zKeys but not the yKeys, so I made the average boolean on the `filterDataIntoDateGroups` an array with what keys should be averaged instead of just a simple boolean that does all keys.


components/visualizations/util.js
```diff
@@ -93,10 +94,10 @@ export function filterDataIntoDateGroups(filter, data, encodings, averageCount)
     return dataInGroups;
 }
 
-export function filterDataByDate(filter, data, encodings, averageCount = false) {
+export function filterDataByDate(filter, data, encodings, averagedEncodings = false) {
     if (FILTER_WEEK !== filter && FILTER_MONTH !== filter) return data;
 
-    return exports.filterDataIntoDateGroups(filter, data, encodings, averageCount);
+    return exports.filterDataIntoDateGroups(filter, data, encodings, averagedEncodings);
```
tVeratti: I don't think `exports.` is necessary but it's nbd to keep it :)

components/visualizations/util.js
```diff
@@ -76,11 +76,12 @@ export function filterDataIntoDateGroups(filter, data, encodings, averageCount)
         }
 
         if (endOfGrouping) {
-            if (averageCount) {
+            if (averagedEncodings) {
                 const averageOut = (k) => { totals[k] = Number((totals[k] / counter).toFixed(1)); };
```
tVeratti: I have to wonder if there isn't anything we could use from lodash that might provide some yummy optimization for loops like these? Not a blocking issue, just a thought.

PR #1129, opened 2019-05-09
TFW-3513 - Create scrollEnd exceptions for Contacts Received vizzies

tVeratti:
I wanted these to default to scroll to the end, which is why things are named `scrollStart` to allow omission to behave with the negative default.


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -142,7 +144,7 @@ export default function VizzyNizzy({ type, results, ...props }) {
         data = results.values;
         metaData = setMetaData(results.encoding);
         isPending = Boolean(!metaData && !data);
-        visProps = { data, metaData, isPending, decimalPrecision };
+        visProps = { data, metaData, isPending, decimalPrecision, scrollStart };
```
troyblank: `scrollStart` seems strange to me, would `scrollToEnd` be more on the nose? If I am trying to figure out why something is scrolling to the end of a vizzy and i see that prop I would be like oh that's what I need to change. I am not sure I would fully get it with `scrollStart`...what you think?

troyblank: or ummm `startScrolledToEnd` i dunno i hate words... maybe i am over thinking this

tVeratti: The utils are called `setScrollEnd` and while it would be more direct to have the flag be called `scrollEnd` I wanted to make sure the default behavior was to scroll to the end, so if the prop was left off (as it is more all but the 2 exceptions), it would scroll to the end.

tVeratti: Make sense? `scrollStart` being the opposite means it needs to be forced to true to override the default behavior.

troyblank: ah yea i gotcha, still seems a bit weird as it makes me think a scroll is happening at the start of something (maybe the mount)....what about `defaultScrollPosition` ?

tVeratti: `scrollLeft` ?

troyblank: and i like more descriptive props! :P `scrollSetLeft` ? otherwise i think `scrollLeft` is better.

troyblank: er `setScrollLeft` sounds less cray haha

PR #1131, opened 2019-05-10
TFW-3519 - added base bar graph

troyblank:
This covers the basics of adding a bar graph, the y axis control does not work yet and the binning is a little off but there is stories to cover that stuff.

Screen Shot 2019-05-10 at 11 19 26 AM


assets/js/mockData/barGraph.js
```diff
@@ -0,0 +1,49 @@
+const comboBarLineData = {
```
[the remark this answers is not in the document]
troyblank: oops!
will fix later

components/visualizations/barGraph.js
```diff
@@ -0,0 +1,71 @@
...
+        let filteredData, meta;
+
+        if (!isPending) {
+            const { yKey } = metaData;
+            filteredData = this.filterData(data, yKey);
+            meta = { ...metaData, yAxisMode };
+        }
+
+        return (
+          <Card
```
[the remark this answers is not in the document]
troyblank: psssh no fair VizzyCard was not available in stores when I did this code. But will update now that it is.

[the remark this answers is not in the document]
jasonhale: Should've pre-ordered!

troyblank:
@jasonhale I updated this PR based on your comments, thanks!

components/visualizations/barGraph.js
```diff
@@ -0,0 +1,67 @@
...
+            xKey: PropTypes.string
+        })
+    };
+
+    state = {
+        yAxisMode: KEY_COUNT
+    }
+
+    /* eslint-disable-next-line react/sort-comp */
+    updateData = (data, yKey) => addPercentsToData(data, yKey);
```
tVeratti: These two functions don't really need to be separate.

`filterData = memoizeOne((data, yKey) => addPercentsToData(data, yKey))`

components/visualizations/barGraph.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    getAxisControls = () => {
+        const { isPending } = this.props;
+
+        if (!isPending) {
+            return <AxisControls onYUpdate={this.onYUpdate} />;
+        }
+
+        return null;
```
tVeratti: Wouldn't allowing this to return `undefined` be just as good?

troyblank: I have always been told/learned `undefined` is for the computer to set and `null` is for humans to set. So that is why I use `null` in situations like this.

components/visualizations/barGraph.js
troyblank: I adjusted this and am going to merge. thanks for the insight.

PR #1132, opened 2019-05-10
TFW-3531 - hook up avg donut data

tVeratti:
I re-pulled the latest `_search` results from testing, for both CX and OE, and combined them into one `visualizations` array for the mock data.


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -170,8 +173,15 @@ export default function VizzyNizzy({ type, results, ...props }) {
         break;
     }
 
+    case DONUT_AVG:
```
jasonhale: This one could just fall through, couldn't it?  I don't see anything different from the two below `DONUT` and `MULTI_HISTOGRAM`.

tVeratti: I goofed and removed it when I rebased. I've got it added back in the next PR - will link.

PR #1133, opened 2019-05-10
TFW-3537 - allowed for single level word bar charts

troyblank:
This allows our `<WordBar />` component to take a single set of data as well as turn off the tab navigation with the `hasTabNavigation` prop. The reason this is a prop and not dynamically figured out with the data is that during the pending state without any data we want to hide the space that the tab nav will be rendered. Otherwise stuff shifts more than desired.

Screen Shot 2019-05-10 at 3 28 58 PM


jasonhale:
Looks good.  Only issue I see is when on a wide graph, hovering over a long bar, it scales enough to cover the label on the left and extend enough to the right to trigger scrollbars.  Was less noticeable before since these graphs weren't so wide.
Can we reduce the x scaling to reduce that occurrence?

Screen Shot 2019-05-13 at 8 53 39 AM

troyblank:
@jasonhale I did notice that and It though I would tackle it when I got to the hover story.... but after looking right now apparently there is none since this is should "just work" so I will fix it here. Thanks!

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -50,15 +68,15 @@ export default class VerticalBar extends PureComponent {
...
             .classed('vertical-bar__label', true);
 
         // Bar
         barWrapper = bar.append('div')
             .classed('vertical-bar__bar-wrapper', true);
         barWrapper.append('div')
             .style('width', d => `${d[KEY_PERCENT]}%`)
-            .attr('class', d => `vertical-bar__bar color-vis-${d.sentiment}`)
+            .attr('class', (d, i) => this.getColorClassForBar(d, i))
             .attr('data-tip', d => sanitizer.sanitize(`<label>${d.key || d.label}</label><div>${d[yKey]} interactions</div>`))
```
tVeratti: The tooltip label currently shows as `undefined` - should these be updated to d[xKey] ?

tVeratti: Just saw that you fixed this on the PR for TFW-3534 - don't need to change it here.

troyblank:
@jasonhale  and @tVeratti  address your comments in my new commit, let me know if you see anything else. Thanks!

components/visualizations/verticalBar/verticalBar.js
troyblank: I updated it anyway here.

PR #1135, opened 2019-05-11
 TFW-3534 - added visible values to vertical bar vizzies

troyblank:


components/visualizations/verticalBar/verticalBar.js
```diff
@@ -89,3 +132,13 @@ export default class VerticalBar extends PureComponent {
         );
     }
 }
+
+// collision detection
```
jasonhale: Left your notes in here. :)

troyblank: You must not tell anyone about these secret notes! 🕵  Removed, thanks.

components/visualizations/verticalBar/verticalBar.beta.scss
```diff
@@ -25,6 +25,22 @@
     width: 100%;
 }
 
+.vertical-bar-text {
```
tVeratti: Technically I think this might be `vertical-bar__text` according to BEM, since `text` is the element.

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -50,19 +79,51 @@ export default class VerticalBar extends PureComponent {
...
 
         // Bar
         barWrapper = bar.append('div')
             .classed('vertical-bar__bar-wrapper', true);
         barWrapper.append('div')
             .style('width', d => `${d[KEY_PERCENT]}%`)
-            .attr('class', d => `vertical-bar__bar color-vis-${d.sentiment}`)
-            .attr('data-tip', d => sanitizer.sanitize(`<label>${d.key || d.label}</label><div>${d[yKey]} interactions</div>`))
+            .attr('class', (d, i) => this.getColorClassForBar(d, i))
+            .attr('data-tip', d => sanitizer.sanitize(`<label>${d[yKey] || d.label}</label><div>${d[yKey]} interactions</div>`))
```
tVeratti: Actually is this the proper label?
Screen Shot 2019-05-13 at 11 28 40 AM

components/visualizations/verticalBar/verticalBar.beta.scss
troyblank: BEM fail, I fixed thanks.

components/visualizations/verticalBar/verticalBar.js
troyblank: oopies yea i did the wrong key... fixed!

PR #1136, opened 2019-05-13
TFW-3538 - Ability to group visualizations into rows

jasonhale:
This introduces the ability to place visualizations into a row so to keep certain visualizations together (and assumes visualizations will always be returned in display order).

This will keep the line graphs for "Contacts Received: Day of Week" and "Contacts Received: Hour of Day" in the same row so other graphs don't share lines with either of these in the case of missing graphs or responsive layout.


components/universalSearch/visualizations/dashboards.js
```diff
@@ -2,12 +2,27 @@ import React, { useContext } from 'react';
...
-import VizzyNizzy, { WORDBAR, VIZZY_GROUPS } from './vizzyNizzy';
+import VizzyNizzy, { WORDBAR, VIZZY_GROUPS, LINEDAYOFWEEK, LINEHOUROFDAY } from './vizzyNizzy';
 import Error from './error';
 import './dashboards.beta.scss';
 
 const isTfe = (id, tfeIds) => tfeIds.indexOf(`${id}`) !== -1;
 
+// Force the lineDayOfWeek and lineHourOfDay graphs into a row so they always occupy the same line
+// ... can be removed when the CMS returns layout.
+export function forceContactsReceivedToRows(dashboardVisualizations) {
```
troyblank: This method seems way too specific to me. Because there is going to be many situations where two things need to occupy the same row I assumed we would have some sort of single method with  a lot of specific logic. Do you think this would be better if it was a function that matches pairs to a row that are in a static object or something like that?

jasonhale: that function was to specifically line up the Contacts Received items, and made to be easily removed later - because, as it stands, those two are the only visualizations so far that aren't supposed to share a row with any other graph (all else are ok with flowing for responsiveness, the min-sizing works so far in all other cases).

If re-use is the concern, we could instead pass the graphs to put in a row as an argument, like `forceToRows(types, dashboardVisualizations)` called by:
```
modifiedDashboard.visualizations = exports.forceToRows([LINEDAYOFWEEK, LINEHOUROFDAY], modifiedDashboard.visualizations);
```

troyblank: @jasonhale cool! that the explanation makes total sense what you did seems logical to me for now.

jasonhale:
Rebased and updated.  Thanks for the reminder, @tVeratti .

PR #1137, opened 2019-05-13
TFW-3582 - Fix Stack graph horizontal scrollbar

jasonhale:
On an initial data load of data that doesn't fill the width (like a single data point), the horizontal scrollbar would still be visible.  Changing the X filter would fix the issue.

Just adding a swift kick to reset the x/y scale after mount and update.


components/visualizations/stack/stack.js
```diff
@@ -64,6 +64,8 @@ export default class Stack extends PureComponent {
             this.maxDataLength = data.length;
             this.setXScale();
             this.setYScale();
+            this.buildXAxis();
```
tVeratti: I don't completely understand this change and how it fixes it - these two get called on lines 59 and 60, so they're doubled up?

jasonhale: Yeah, they do.  However, if the component is no longer pending, the x and y scales get re-calc'd, resetting the content width.  So there's a need to rebuild the axis with the new calculated width.

jasonhale: After investigating, you're correct.  It shouldn't have to get called twice.  Just have to build the y-axis before getting available content width.

Thanks for being persistent!

PR #1139, opened 2019-05-13
TFW-3520 - added sentiment binning to word bar charts

troyblank:
This is currently just for @tVeratti so she can see what I am doing this far.


troyblank:
OK this is good to review now, just adds sentiment binning to word bars.

Screen Shot 2019-05-10 at 3 28 58 PM

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -69,18 +69,33 @@ export default class VerticalBar extends PureComponent {
...
+    }
 
-            color = colorScale[Math.floor(bin)];
-        }
+    hasSentimentBinning() {
+        const { data } = this.props;
+        let hasSentimentBinning = false;
 
-        return `vertical-bar__bar color-vis-${color}`;
+        Object.keys(data[0]).forEach((key) => {
```
tVeratti: This should be
```
const hasSentimentBinning = Object.keys(data[0]).some( ..... )
```
And let the return of the overall some set it instead :)

troyblank: damn forgot about that trick, updated.

PR #1143, opened 2019-05-13
TFW-3520 - added text wrapping for bar graphs

troyblank:
Got a little dirty with this as the only bar graph we have wraps text in this way. In the future we will have to dynamically set the bottom margin and width the text has to wrap. But for now this is a quick and dirty way to make the graph look the way we want.

Screen Shot 2019-05-13 at 5 14 12 PM


components/visualizations/util.js
```diff
@@ -227,4 +228,38 @@ export function addShapeDropShadowDefinitions(container, id) {
...
+            .attr('y', y)
+            .attr('dy', `${dy}em`);
+
+        words.forEach((word, i) => {
+            line.push(word);
+            tspan.text(line.join(' '));
+
+            if (width < tspan.node().getComputedTextLength()) {
+                line.pop();
+                tspan.text(line.join(' '));
```
tVeratti: Took me minute to re-reading this to fully understand how it was intended to handle wrapping. Might be cool to add a comment here about starting a new line at this point, just to make it more obvious up front :)

troyblank: Sounds good added a comment.

PR #1144, opened 2019-05-14
TFW-3512 - Add center text to Donut chart

jasonhale:
Adds text label and value to the center of the Donut chart.

The label and unit for the average may have to be dynamic in the future.  Set now until the source for those are figured out unless anyone feels it should be passed into the DonutGraph within the VizNiz.


components/visualizations/donut/donut.beta.scss
```diff
@@ -13,12 +13,22 @@
 .donut-chart__text {
     transition: opacity $animation-time-subtle $animation-curve-subtle;
     font-size: 1.2rem;
-    font-weight: 600;
```
troyblank: oops, that was me, thanks!

components/visualizations/donut/donut.js
```diff
@@ -99,18 +100,21 @@ export default class Donut extends PureComponent {
         this.segments = this.donut.append('g');
         this.texts = this.donut.append('g')
             .attr('transform', 'translate(0 5)');
+        this.centerText = this.segments.append('g');
```
troyblank: Shouldn't this be a conditional append?

components/visualizations/donut/donut.js
```diff
@@ -149,6 +153,43 @@ export default class Donut extends PureComponent {
...
+
+        this.centerTextValue = this.centerText
+            .append('text')
+            .attr('dy', `${(valueTextSize * lineHeightScale) / 2}rem`)
+            .attr('class', 'donut-chart__avg-value')
+            .attr('text-anchor', 'middle')
+            .text(`${average} s`);
+
+        // if the text is too wide, scale down so it doesn't touch the donut
+        /* istanbul ignore next */
```
troyblank: isn't the whole thing istanbul ignored? no reason to ignore individuals right?

components/visualizations/donut/donut.js
```diff
@@ -149,6 +153,43 @@ export default class Donut extends PureComponent {
...
+        const textValueWidth = (this.centerTextValue.node()) ? this.centerTextValue.node().getBBox().width : 0; // for tests only
```
troyblank: This comment is confusing, it protects more than tests. Browser bugs or rogue dom manipulation. I would plead to remove.

components/visualizations/donut/donut.js
```diff
@@ -149,6 +153,43 @@ export default class Donut extends PureComponent {
...
+        const scale = (maxWidth * 0.8) / textValueWidth;
+        /* istanbul ignore next */
```
troyblank: same here as above.

components/visualizations/donut/donut.js
jasonhale: It wasn't ignoring it.  I'll run through the tests again to double-check.

components/visualizations/donut/donut.js
jasonhale: Was using it more as a placeholder.  The contents are in a conditional below.  If you feel this is extraneous, i can modify.

components/visualizations/donut/donut.js
jasonhale: No prob.  We aren't checking for `node` in the other graphs - and honestly, if it gets this far, there's no reason i can think of that `node` wouldn't be there - in which case error boundaries take over.  This was because the graph would error in tests.  But, yeah, I can definitely take out the comment if it's confusing. 🤘

jasonhale: OH!  and just modified to use `getComputedTextLength()` instead of `getBBox()`.

components/visualizations/donut/donut.js
troyblank: Yea I see that, but this group should not be rendered if nothing is put it in. I would argue this needs to be conditional as well. We might want to add it to the content itself since it's not modified during an update.

troyblank: I should also explain the reason something like `this.segments` exists is because it needs to be re-rendered into during a change of x or y axis. I am not sure your center text needs that so might be overkill to keep a container as an instance variable.

jasonhale: Going through - I feel this needs to stay here.  When new data comes in and `buildChart()` is run again, the placeholder needs to be there so it can be referenced to be removed and a new average to be put in it's place.

Let me know if you see a problem with that logic.

components/visualizations/donut/donut.js
jasonhale: Nope.  Each is required.

components/visualizations/donut/donut.js
troyblank: That's cool, but can we make it conditional? not exist if there is no need for it?

components/visualizations/donut/donut.js
troyblank: That's super bizarre. I might look into this in a bit istanbul should not be doing that.

jasonhale:
@troyblank - pr updated.
Notes:
- the design calls for a consistent with of 40px thickness on the sections.  I've added it here (last commit, can remove if needed), but didn't add rescaling on resize.  Feeling that's out of the scope of this ticket.
- Added to the components and Search Visualizations demo pages.
- other comments in line above.

components/visualizations/donut/donut.js
troyblank: These are not required; it appears the ignore is in the wrong place on the function should be : ```    addCenterText (maxWidth) /* istanbul ignore next: d3 chaining will not render to jsdom in react */ {``` If you set it like that you don't need any more ignores inside. Hopefully that helps.

components/visualizations/donut/donut.js
troyblank: such as ```if (showAverageText) this.centerText = this.segments.append('g');```

troyblank:
@jasonhale it appears ALL donuts are getting the skinny treatment when it should just be the ones with centered text.

Screen Shot 2019-05-14 at 12 50 35 PM

components/visualizations/donut/donut.js
jasonhale: I found where my mental block was at.   Updated.

components/visualizations/donut/donut.js
jasonhale: Nice.  Yup, that did it.

jasonhale:
Removed the segment thickness changes to the donut, and addressing them in a later ticket (created TPB-1962 for follow-up work)

PR #1145, opened 2019-05-14
 TFW-3509 - added tooltips to bar graphs

troyblank:
This might need some work after the toggle is working. But for now I made them work.

Screen Shot 2019-05-14 at 12 37 48 PM


components/visualizations/bar/bar.js
```diff
@@ -130,9 +191,15 @@ export default class Bar extends PureComponent {
             .attr('x', (d, i) => this.xScale(i) - (barWidth / 2))
             .attr('width', barWidth)
             .attr('y', d => this.yScale(d[yKey]))
-            .attr('height', d => (height + marginTop) - this.yScale(d[yKey]))
+            .attr('height', d => this.getBarHeight(d))
             .attr('rx', 2)
-            .attr('ry', 2);
+            .attr('ry', 2)
+            .attr('data-tip', d => this.buildToolTip(d))
```
tVeratti: I think you could simplify these things with 1:1 params as `.attr('data-tip', this.buildToolTip)`

troyblank: @tVeratti  I need `this` to not be in context of d3 selection, so I don't think this is possible. I just tried it and it fails as `metaData` does not exist in the selection.

tVeratti: That's because `buildTooltip` should be self-bound anyway, just to prevent future cases like this where it might called out of the lifecycle :)

tVeratti: SAme with `this.getBarHeight` above. It's one less context that needs to be created for an anonymous function

troyblank: Ah cool, makes sense, updated!

PR #1146, opened 2019-05-14
TFW-3542 - added sentiment percentage tooltips to vertical bar sentiment binning

troyblank:
adds sentiment tooltip for bars with sentiment binning

Screen Shot 2019-05-14 at 4 46 35 PM


components/visualizations/verticalBar/verticalBar.js
```diff
@@ -67,24 +70,25 @@ export default class VerticalBar extends PureComponent {
             .attr('class', 'vertical-bar');
     }
 
-    getColorClassForBar(d, i) {
+    getClassesForBar(d, i) {
         const { data } = this.props;
         const hasSentiment = Object.prototype.hasOwnProperty.call(d, 'sentiment');
         const hasSentimentBinning = this.hasSentimentBinning();
         const colorScale = [blues[8], blues[4], blues[0]];
```
tVeratti: Since you're not actually using the colorScale anymore except as a way of making a count of colors, should this instead be an int?

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -93,6 +97,26 @@ export default class VerticalBar extends PureComponent {
         return Object.keys(data[0]).some(key => SENTIMENT_KEYS.includes(key));
     }
 
+    buildToolTip = (d) => {
+        const { metaData } = this.props;
+        const { xKey, yKey } = metaData;
+        const hasSentimentBinning = this.hasSentimentBinning();
```
tVeratti: This reminds me, I will lookin into a possible better way to determine if there is binning using encoding. I believe the test data has an entirely separate indicator that we can use.

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -121,9 +145,9 @@ export default class VerticalBar extends PureComponent {
             .classed('vertical-bar__bar-wrapper', true);
         innerBarWrapper = barWrapper.append('div')
             .style('width', d => `${d[KEY_PERCENT]}%`)
-            .attr('class', (d, i) => this.getColorClassForBar(d, i))
+            .attr('class', (d, i) => this.getClassesForBar(d, i))
```
tVeratti: Can be `.attr('class', this.getClassesForBar)` because it does pass index by default too.

components/visualizations/verticalBar/verticalBar.js
troyblank: Cool, from my talks with various people it did not seem like this would be in the encoding. But if it is that would be cool.

components/visualizations/verticalBar/verticalBar.js
troyblank: Sorry will have to talk to you in person about this, not exactly sure what you mean. The `colorScale` is being used and it needs to be ranged to 1 of those three colors. I reference the colors back to the original array used for linear scales in case they shift by a bit.

components/visualizations/verticalBar/verticalBar.js
troyblank: updated.

components/visualizations/verticalBar/verticalBar.js
tVeratti: Right, it's just a ranged, but you're not actually using any of the values in the array - so it seems like you just need like `const numColors = 3` instead of `colorScale.length()`

troyblank: is it not used here?

```classes = ` color-vis-${colorScale[Math.floor((i / data.length) * colorScale.length)]}`;```

The colors do show.

tVeratti: Nevermind, I did not see the actual array being accessed, totally blind.

troyblank: I would rather you make 1 out of every 1000 comments that had a bit of blindness than none at all :)

PR #1149, opened 2019-05-15
TFW-3518 - Finish up duration hook up for histograms

tVeratti:


tVeratti:
Two pieces that are highlighted by finishing this work:
- Bar values seem to fade out when hovered.
- Precision probably needs to be added to value labels.

Not in scope of this story but we might need to add a bug/story to address.

assets/js/mockData/histogram.js
```diff
@@ -1364,54 +1360,40 @@ export default {
...
         {
             values: [
-                {
-                    duration: 2.0444816675972457,
-                    doc_count: 21492,
-                    key: ''
-                },
                 {
                     duration: 459.61900049480454,
-                    doc_count: 2021,
```
tVeratti: I spoke with Jordan about the duration data. They will be removing `doc_count` from these values because they are not necessary and make it unclear which value should be rendered.

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -92,9 +92,8 @@ export default class VerticalBar extends PureComponent {
     }
 
     hasSentimentBinning() {
-        const { data } = this.props;
-
-        return Object.keys(data[0]).some(key => SENTIMENT_KEYS.includes(key));
+        const { data, binning } = this.props;
```
tVeratti: Data with binning seems to have this object attached to it. We can shortcut the keys check if it's there, but I left the followup just in case the API doesn't provide this when it should.

components/visualizations/wordBar.js
```diff
@@ -14,7 +14,7 @@ export const addPercentsToDataLoop = memoize((values, yKey, hasTabNavigation) =>
     if (hasTabNavigation) {
         values.forEach((d) => {
             const dataValues = d.values;
-            const maxValue = dataValues[0][yKey];
+            const maxValue = dataValues.sort((a, b) => b[yKey] - a[yKey])[0][yKey];
```
tVeratti: We're gonna have to introduce front-end sorting with the controls on the histograms, but we can move it upward when we're ready to add those controls.

components/visualizations/verticalBar/verticalBar.js
```diff
@@ -92,9 +92,8 @@ export default class VerticalBar extends PureComponent {
...
+        return binning || Object.keys(data[0]).some(key => SENTIMENT_KEYS.includes(key));
```
jasonhale: is `binning` sentiment specific, or used for general binning purposes?

tVeratti: Hmm good question actually, I hadn't seen it present on any other data yet. Do we have binning other than sentiment for the vertical bars?

components/visualizations/wordBar.js
```diff
@@ -29,6 +29,31 @@ export const addPercentsToDataLoop = memoize((values, yKey, hasTabNavigation) =>
...
+        return metaData;
+    } else {
+        // Return basic encoding values...
+        return {
+            xKey: encoding.x.fields[0],
+            yKey: encoding.y.fields[0]
+        };
+    }
+});
+
```
troyblank: This is great, one thing that seems to be missing though in this PR is one or so unit tests around this. Seems prime for this as you can expect a return like clockwork.

components/visualizations/verticalBar/verticalBar.js
jasonhale: Not yet.  Just a general question.  More speaking to the fact that the checks for "binning" that seems generic inside a function that is specific to "sentiment".

This works, and no wordbars have anything other than sentiment at the moment (except maybe avg-histogram?)

components/visualizations/wordBar.js
tVeratti: Sure I'll add some extra :)

PR #1152, opened 2019-05-15
TFW-3532 - Add metric to VerticalBar graph

jasonhale:
Adds that average text to the bottom of the wordbar when it's passed.

also...
- set `percentileRangeAvgHistogram` to use the WordBarChart component
- set sizes for that and the `avgDonut` so they display correctly

Screen Shot 2019-05-15 at 4 54 31 PM


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -180,7 +181,21 @@ export default function VizzyNizzy({ type, results, ...props }) {
...
+    case AVG_HISTOGRAM: {
+        const { values, encoding, average } = results;
+        isPending = Boolean(!encoding && !values);
+        visProps = {
+            values,
+            encoding,
+            isPending,
+            decimalPrecision,
+            average,
+            showAverage: true,
```
tVeratti: Could `showAverage` be determined based on if a non-undefined value is given for `average` or are there cases where we pass an avg and don't want to show it?

jasonhale: Crap.  That should've been removed.  Thanks for catching.

jasonhale: And fixed.

tVeratti: Thanks! Extra credit xD

PR #1153, opened 2019-05-15
TFW-3524 - added variable heights for vertical bars in visualizations

troyblank:


jasonhale:
Everything looks good.
One thing... this makes the bars on SilenceRatio much bigger than the design.  That ok, or does there need to be a way to disable this?

Screen Shot 2019-05-15 at 5 16 18 PM

jasonhale:
OH!  And labels aren't centered
Screen Shot 2019-05-15 at 5 18 04 PM

troyblank:
@jasonhale  the heights are based on the amount of data per design rules, so yea depending on data it will be bigger or smaller it could be big on any chart in the design if data is different. I see the label issue will fix soon.

troyblank:
@jasonhale fixed the centering issue, thanks for pointing out I totally looked over it.

components/visualizations/verticalBar/verticalBar.beta.scss
```diff
@@ -12,22 +12,34 @@
...
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        overflow: hidden;
+    }
 }
 
 .vertical-bar__bar-wrapper {
     position: relative;
     width: 100%;
+    margin-bottom: $space-base;
```
jasonhale: Just a suggestion... moving this margin declaration to `.vertical-bar__wrapper` would remove need for the `top: ...` positioning in both `.vertical-bar__label` and `.vertical-bar__text`.

troyblank: I know I tried this last night as I had the same thought but something stopped me from doing it... but for the life of me I forgot what....going to try again here in a bit just to make sure as it's a good idea.

troyblank: Actually not sure what my issue was, worked perfectly this time around :) thanks for the improvement.

jasonhale: Awesome.  Thanks for updating!

PR #1154, opened 2019-05-15
TFW-3522 - Histogram sort toggle

tVeratti:


tVeratti:
@troyblank @jasonhale Updated and ready

components/visualizations/wordBar.js
```diff
@@ -65,24 +72,86 @@ export default class WordBar extends PureComponent {
...
     onToggle = ({ value }) => {
         this.setState({ activeValue: value });
     }
 
-    getMenuFromData = data => Object.keys(data).map(key => ({ label: key, value: key }));
+    onSortUpdate = (e) => {
+        const { encoding } = this.props;
+        const { value } = e;
+        let sortBy;
+        if (value.indexOf(ALPHA_KEY) !== -1) {
```
jasonhale: Can we just use `value.includes(ALPHA_KEY)`...?  I don't remember if IE11 is a requirement.

troyblank: word to that.

tVeratti: good call - updated!

PR #1163, opened 2019-05-17
TFW-3517 - Add legend to silence ratio

tVeratti:
I'm putting this in draft because I want to wait for TFW-3521 to get merged in before rebasing to it. I will handle the conflicts there and handle the sentiment legend in the new method.


tVeratti:
> The addition of the legend is pushing that "Average" out the bottom. Just needs some flex-box care.
This is strange, I do not see it reproduced locally. What browser was this with?
Screen Shot 2019-05-20 at 9 52 59 AM

tVeratti:
Ok so the issue came up in base layout / vizzy, where vertical-bar containers had a height of `100%`. It is no longer necessary so removing the height fixed it and did not interfere with other vertical bars.

PR #1164, opened 2019-05-20
TFW-3529 - update auto focus parameters

tVeratti:
NOT 4.45 - HOLDING ON MERGE


tVeratti:
I spoke with Josh about the "+ Add" button animation triggering on close. He said it's not a major concern right now. Rest is fixed.

PR #1165, opened 2019-05-20
TFW-3606 - Allow graphs to resize in their containers

jasonhale:
Allows graphs to fill their container vertically.


components/visualizations/donutGraph.beta.scss
```diff
@@ -1,12 +1,15 @@
 .donut-card {
     display: flex;
-    align-items: center;
+    flex: 1;
+    // align-items: center;
```
troyblank: remove dis

components/visualizations/donutGraph.beta.scss
```diff
@@ -1,12 +1,15 @@
...
-    height: 100%;
+    // height: 100%;
 }
 
 .donut-card__container {
     display: flex;
+    flex: 1;
     flex-flow: column nowrap;
-    height: 100%;
+    // height: 100%;
```
troyblank: remove all commented out lines in this file.

jasonhale:
Concerning managing heights - setting the container to 100% is only part.  The heights of the svgs had to be calculated since the height is factored in to the y-scaling for the `range()`. 
Example:
```js
this.yScale = d3.scaleLinear()
    .domain([domainMin, domainMax])
    .range([this.contentHeight - marginTop, marginBottom]);
```
Correct me if I'm wrong in how d3 requires range to be calculated.

As for the images of the graphs extending below - those are caused by the header content changing size.  There's already a ticket out there that addresses the card headers.

Just thinking through it - i wonder if addressing that, if necessary, should be handled in a follow up if only because it may cause some rework of how graphs are re-scaled on `window.resize`.  I'll look into it.

troyblank:
@jasonhale  ah yea you are right, as the line graphs have to be a portion of some of the charts.  Disregard my comment about the 100% height thing. Yea so we just have a little rendering issue and this will be good, thanks!

jasonhale:
@troyblank - the container wasn't allowing charts to expand their container when needed.  Fixed that issue by forcing certain graphs (vertical bar graphs, so far) to use the minimum allowed space.  Running through and seeing if there's anything else I missed.

PR #1170, opened 2019-05-22
TFW-3672 - added color mapping to nps scores

troyblank:
Screen Shot 2019-05-22 at 1 48 16 PM


components/visualizations/donutGraph.js
```diff
@@ -14,7 +14,7 @@ export default class DonutGraph extends PureComponent {
         id: PropTypes.string.isRequired,
         data: PropTypes.arrayOf(
             PropTypes.object
-        ).isRequired,
```
tVeratti: Thank you :)

PR #1172, opened 2019-05-22
TFW-3703 - only sort wordbar when canUpdateSorting is true

tVeratti:


components/visualizations/wordBar.js
```diff
@@ -79,10 +79,10 @@ class WordBar extends PureComponent {
         sortMode: SORT_DESC
     }
 
-    getSortedValues = memoize((values, sortBy, sortMode) => orderBy(values, [sortBy], [sortMode]));
+    getSortedValues = memoize((values, sortBy, sortMode, allow) => (!allow ? values : orderBy(values, [sortBy], [sortMode])));
```
jasonhale: Not a blocker - but it's usually easier to quick-grok if the comparison is truth-based (instead of using a "not"/!).

like: 
```js
...(allow ? orderBy(values, [sortBy], [sortMode]) : values)));
```

tVeratti: Hmm, I can dig it. Will update!

PR #1173, opened 2019-05-23
TFW-3638 - Show label for single dashboard

jasonhale:
Screen Shot 2019-05-22 at 6 26 09 PM


jasonhale:
Yeah, the tooltip is no longer necessary and marked in the requirements of the ticket.  Thanks for calling that out!

PR #1174, opened 2019-05-23
 TFW-3641 - donut now can enlarge its diameter if it has a metric

troyblank:
Screen Shot 2019-05-23 at 9 21 32 AM


components/visualizations/donut/donut.js
```diff
@@ -44,28 +44,34 @@ export default class Donut extends PureComponent {
     };
 
     layout = {
-        width: 300,
-        height: 300,
-        margin: 56
-    };
-    segmentGap = 0.02;
+        width: 332,
```
jasonhale: the styles dictate the svg has a max-width of 300px.  Setting the viewbox greater than that means the content will always be scaled down (which I'm guessing is why segmentSize is 44 instead of 40?).

Changing these to 300, and segment size to 40, could bring the graph closer to the size/ratio pictured in the designs.  Unless there's a reason i'm not catching for those sizes?

troyblank: Good point; the `max-width` is not even needed anymore as we don't plot outside that as we are using specific values instead of percentages.

I updated the code to remove this and make those values more of what they aim to be.

Without this change does get closer but it will never be exact unless the container of the graph is at least 300 x 300, which in our layout it's usually not. Which is cool for now I feel but maybe later we can tweak the overall layout of things to get them "perfect".

Screen Shot 2019-05-23 at 10 22 33 AM

components/visualizations/donut/donut.js
```diff
@@ -103,22 +117,17 @@ export default class Donut extends PureComponent {
...
-        const pieHeight = this.layout.height - this.layout.margin;
-
         const showMetricText = Boolean(metric);
 
         // Pie like layout setup
-        const radius = Math.min(pieWidth, pieHeight) / 2;
-        const segmentArc = d3.arc().innerRadius(radius * segmentSize).outerRadius(radius - 1);
-        const textSegmentArc = d3.arc().innerRadius(radius * (segmentSize + 0.5)).outerRadius(radius + 23);
-        const hoverArc = d3.arc().innerRadius(radius * (segmentSize - 0.05)).outerRadius(radius + 3);
+        const segmentArc = d3.arc().innerRadius(radius - segmentSize).outerRadius(radius - 1);
```
tVeratti: I appreciate all of these numbers as vars, makes a lot more sense when reading.

troyblank: It's always good when you can improve and add functionality at the same time :)

PR #1182, opened 2019-05-24
TFW-3710 - Contain wordbars within content boundary

jasonhale:
This keeps the WordBars, as a group, from extending past the content (and right edge).
Screen Shot 2019-05-24 at 1 43 00 PM


jasonhale:
Holdup... re-doing bad function names...

PR #1188, opened 2019-05-28
TFW-3581 - Remove overlapping Donut labels

jasonhale:
Screen Shot 2019-05-28 at 3 44 15 PM


components/visualizations/donut/donut.js
```diff
@@ -162,13 +162,67 @@ export default class Donut extends PureComponent {
...
             this.addCenterText(innerDonutWidth);
         }
     }
 
+    removeOverlappingLabels() {
+        let prevText;
+
+        /* istanbul ignore next: d3 chaining will not render to jsdom in react currently */
+        // eslint-disable-next-line func-names
+        this.texts.selectAll('text').each(function (d, i) {
```
troyblank: This seems like it's doing way too much for one literal. This is what I would suggest we do but feel free to question it. We make a util method for collision detection that determines if eleA collides with eleB, this literal then uses that method to compare previous with current text and then alters based on that. This way collision can be tested (as it should be), and the dom manipulation can be isolated. This also helps with reusability and readability (we can focus on the manipulation and collision separately).

jasonhale: nah, fair assessment.  Lemme clean up.  (I originally wanted to chain this off where the `this.texts` are being placed (just above) but was running into "this" scope issues.)

jasonhale:
@troyblank - Updated to isolate the collision detection function.  Thanks for the input.

components/visualizations/util.js
```diff
@@ -275,4 +275,28 @@ export function wrapText(target, width) {
     });
 }
 
+export function getTranslateBounds(node) {
```
troyblank: This might be use-full outside visualizations... but until then we should remember this is here.

PR #1194, opened 2019-05-29
TFW-3782 - added a none sentiment to word bar graphs

troyblank:
I also removed any not calculated sentiment from the tooltip.

Screen Shot 2019-05-29 at 4 47 52 PM


test/components/visualizations/wordBar.spec.js
```diff
@@ -164,14 +164,13 @@ describe('Word Bar Chart', () => {
     });
 
     it('should render a sentiment legend', () => {
-        const wrapper = dive(<WordBarChart label={title} id={id} values={valuesSentiment} metaDatas={metaDatas} />);
-        const reverseSentimentColors = [...sentiment].reverse();
+        const wrapper = dive(<WordBarChart label={title} id={id} values={valuesSentiment} encoding={encoding} />);
```
jasonhale: Jenkins is hollering about this one.  Looks like WordBarChart still uses `metaDatas`, not `encoding`.

troyblank: Yea it was a bad auto merge, just fixed it @jasonhale

jasonhale: ha!  looks like I commented the same time you were fixin.  Thanks!

components/util/number.js
```diff
@@ -21,6 +21,8 @@ export function boundNumber(number, min, max) {
 // If altTotal is set use that instead of allowing the numbers themselves calculate the total.
 // If altTotal is set result percentages will not equal 100.
 export function calculatePercent(numbers, altTotal) {
+    if (0 >= numbers.length) return numbers;
```
jasonhale: an array length is always a non-negative integer - shouldn't have to check if it's less than 0.  Makes it look like there's some reason it might be.  Is there a different reason for that type of comparison?

troyblank: Ah, yea you are right, I just was being all overkill like I try to be on checks. I just made it a ```===```

PR #1196, opened 2019-05-30
TFW-3737 - added bookmark button to universal search

troyblank:
I added a button (if you have permissions).

Screen Shot 2019-05-30 at 12 10 00 PM

Added disabled state to icon buttons as well ass tooltip prop.

This does everything the requirements say besides for the click action. I figured it would not be worth doing since the modal has yet to be worked on.


troyblank:
I forgot I need to make this only enable when a user has a query, need to close for a bit.

components/headerPane/headerPane.beta.scss
```diff
@@ -83,3 +73,38 @@ $header-pane-height-small: $space-xxl;
...
+        position: relative;
+        margin: 0 $space-xs;
+
+        &:after {
+            position: absolute;
+            top: 50%;
+            right: -$space-xs;
+            width: 1px;
+            height: $space-md;
+            margin-top: calc(-#{$space-md} / 2);
```
jasonhale: Another option, if you don't want to use any vars and margin, is 
```css
transform: translateY(-50%);
```

troyblank: Oh yea good call, i will update it when I reopen.

PR #1197, opened 2019-05-30
TFW-3737 - added bookmark button to universal search

troyblank:
added a button (if you have permissions).

Screen Shot 2019-05-30 at 12 10 00 PM

Added disabled state to icon buttons as well ass tooltip prop.

This does everything the requirements say besides for the click action. I figured it would not be worth doing since the modal has yet to be worked on.


components/headerPane/headerPane.beta.scss
```diff
@@ -83,3 +73,38 @@ $header-pane-height-small: $space-xxl;
 .header-pane__content {
     padding: 0 0 $space-sm;
 }
+
+.header-pane__icon-navigation {
+    display: flex;
+    position: absolute;
```
jasonhale: This was ok when it was one icon of determinate width.  But adding extra content expands the space - which may cause overlapping issues with long titles and/or use of the `headerNavigation` prop and the content it adds.

Screen Shot 2019-05-31 at 9 05 38 AM

Using flexbox above the `$size-small` and absolute positioning below could help mitigate the issue.  Or possibly continue with flexbox and use reordering to get placement right.

troyblank:
@jasonhale ok updated to be more responsive thanks. There is coverage issues on develop right now... so this did not introduce them FYI.

components/iconButton/iconButton.js
```diff
@@ -9,13 +9,15 @@ export default class IconButton extends PureComponent {
     }
 
     render() {
-        const { icon, active } = this.props;
+        const { icon, active, disabled, tooltip } = this.props;
         const iconProps = { filled: active };
 
         return (
           <button
-            className={styles['icon-button']}
```
tVeratti: Thanks for updating this :)

components/universalSearch/universalSearch.js
```diff
@@ -136,6 +137,23 @@ class UniversalSearch extends PureComponent {
 
     renderVisualizations = () => <Dashboards />
 
+    renderBookmarkButton = () => {
+        const { canCreateBookmarks, queries, tableData = {} } = this.props;
+        const { results } = tableData;
+        const hasQueries = 0 < queries.length;
+        const hasResults = Boolean(results);
+
+        const bookmarkButton = canCreateBookmarks
```
tVeratti: Not gonna block for this change, but it might be easier to write
```
if (canCreateBookmarks) {
    return (
         <IconButton ... />
    );
}
```

And let the function return undefined when false.

troyblank: ah, yea good call, I will updated that on the current work I am doing after this PR then. :hattip:

jasonhale:
It's super close.  But, below 700px wide, the icons stack below the header.
Screen Shot 2019-06-03 at 12 02 55 PM

if we wrap the content from `headerNavigation` prop in a _div_, and use flex's `order` below the 700px breakpoint, we can get this to flow correctly.  To go from this:
Screen Shot 2019-06-03 at 1 18 46 PM

to this:
Screen Shot 2019-06-03 at 1 33 04 PM
Screen Shot 2019-06-03 at 1 32 42 PM

troyblank:
@jasonhale  sorry about that, I must have made a last minute change and messed it all up. But now it's working. I don't understand your screenshots with the bookmark button and the create navigation, those should not exists together currently. I know there is no logic to stop that from happening but there is also no design that instructs what to do either.

PR #1201, opened 2019-06-03
TFW-3741 - added bookmark creation modal to the create bookmark flow

troyblank:
This builds off of PR #1197

This just gets the mechanics of the modal showing and hiding for creating a bookmark.

Screen Shot 2019-05-31 at 11 23 29 AM


components/universalSearch/universalSearch.js
```diff
@@ -136,8 +142,38 @@ class UniversalSearch extends PureComponent {
...
+        const hasResults = Boolean(results);
+        let bookMarkButton;
+
+        if (canCreateBookmarks) {
+            bookMarkButton = (
+              <IconButton
+                icon={<IconBookmark />}
+                disabled={!hasQueries || !hasResults}
+                tooltip={'Create Bookmark'}
+                onClick={() => { this.showBookmarkModal(); }}
```
troyblank: @tVeratti  I know we would prefer to have the actual instance functions be arrow functions but for some reason  you can't access them via `UniversalSearchComponent.prototype` if you do, so there was no way to test them that I could figure out. So maybe you know a better way.

test/components/universalSearch/universalSearch.spec.js
```diff
@@ -275,4 +286,51 @@ describe('Universal Search', () => {
         assert.isTrue(wrapper.exists('IconButton'));
         assert.isTrue(wrapper.exists('IconBookmark'));
     });
+
+    it('should be able to show the bookmark creation modal', () => {
+        const showBookmarkModal = sinon.spy(UniversalSearchComponent.prototype, 'showBookmarkModal');
```
tVeratti: You're right that they do not show up on prototypes. They're bound to the instance.

That's why I'd push to use the instance instead of the prototype for tests. We kinda talked about this before and something was a little weird but I know after the fact that Jason found the issue and how to make it work (I think we had a small typo).

Essentially you should just move the spy down after the wrapper and
```
const instance = wrapper.instance();
const showBookmarkModal = sinon.spy(instance, 'showBookmarkModal');
```

tVeratti: https://github.com/topbox-rci/rci-components/blob/7d52e235b8b032f66117a8fb7bba19b7e4cc74cd/test/components/datepicker/calendar.spec.js#L155

troyblank: So I tried this at first since this is how we do things, the problem is getting past the router, I get this error when doing as advised;

``` 2) Universal Search
       should be able to show the bookmark creation modal:
     TypeError: Attempted to wrap undefined property showBookmarkModal as function
      at wrapMethod (node_modules/sinon/lib/sinon/util/core/wrap-method.js:70:21)
      at Object.spy (node_modules/sinon/lib/sinon/spy.js:35:16)
      at Context.<anonymous> (test/components/universalSearch/universalSearch.spec.js:306:41)

  3) Universal Search
       should be able to hide the bookmark creation modal:
     TypeError: Attempted to wrap undefined property hideBookmarkModal as function
      at wrapMethod (node_modules/sinon/lib/sinon/util/core/wrap-method.js:70:21)
      at Object.spy (node_modules/sinon/lib/sinon/spy.js:35:16)
      at Context.<anonymous> (test/components/universalSearch/universalSearch.spec.js:331:41)```

The only way to get this to work is to ignore there is a button there at all which is half the reason for this text to exist (to make sure the conditions are right for a click).... maybe we need to talk about this in real time to solve it but I am not seeing a way to even see this on the instance as well as prototype.

tVeratti: Add this to a list for reasons why should take time to upgrade out testing platform :)

I forgot about the inability to access the instance in the HOC which is really weird but I trust the time you spent looking into to. Will approve just remember this when we're making a case for some technical downtime stories :)

PR #1202, opened 2019-06-03
TFW-3821 - [Bugfix] Wordbar card spacing

jasonhale:
Specificity in the Test env gets a little... funky.  But this probably needed to be more specific anyway.

Using the same specificity that's used in `vizzyCard.beta.scss`.


jasonhale:
Yeah - no bueno. Looks like `.helper__content-wrap` has extra padding declaration that gets goofy at that size.  Crap.

jasonhale:
@troyblank - added the small update for that spacing.
Screen Shot 2019-06-04 at 8 03 13 AM

PR #1203, opened 2019-06-03
TFW-3807 - [Bugfix] Hard code metadata for histogram x axis

tVeratti:


components/universalSearch/visualizations/vizzyNizzy.js
```diff
@@ -264,8 +264,8 @@ export default function VizzyNizzy({ type, results, ...props }) {
             const interactionsKey = encoding.y.fields[0];
             const secondsKey = encoding.y.fields[2];
             metaDatas = [
-                { xKey: encoding.x.fields[0], yKey: interactionsKey },
-                { xKey: encoding.x.fields[0], yKey: secondsKey, tooltipKeys: [secondsKey, interactionsKey] }
+                { xKey: 'doc_count', yKey: interactionsKey },
```
jasonhale: Changing this causes the label for the bar to be the displayed value, not the proper label...
Screen Shot 2019-06-03 at 4 10 07 PM

tVeratti: I will update this shortly, thanks :)

PR #1205, opened 2019-06-04
TFW-3740 - added all fields that will be available for creating a bookmark

troyblank:
This just adds general functionality of how the form fields work. There is a lot of style issues and this story does not tackle the logic of making the bookmark global, but it's a good step forward.

Screen Shot 2019-06-04 at 9 44 57 AM

The user should be able to view and input the Details section of the bookmark including the following in order:

* Title
* Owner
* Description
The user should be able to input a "Title"
The user should be able to view the "Data Channel" field showing the current Engagement the signed in user is in (pre-populated)
The user should be able to view the "Owner" field showing the current signed in user is in (pre-populated)
The user should be able to input a "Description" to describe additional human context about the Bookmark
The Description max character length of the description should be set to 200 characters
The user should be able to see a status of how many characters out of 200 they have input in the upper right-hand corner of the text box
If the user reaches the 200 character count the text box should not allow any further text input
Ability to set a visibility rule attribute of “Only Me” or “Global”


test/components/universalSearch/bookmarkModal.spec.js
```diff
@@ -48,13 +56,54 @@ describe('Bookmark Modal', () => {
...
+                </ul>
+                <Button styleVariations={['small', 'primary']}>Cancel</Button>
+                <Button styleVariations={['small', 'primary']}>Create</Button>
               </form>
             </Modal>
           </div>
         ));
     });
+
+    it('should render not sowing by default', () => {
```
tVeratti: *Showing :)

troyblank: oopies fixed.

components/form/select/select.js
```diff
@@ -219,6 +219,7 @@ export class Select extends PureComponent {
                   className={'select__button'}
                   name={name}
                   id={id}
+                  type={'button'}
```
jasonhale: 💯

components/universalSearch/bookmarkModal.js
```diff
@@ -37,10 +44,45 @@ export default function BookmarkModal({ showing = false, onHide }) {
...
+                  name={'status'}
+                  label={'Status'}
+                  options={[
+                    { label: 'Active', value: 'active' },
+                    { label: 'Inactive', value: 'inactive' }
+                  ]}
+                />
+              </li>
+              <li>
+                <Input
```
jasonhale: The other inputs have an ID - need one here, also?

troyblank: Yup! it does. i just added with my current work, so expect it next PR!

PR #1210, opened 2019-06-05
TFW-3729 - added new styles for inputs

troyblank:
OK here is what i did here. I looked at the Input, TextArea, and Select components and conformed them to the new style and also made sure they looked decent on the bookmark modal.

Screen Shot 2019-06-05 at 4 02 34 PM

I double checked most places to make sure nothing broke visually, hopefully I did not miss anything. There still needs work to do the following. 1. update modal styles to match new design, 2. fix the issue over the calendar making scroll bars in the modal instead of overflowing outside the modal. 3. update calendar picker styles in general (pretty sure there is a ticket for this).


assets/sass/base/silents/_forms.scss
```diff
@@ -1,3 +1,5 @@
+$input-explicit-height: 5.2rem;
```
jasonhale: Should this go ahead and get changed to 4rem (not including the border) to match the designs in InVision?

troyblank: The problem is the invisions expect a 14px font, we have to do 16px or a huge bug on iOS appears :(

components/universalSearch/bookmarkModal.js
```diff
@@ -8,8 +8,8 @@ export default function BookmarkModal({ showing = false, onHide, user }) {
     return (
       <div className={'bookmark-modal'}>
         <Modal headline={'Create Bookmark'} isShowing={showing} onHiding={onHide}>
-          <form>
-            <ul className={'bookmark-modal__field-list'}>
+          <form className={'helper__form'}>
+            <ul>
```
jasonhale: didn't realize it before... and this is only nit-picky... but why an unordered list?  Seems like having to rework the styles that are inherit in lists would be unnecessary compared to `div` instead.

no big deal.  just a curiosity to me.

jasonhale:
Timepicker fonts are all bolded.  Don't know if that's an issue - maybe tack on to the datepicker style probs.

Screen Shot 2019-06-06 at 3 25 00 PM

components/universalSearch/bookmarkModal.js
troyblank: I dunno, I have always seen it done this way for forms. I feel like it's semantic as it really is an unordered list of fields. But up for debate if we want to change it.

troyblank:
@jasonhale  the new inputs 

> Timepicker fonts are all bolded. Don't know if that's an issue - maybe tack on to the datepicker style probs.
> 
> Screen Shot 2019-06-06 at 3 25 00 PM

@jasonhale  the new inputs are semi bold, so I assumed this is desired... sent a question to casey about this so we will see which way is the happy path with him i guess.

components/universalSearch/bookmarkModal.js
jasonhale: Heh - just the opposite for me in practice and experience.  Seems non-semantic as it's not really a list of things as much as it is just vertically oriented content.  ¯\_(ツ)_/¯
But, yeah, no issues.  Just wondering if there was a bigger purpose.

troyblank:
> > Timepicker fonts are all bolded. Don't know if that's an issue - maybe tack on to the datepicker style probs.
> > Screen Shot 2019-06-06 at 3 25 00 PM
> 
> @jasonhale the new inputs are semi bold, so I assumed this is desired... sent a question to casey about this so we will see which way is the happy path with him i guess.

@jasonhale  casey says leave it as is, thanks for checking on it though never know when something is not as intended.

components/universalSearch/bookmarkModal.js
troyblank: I am def not married to it; I could see how divs would just be as fine. Since this currently is nothing new for our forms (in the past) I am going to count this as good for this PR but if you ever want to change it up so long as @tVeratti  doesn't have a strong opinion, I am cool with changing this up.

tVeratti: I think the two most important reasons for being semantic are SEO and A11y. We don't care about SEO so in the case of A11y, the difference I believe is that the `ul` gets read a bit differently. It'll say like `4 items` and then list them.

We should probably think about how it would impact the users and the best experience. I don't think I'd consider form fields as part of an unordered list, but it's also not going to cause any major issues imo if we had it.

troyblank: Here is a reason for ul's > http://www.jfciii.com/blog/2007/12/31/use-of-ul-or-ol-in-html-forms/

but like i said I am not advocating we stick with it.

tVeratti: Interesting, that does elaborate on the screen reader part. I'm fine with either way.

jasonhale: The article brings up an interesting point - i hadn't thought of the idea knowing _how many fields_.  Works as long as our forms are all vertical stacks - which they are so far i believe.

I haven't worked with accessibility in that regard and cursory investigation into more recent data isn't bringing up much - but I see examples with and without the list.

so, yeah, i'm good either way.

PR #1211, opened 2019-06-06
TFW-3768 - Bookmarking, bookmark visibility

jasonhale:
_This accompanies work being done in TFW-3768 in RCI._

Checks permission for whether a user can create a global or personal-only bookmark, and updates possible choices.

Modified a few props (user, channel) in the process of working through this.

Extra: removed the unused timezone function throwing off code coverage checks.


components/util/timezone.js
```diff
@@ -1,23 +1,9 @@
...
-    const date = new Date(initialDate);
-    const year = date.getFullYear();
-    const month = date.getMonth() + 1;
-    const day = date.getDate();
-    const time = date.toTimeString().split(' ')[0];
-
-    return moment.tz(`${year}-${month}-${day} ${time}`, 'YYYY-MM-DD hh:mm:ss', timezone || defaultTimezone).utc().format(API_FORMAT);
-};
-
-
```
troyblank: duuuuuuude, thanks!!!

components/universalSearch/bookmarkModal.js
```diff
@@ -2,8 +2,14 @@ import React from 'react';
...
 
-export default function BookmarkModal({ showing = false, onHide, user }) {
-    const { channel, name } = user;
+export default function BookmarkModal({ showing = false, onHide, channel, userName, canCreateGlobalBookmarks }) {
+    const { name } = channel;
+    const visibilityOptions = [];
+    if (canCreateGlobalBookmarks) {
+        visibilityOptions.push({ label: 'Global', value: 'global' });
+    }
+    visibilityOptions.push({ label: 'Only Me', value: 'onlyMe' });
```
tVeratti: In prep for these, I added a util file with two key constants in RCI. We'll want to decide how to keep those in sync.

jasonhale: After taking a look and your explanations, i agree that maybe they don't even need to be in RCI.
Added here under the `universalSearch` folder labelled `bookmarkUtil.js`.
Thanks for pointing it out.

PR #1212, opened 2019-06-06
 TFW-3733 - added bookmark header to search header

troyblank:
Screen Shot 2019-06-06 at 5 06 15 PM

This was a tad easier than anticipated so I put an extra commit here which figures out how to fit `<SelectDropDown />` into the new button on this header. The idea here is that anything not passed parent layout information assumes the select is already in something it doesn't need to be placed within. I tested this out with this PR and it's going to work great I think. To see it in action just manipulate the 'show' state of `<HamburgerNavigation />` (no handler yet).

I was unsure if this needs to hide based on permissions, but would be easy to do so if it did.


assets/js/components.js
```diff
@@ -548,6 +549,12 @@ renderHtml(
         location={'/customerExperience/contactDuration'}
         onMenuSelection={selection => print(selection, `Menus selection made: ${selection.topLevelMenuValue}`, false)}
       />
+      <br />
+      <br />
+      <br />
+      <HamburgerNavigation label={'Can Haz Hamburger'} />
```
tVeratti: 👍

components/navigation/hamburgerNavigation/hamburgerNavigation.js
```diff
@@ -0,0 +1,41 @@
...
+            <button
+              className={'hamburger-navigation__button'}
+              type={'button'}
+            >
+              { label }
+              <IconMenu />
+            </button>
+            <div className={'hamburger-navigation__drop-down'}>
+              <header>
+                <h1>Bookmarks</h1>
```
tVeratti: I've always understood `h1` as the page title or main heading. Might want to dip this one down a bit to either `h2` or `h3`?

components/selectDropDown/selectDropDown.js
```diff
@@ -314,7 +315,7 @@ export default class SelectDropDown extends PureComponent {
 
     assignMenuPosition = () => {
         const {
-            parentLayoutInformation,
+            parentLayoutInformation = {},
```
tVeratti: Would it make sense to use default props here? The main issue to consider there is that if the layout is set in one render, then removed in another, the default props don't kick in later. I don't think we have any cases where that scenario is true though.

components/navigation/hamburgerNavigation/hamburgerNavigation.js
jasonhale: Yep - i'd agree.  `<header>` doesn't introduce a new section in the outline - so there could be competing h1's.
Is it a styling decision?

jasonhale: Heh - just realized this could've been addressed in the future work for this actual dropdown bit (TFW-3731) - sorry dude.  ;)

troyblank: All good i updated to h2!

components/universalSearch/bookmarking/bookmarkHeader.beta.scss
```diff
@@ -0,0 +1,18 @@
...
+.bookmark-header {
+    @include breakpoint-below($breakpoint-mobile) {
+        padding: 0 $space-md;
+    }
+
+    display: flex;
+    position: relative;
+    align-items: center;
+    justify-content: space-between;
+    height: $space-lg;
```
jasonhale: Sorry - piling it on here... the bookmark collapses after a search is returned.  Setting `flex: 0 0 3.2rem;` fixes the height when flexbox does it's reflow thang.

Screen Shot 2019-06-07 at 2 09 59 PM

components/selectDropDown/selectDropDown.js
troyblank: great suggestion, made this change thanks.

components/universalSearch/bookmarking/bookmarkHeader.beta.scss
jasonhale: Sorry - piling it on here... the bookmark collapses after a search is returned.  Setting `flex: 0 0 3.2rem;` fixes the height when flexbox does it's reflow thang.

Screen Shot 2019-06-07 at 2 09 59 PM

troyblank: Cool thanks fixed this looking into permissions.

PR #1213, opened 2019-06-07
TFW-3842 - fixed bug where tooltip would get stuck on vertical bars

troyblank:
I am confident I have made this significantly better. I am not confident I fixed every single mouse out, but I can't reproduce it anymore out of many times of trying. I think this might be as fixed as it gets given you can very rarely reproduce the bug on regular buttons which have no direct tootlip manipulation and is possibly a symptom of a bug in the lib itself.


troyblank:
@tVeratti  no, the reason is because this vizzy was special in biz rules where the whole thing needed to be hoverable but the tooltip ONLY showed on the bar. We haven't had anything that specific past this.

PR #1216, opened 2019-06-08
TFW-3751 - Add inline toggle to input

tVeratti:


components/form/iconPopout.beta.scss
```diff
@@ -19,6 +19,10 @@ $arrow-size: 4px;
 
         .icon-popout__button:after { @include arrow($arrow-size, $color-white, 'up'); }
     }
+
+    &--disabled {
```
jasonhale: This definitely works.  Would it be better to override the cursor at the button instead?
```css
.icon-popout__button {
  ...
  &:disabled {
    cursor: default;
  }
}
```

Didn't know if there were other issues involved here.

tVeratti: I think I was planning to do more with this but decided restyling the popout would probably fit better as its own styling update story.

jasonhale: Yeah.  Ok.  This is odd to me, but it works!  Thanks

components/form/input/input.beta.scss
```diff
@@ -143,3 +143,22 @@
...
+
+    .input__label {
+        margin-bottom: 0;
+    }
+
+    .input__wrap {
+        flex: 1;
+    }
+
+    // Toggle
```
troyblank: is this really for the toggle? I never would have guessed :P ... I would consider removing this redundant comment in seriousness but not a show stopper.

components/form/toggle.beta.scss
```diff
@@ -32,45 +32,45 @@
     }
 
     :checked ~ div {
-        .nob {
+        .toggle__nob {
```
troyblank: awesome thanks for making old stuff up to snuff.

jasonhale:
@troyblank - that's biz rule is handled by work I'm doing that builds off this.  PR coming soon.

PR #1218, opened 2019-06-10
TFW-3731 - added bookmark selector

troyblank:
Screen Shot 2019-06-10 at 4 26 15 PM

I thought about linking some key events to the select drop down from extra accessibility but it might be a little outside of this scope as it's doable but a little tricky...and we don't have requirements for it so I didn't want to spend too much time outside of the requirements. But could easily be added as another PR.


troyblank:
not sure how to get out of github saying i have conflicts... will reopen this.

components/navigation/hamburgerNavigation/hamburgerNavigation.beta.scss
```diff
@@ -24,8 +30,23 @@
...
+        margin-bottom: $space-xs;
+    }
+}
+
 .hamburger-navigation__drop-down {
     @extend %action-shadow;
+
+    @include breakpoint($size-take-over-width) {
+        width: 100%;
+        //height: 100%;
```
jasonhale: Left an oops

troyblank: Cool, yea def ooops but fortunetly that was taken care of with another PR after this, so it's gone on develop 😎

PR #1220, opened 2019-06-11
TFW-3743 - Bookmarking, daterange toggle

jasonhale:
Changed the default to the new toggle to `false`.
Added logic to clear an input field's value when toggle gets set from on to off.
Added date range changing state to BookmarkModal


troyblank:
appears the lag is due to the date helper not the toggle.

PR #1224, opened 2019-06-13
TFW-3736 - Bookmark Error

tVeratti:


tVeratti:
One test to fix, rest should be reviewable.

components/universalSearch/universalSearch.js
```diff
@@ -194,13 +203,18 @@ export class UniversalSearchComponent extends PureComponent {
...
-                onHide={() => { this.hideBookmarkModal(); }}
-              />
+              { showingBookmarkModal &&
+                <BookmarkModal
+                  onHide={() => { this.hideBookmarkModal(); }}
+                />
+              }
+              {
+                bookmarks.loadError &&
+                <Modal headline={messages.BOOKMARK_ERROR_HEADLINE} isShowing={true}>
```
troyblank: This seems right on as you don't wanna render the html until there is a error, but also curious if `isShowing` could control if it renders or not 🤔  anyway just ramblings here nothing concrete.

PR #1226, opened 2019-06-14
TFW-3756 - Bookmark title

jasonhale:
Adds the Bookmark title to the head of Search.  Guessing at shape of bookmarks - will check and update accordingly.

Modified the `<Header />` a tiny bit to make the title's area fit to its content better. Haven't seen issue yet - downsides?


components/universalSearch/universalSearch.js
```diff
@@ -171,6 +172,17 @@ export class UniversalSearchComponent extends PureComponent {
         this.setState({ showingBookmarkModal: false });
     }
 
+    renderBookmarkTitle() {
+        const { bookmark } = this.props;
```
tVeratti: The active bookmark will be found under
`const { activeBookmark } = this.props.bookmarks`

jasonhale:
Haven't had a flakey test yet, so not sure about tidying that up.  However, talked over with Casey, the hover state didn't make sense.  So, just removed the dashed line.  PR updated from review comments.

However, noticed that I'm not setting the query when a bookmark is loaded.  Focusing on that next - requires translation from v2 query to search v1.

PR #1229, opened 2019-06-17
TFW-3894 - added char count to max length text areas

troyblank:


components/form/textArea/textArea.js
```diff
@@ -32,7 +35,11 @@ export default class TextArea extends PureComponent {
     }
 
     onChange = (e) => {
-        this.setState({ value: e.target.value });
+        const { value } = e.target;
+        const { maxLength } = this.props;
+        const charactersLeft = maxLength ? maxLength - value.length : maxLength;
```
tVeratti: I think this is something that doesn't need to be kept in state to be honest. If the value of the input changes for any reason other than the direct `onChange` we'd have it out of sync. Instead I feel like this number can just be counted at render - it's not remotely an expensive calculation so I'm not worried about it affecting any kind of perf.

What do you think?

components/form/textArea/textArea.beta.scss
```diff
@@ -29,12 +29,26 @@
     @extend %forms-clear-button;
 }
 
+.text-area__character-count {
+    position: absolute;
+    top: $space-base;
```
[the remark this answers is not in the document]
troyblank: Yea the design was not really in an 8px grid for the counter so i fudged it one way, but I am happy to do the other way :) just changed. I left the right at the lower number because space is crucial.

components/form/textArea/textArea.js
troyblank: Yea I think that is much mo betta and less complicated. Changed.

PR #1231, opened 2019-06-18
TFW-3748 - Filter bookmarks

tVeratti:


assets/js/demo/components/bookmarkManage.js
```diff
@@ -2,6 +2,12 @@ import React from 'react';
 import { BookmarkManage } from '../../../../components/index';
 
 export default function BookmarkManageDemo() {
+    const bookmarks = { all: [
+        { id: 'a1', title: 'test 1', visibility: 'Global', username: 'First Last', status: 'Active', date: '00/00/000' },
```
jasonhale: The API documentation has been updated with all the fields a bookmark has.  For instance, there is a `dateRange` instead of `date`.  It just doesn't look like we've created full-blown bookmarks yet.
Correct me if i'm wrong on this one.

tVeratti: Yeah I just didn't bother mapping all of them for the demo since it's all defined really in the config in RCI. It could change later, but this is just whatever for the sake of seeing data.

jasonhale:
Code looks good - just have the failing test and merge conflicts.

PR #1235, opened 2019-06-19
TFW-3739 - Handle save error conflicts

tVeratti:
Screen Shot 2019-06-19 at 8 53 46 PM

Screen Shot 2019-06-19 at 7 33 50 PM


components/universalSearch/universalSearch.js
```diff
@@ -164,6 +164,45 @@ export class UniversalSearchComponent extends PureComponent {
         return bookMarkButton;
     }
 
+    renderBookmarkLoadError = () => {
+        const { messages } = this.props;
+        return (
+          <Modal headline={messages.BOOKMARK_ERROR_HEADLINE} isShowing={true} styleVariations={['small']} closeable={false}>
```
troyblank: `closeable={false}` is trapping some users, they should be able to cancel on some errors instead of forcing them to save for example.

tVeratti:
Screen Shot 2019-06-20 at 3 35 56 PM

components/modal/modal.beta.scss
```diff
@@ -1,6 +1,6 @@
 @import '../../assets/sass/base/base';
 
-$modal-width--small: 48rem;
+$modal-width--small: 56rem;
```
jasonhale: We have a var set for this... `$space-xxxl`.  Reason for not using that?

tVeratti: Nope not really, just felt more like an abstract size that something for padding/margin, but I can change it to use the spacing vars.

jasonhale: In this file a few sizes are set directly instead of using the global spacing vars.  Should consider using the vars (especially in instances like font sizes), unless there was a reason not to.

jasonhale: > Nope not really, just felt more like an abstract size that something for padding/margin, but I can change it to use the spacing vars.

Yup.  that makes sense.  Though in cases like font sizes or margins for spacing, i'd say those are less abstract.  Not a huge deal.  Was just something i noticed.

PR #1237, opened 2019-06-20
TFW-3760 - Modify bookmark in US

jasonhale:
Sets up being able to update a bookmark via the search bar when the query is changed.  Will have related work in RCI.

Adds the new Save icon - replacing the old one.
Shows the save button and "edited" callout when the query changes.


jasonhale:
> ...With this PR there is no way to do a normal search and see the flow of making a new bookmark.

No, there's not.  We hadn't had the bits to demo triggering a bookmark load until very recently - which is why it was just coded in.  I'll see about adding that flow.

jasonhale:
@troyblank - demo able to apply a bookmark set via URL.

jasonhale:
Demo updated

jasonhale:
> you set a search and you get a save icon and not a normal create bookmark icon.
- fixed.

> although if you go back to a normal search ( by clicking search) it doesn't appear to clear out your bookmark
Clearing out a bookmark isn't specified in requirements and outside the scope of this work.  I ran that case through Casey.  Working it out.  We'll need to add in a ticket to address that.

> ...clear the search you still get the ability to save, which I think is not good as we should not allow creating bookmarks with no queries.
Yeah, makes sense to me.  Not spelled out in requirements.  We'll need to check on the business expectations of that and update.  I'd argue that's out of the scope of this work, too.

> ...link to demo...
Updated.  Now under Bookmarks, titled "Bookmark Loaded"

troyblank:
@jasonhale  makes sense yea I agree those flows aren't really laid out in the requirements and would agree we need more stories to cover these. Let's be sure to bring this up in standup and if you get time feel free to make the tickets to be assessed, hopefully we both won't forget.

PR #1241, opened 2019-06-21
TFW-3735 - More Button

tVeratti:
Screen Shot 2019-06-21 at 4 18 10 AM

Permissions affecting the options showing will be added next - I spoke with TA and she's working adding some properties to each bookmark on the search call to indicate the permissions specific to that item (not broad as we've done with all other permissions).


tVeratti:
This is expanded upon and fixed in the PR for TFW-3837.
I'm going to close this PR and update the other with this ticket's name as well, instead of trying to navigating the conflicts and backwards fixes.

PR #1243, opened 2019-06-21
TFW-3905 - Update loading indicator

tVeratti:


components/universalSearch/universalSearch.js
```diff
@@ -238,10 +248,12 @@ export class UniversalSearchComponent extends PureComponent {
             bookmarkOptions,
             canViewBookmarks,
             manageBookmarkURL,
-            bookmarks = {}
+            bookmarks = {},
+            activeBookmark
         } = this.props;
 
         const { showingBookmarkModal } = this.state;
+        const isLoading = activeBookmark && !activeBookmark.query;
```
jasonhale: I added `activeBookmark.loaded` in the reducer to note when an active bookmark was done loading.  Though, checking for a query is less verbose and makes total sense (as there should not be a bookmark without a query).

I'll see about updating my PR.

tVeratti: Ah sorry for not seeing the difference. If you're cool with going this way then I won't change here.

PR #1245, opened 2019-06-24
Get queryIdCount on Searchbar update

jasonhale:
Noticed that when queries get updated when a bookmark is loaded, the id's of the query pills might be duplicated (mainly id: 0) - causing some pills to not show.


components/universalSearch/searchBar.js
```diff
@@ -37,8 +37,18 @@ class SearchBar extends PureComponent {
...
     }
 
-    componentDidUpdate() {
+    componentDidUpdate(prevProps) {
         this.shouldFocusAddButton();
+
+        this.checkQueryLength(prevProps.queries, this.props.queries);
+    }
+
+    checkQueryLength = (prevQueries, queries) => {
```
troyblank: There is no test case for attempting to set the `queryIdCount` when the length of `preQueries` matches the current ones here.

jasonhale: @troyblank - thanks for saving my bacon on that one.  Good call.  Going full comparison.  PR updated.

PR #1247, opened 2019-06-24
Tfw 3767

troyblank:
View mode of bookmarks, needs designed new radios yet and removal of date input, but otherwise disables work.

Also cancel button will be replaced with "edit" when edit modal works :)

Screen Shot 2019-06-24 at 11 45 36 AM


components/universalSearch/bookmarkModal.js
```diff
@@ -5,18 +5,49 @@ import { VIS_ONLY_ME, VIS_GLOBAL, STATUS_ACTIVE, STATUS_INACTIVE } from './bookm
...
+    const [status, setStatus] = useState({ label: 'Active', value: STATUS_ACTIVE }); // eslint-disable-line no-unused-vars
+    const [dateRange, setDateRange] = useState(); // eslint-disable-line no-unused-vars
+    const newMode = mode === NEW_MODE;
+    const viewMode = mode === VIEW_MODE;
+
+    const getHeadline = () => {
+        let headline;
+
+        switch (mode) {
+        case VIEW_MODE:
```
jasonhale: Is the edit mode title handled in a different ticket?

tVeratti: Yeah I'm guessing so - there's a whole story for Edit

troyblank: yea correct current PR out there for edit now.

PR #1248, opened 2019-06-24
 TFW-3747 - added edit mode on bookmark modal

troyblank:
builds off view bookmark modal; adds edit mode.


components/universalSearch/bookmarkModal.js
```diff
@@ -41,6 +44,9 @@ export default function BookmarkModal({ bookmark = {}, mode = NEW_MODE, onHide }
         case VIEW_MODE:
             headline = 'Bookmark Details';
             break;
+        case EDIT_MODE:
+            headline = 'Edit Bookmark';
```
tVeratti: `messages` are already available in the context so it should be real fast to add these in there and use. Not gonna request that change but if you have time, it's good for consistency.

troyblank: oh yea good call, i might make it another PR after since so much is hinging on this including my current work lol. But i wrote it down to do!

PR #1251, opened 2019-06-24
TFW-3967 - changed bookmark drop downs to radios

troyblank:
Making this a two parter since we are all working in the same area
; this updates the drop downs to radios functionally. Going to add the label and switch to beta.scss for radio groups next!

Screen Shot 2019-06-24 at 4 51 45 PM


components/form/radioGroup.js
```diff
@@ -26,17 +26,17 @@ export default class RadioGroup extends PureComponent {
     }
 
     getInput(value) {
-        const { id, disabled } = this.props;
+        const { name, disabled } = this.props;
         const isChecked = String(this.state.value) === String(value);
 
         return (
-          <input type="radio" name={id} value={value} onChange={this.onChange} checked={isChecked} disabled={disabled} />
+          <input type="radio" name={name} value={value} onChange={this.onChange} checked={isChecked} disabled={disabled} />
```
tVeratti: Good catch here.

troyblank: Yea making sure all inputs sorta act the same, I updated the whole whopping 3 places we used this component :)

PR #1252, opened 2019-06-24
 TFW-3974 - Navigate on bookmark selection

jasonhale:


assets/js/demo/components/universalSearch.js
```diff
@@ -57,11 +85,18 @@ export default class UniversalSearchDemo extends Component {
...
 
-        if (location.search) {
+        if (location.search.length > 0) {
             const { bookmarks } = this.state;
+            const id = location.search.split('id=')[1];
+
+            const optionedbookmark = this.bookmarkOptions.find(b => b.value === id);
 
-            if (!bookmarks.activeBookmark) {
+            console.log(optionedbookmark);
```
troyblank: forgot to remove this :)

components/universalSearch/bookmarking/header.js
```diff
@@ -10,17 +10,13 @@ export default class BookMarkHeader extends PureComponent {
...
         return (
           <div className={'bookmark-header'}>
             <IconBookmark size={'medium'} />
             <HamburgerNavigation label={'Bookmarks'} onOpen={this.onSelectorOpenHandler}>
               <BookmarkSelector
-                options={options}
                 ariaID={'Bookmarks'}
                 ref={selector => (this.selector = selector)}
-                manageBookmarkURL={manageBookmarkURL}
               />
```
troyblank: This has caused a weird coverage issue with ref 🤷‍♂

assets/js/demo/components/universalSearch.js
jasonhale: sunnuvab...

components/universalSearch/bookmarking/header.js
jasonhale: Coverage has been all over the place.  Will check this out.

troyblank: Yea using enzyme has not been great as of late, maybe when we get less busy we can talk about a transition to something like jest or something :/

jasonhale: SO... modified the component so coverage would be _correct_.

jasonhale:
Updated this PR to allow for actual coverage of the problematic function in bookmark selector component.

Also, options for the bookmarks selector are being build component-side, not in RCI and passed.

PR #1256, opened 2019-06-25
 TFW-3981 - formatted data on bookmark save form handler

troyblank:


assets/js/demo/components/bookmarkManage.js
```diff
@@ -5,10 +5,10 @@ import messages from '../../../../components/messages/messages.en-us';
...
     const bookmarks = { all: [
-        { id: 'a1', title: 'test 1', visibility: 'Global', username: 'First Last', status: 'Active', date: '00/00/000', description: 'Some description' },
-        { id: 'a2', title: 'test 2', visibility: 'Only Me', username: 'First Last', date: '00/00/000' },
-        { id: 'a3', title: 'test 3', visibility: 'Global', username: 'First Last', status: 'Inactive', date: '00/00/000', description: 'Some other description' },
-        { id: 'a4', title: 'test 4', visibility: 'Global', username: 'First Last', status: 'Active', date: '00/00/000' }
+        { id: 'a1', title: 'test 1', visibility: 'Global', user: { name: 'First Last', id: '123' }, status: 'Active', date: '00/00/000', description: 'Some description' },
+        { id: 'a2', title: 'test 2', visibility: 'Only Me', user: { name: 'First Last', id: '234' }, date: '00/00/000' },
+        { id: 'a3', title: 'test 3', visibility: 'Global', user: { name: 'First Last', id: '323' }, status: 'Inactive', date: '00/00/000', description: 'Some other description' },
+        { id: 'a4', title: 'test 4', visibility: 'Global', user: { name: 'First Last', id: '456' }, status: 'Active', date: '00/00/000' }
```
tVeratti: I didn't realize you meant in the bookmark objects. This is something we need to ask for from the API or we want to somehow map this from RCI in the reducer?

components/universalSearch/bookmarkModal.js
```diff
@@ -21,18 +21,18 @@ export function BookmarkModalComponent({ bookmark = {}, mode = NEW_MODE, onHide,
...
-    if (canCreateGlobalBookmarks) visibilityOptions.push({ label: 'Global', value: VIS_GLOBAL });
+    visibilityOptions.push({ label: 'Only Me', value: VIS_ONLY_ME, name: 'visibility' });
+    if (canCreateGlobalBookmarks) visibilityOptions.push({ label: 'Global', value: VIS_GLOBAL, name: 'visibility' });
 
     const defaultVisibilityOption = bookmark.visibility ? visibilityOptions.find(o => bookmark.visibility === o.label) : visibilityOptions[0];
     const defaultStatusOption = bookmark.status ? statusOptions.find(o => bookmark.status === o.label) : statusOptions[0];
     const [visibility, setVisibility] = useState(defaultVisibilityOption);
     const [status, setStatus] = useState(defaultStatusOption); // eslint-disable-line no-unused-vars
-    const [dateRange, setDateRange] = useState(); // eslint-disable-line no-unused-vars
+    const [description, setDescription] = useState(bookmark.description ? { value: bookmark.description, name: 'description' } : null); // eslint-disable-line no-unused-va
```
tVeratti: I can't believe I didn't notice the description was missing!

components/universalSearch/bookmarkModal.js
```diff
@@ -57,10 +57,11 @@ export function BookmarkModalComponent({ bookmark = {}, mode = NEW_MODE, onHide,
     };
 
     const onSubmit = (e) => {
-        const form = {};
-        [title, visibility, status].forEach((s) => {
-            form[s.name] = s.value;
+        const form = { owner: user.id, mode };
```
tVeratti: I'm actually not sure how necessary this (`user.id`) is, but it can't hurt.

components/universalSearch/bookmarkModal.js
```diff
@@ -108,7 +109,7 @@ export function BookmarkModalComponent({ bookmark = {}, mode = NEW_MODE, onHide,
                   id={'bookMarkOwner'}
                   name={'owner'}
                   label={'Owner'}
-                  defaultValue={bookmark.username || userName}
+                  defaultValue={bookmark.user ? bookmark.user.name : user.name}
```
jasonhale: if this is 'user' from global user state, `user.name` should be `'${user.firstName} ${user.lastName}'`

assets/js/demo/components/bookmarkManage.js
troyblank: either way works, we will just need both name and id when a bookmark is loaded.

components/universalSearch/bookmarkModal.js
troyblank: yea, it's the lesser of the things broken right now though :)

components/universalSearch/universalSearch.js
```diff
@@ -52,8 +52,9 @@ export class UniversalSearchComponent extends PureComponent {
     }
 
     onBookmarkSave = (bookmark) => {
-        const { onBookmarkSave } = this.props;
-        onBookmarkSave(bookmark);
+        const { onBookmarkSave, queries } = this.props;
+
+        onBookmarkSave({ bookmark, queries });
```
jasonhale: why not
```
bookmark.query = queries
```
and pass `bookmark`?

components/universalSearch/bookmarkModal.js
troyblank: I think it's needed in the bookmark save payload. I assume during editing the api has to know the user id in order to re-save the bookmark maybe but I am not fully sure either if it's really really needed as we can derive that from the current saved information i guess.

tVeratti: Yeah better to keep it in there. 👍

jasonhale: I believe the API will get the user from the token passed to the API

components/universalSearch/bookmarkModal.js
troyblank: What about the load from edit how is that going to look? Until that can be confirmed figuring out formatting seems arbitrary.

components/universalSearch/bookmarkModal.js
tVeratti: That' what I was thinking Jason, but this way it has both in case there's a reason it should overwrite or if any logic is needed there. Hence - I don't think it hurts :)

components/universalSearch/bookmarkModal.js
jasonhale: no - i mean that you're referencing "user" from the global store.  "user.name" isn't a property on there.  so regardless how this plays out - that's going to throw an exception.

troyblank: ah yea because they conflict in name, so it would not be the universal search prop it would be the global one... gotcha. Updated.

components/universalSearch/universalSearch.js
troyblank: done and done, thanks.

components/universalSearch/universalSearch.js
```diff
@@ -230,7 +231,7 @@ export class UniversalSearchComponent extends PureComponent {
             ];
 
         messaging[1] = messaging[1].replace('%title', activeBookmark.title);
-        messaging[1] = messaging[1].replace('%name', userName);
+        messaging[1] = messaging[1].replace('%name', user.name);
```
jasonhale: sorry - another user.name problem.

PR #1266, opened 2019-06-27
TFW-3997 - added max length of 60 to title of bookmarks

troyblank:
OK so i swear I added this but it must have went away with merge conflicts....

BUT instead of crying over spilled milk I made lemonade flavored milk from lemons or something, I used a now very supported `maxLength` dom property (like we do in textarea). This has a plus and minus along with it. Plus is we don't have to manage validation of length in the component and it allows default props to be large than the max, meaning if we decide to reduce the max later the old titles won't get cropped. The minus is that there is no way unit test this browser feature so I just checked if the prop exists because enzyme simulated events just force inputs the value into state no matter what without consideration of the way a real browser would work.


jasonhale:
> ...lemonade flavored milk...

::gag sounds::

PR #1273, opened 2019-06-29
TFW-4008 - added save option permission consideration and style tweaks

troyblank:
Screen Shot 2019-06-29 at 11 16 26 AM

Did not add close button, figured it was too much of an extra to do at this speed.


components/universalSearch/bookmarkButton.js
```diff
@@ -66,4 +76,4 @@ export class BookmarkButtonComponent extends PureComponent {
     }
 }
 
-export default withMessages(BookmarkButtonComponent);
+export default withMessages(onClickOutside(BookmarkButtonComponent));
```
jasonhale: wrap all the things!

jasonhale:
Screen Shot 2019-06-29 at 12 23 40 PM
in the designs, the dropdown covers the button.  will that be incoming work?

components/universalSearch/bookmarkButton.js
tVeratti: So if we used context on the BookmarkButtonComponent we wouldn't need `withMessages` nor would we need to pass the permission for `activeBookmark` - that might help a lot with wrapping in multiple layers.

Since it's not a functional component, you'd have to use the context consumer but it saves from have double HOC.

troyblank: Context seems to make testing a ton worse in my experience. But I will try this.

troyblank: I am going to need you to walk me through how our current context is set up for search. It seems it's set up to be an HOC which means I would be replacing one double HOC with another one. Hit me up when you got time thanks.

troyblank:
@tVeratti  I am going to go ahead and merge this. I realize you have a request to change but we sorta addressed it together so I am sure you are on board and i adjusted unit tests just like our others so they are very consistent. If you still have issues with this PR I promise to address them just let me know.

PR #1282, opened 2019-07-01
TFW-4012 - added last accessed information on bookmark details

troyblank:
Checks if info is there and shows it if it is. I used the docs on swagger to build this not seeing this info in the testing env yet.

Screen Shot 2019-07-01 at 9 31 13 AM


components/universalSearch/bookmarkModal.js
```diff
@@ -45,6 +47,9 @@ export function BookmarkModalComponent({ bookmark = {}, defaultMode = NEW_MODE,
     const [description, setDescription] = useState(bookmark.description ? { value: bookmark.description, name: 'description' } : null); // eslint-disable-line no-unused-va
     const saveButtonText = (editMode) ? messages.BOOKMARK_MODAL_SAVE_ACTION : messages.BOOKMARK_MODAL_NEW_ACTION;
     const isGlobal = visibility && visibility.value === VIS_GLOBAL;
+    const lastAccessed = bookmark && bookmark.lastOpened ? moment(bookmark.lastOpened).format('MM/DD/YY') : null;
```
jasonhale: Shouldn't have to check for `bookmark` as it's required.  If it's not being passed, this component will have issues well before these lines.

troyblank: This is fixed.

troyblank: But yea great call, totally don't need to be checking that.

troyblank:
@tVeratti  merging this in but if you see any issues I can fix later you have my word.

PR #1283, opened 2019-07-01
remove redundant script loads

jasonhale:
Super not-important.
Addresses where demo is loading scripts twice.  One of the issues this causes is loading and applying two instances of styles.

So now, when debugging your css, you'll only see one instance of styles instead of two.


jasonhale:
Ooooh - yeah.  Lookin'...

jasonhale:
Fixed, yo.

PR #1294, opened 2019-07-15
TFW-4084 - Remove search helper

tVeratti:


test/components/universalSearch/searchBar.spec.js
```diff
@@ -287,28 +233,28 @@ describe('Universal Search - Search Bar', () => {
...
-        );
-        const searchButton = wrapper.find('.search-bar__inputs button').at(1);
-        const expectedQueries = [{ id: 0, dimension, value, type: 'include', data: {} }];
-
-        wrapper.setState({ value, dimension });
-        searchButton.simulate('click');
-
-        assert.deepEqual(expectedQueries, testQueries);
-    });
+    // it('should be able to add queries by clicking the search button', () => {
```
troyblank: We saving this for it's next incarnation?

components/universalSearch/searchBar.js
```diff
@@ -232,39 +228,6 @@ class SearchBar extends PureComponent {
                   onBlur={this.onBlur}
                 />
               </div>
-              <div className={'search-bar__helper'}>
-                {/* Dimension Value */}
-                <HelperSwitch
```
troyblank: Is this component needed anymore? should it also be removed?

test/components/universalSearch/searchBar.spec.js
tVeratti: Yeah, it's close enough that I didn't want to completely nuke it. It will be useful.

components/universalSearch/searchBar.js
```diff
@@ -232,39 +223,6 @@ class SearchBar extends PureComponent {
...
-                  key={dimensionId}
-                  getRef={this.getValueRef}
-                  attributes={dimensionAttributes}
-                  onChange={this.onValueUpdate}
```
jasonhale: 3 functions here - `helperBlurred`, `onValueUpdate`, `helperFocused` - aren't being used anymore and should probably be removed and/or commented out.

components/universalSearch/searchBar.js
tVeratti: I think it can be used to make the modal, if at least through reviewing its code. I wouldn't remove it quite yet because they have a lot of similar functionality, jsut different display.

components/universalSearch/searchBar.js
tVeratti: Good catch, removing these now

PR #1297, opened 2019-07-16
TFW-4095 -  Add button style variation to autocomplete

tVeratti:
### Nakey Button
Screen Shot 2019-07-16 at 3 13 00 PM

### Focused (floating options)
Screen Shot 2019-07-16 at 3 13 07 PM


jasonhale:
The only concern i have is animation.  Its appearance is super abrupt.  Maybe we can address that later.

components/form/autoComplete/autoComplete.js
```diff
@@ -124,6 +127,16 @@ export default class AutoComplete extends PureComponent {
         this.setState({ itemFocus });
     }
 
+    onButtonActivate = () => {
```
troyblank: there is no unit test for when the button activates.

PR #1299, opened 2019-07-17
TFW-4108 - added a multi select drop down component

troyblank:
Added multi select styles / functionality for select drop down as well as made a little helper wrapper that allows multi select select drop down easier. It started handling state of the values selected but then realized that could change from outside so I let the demo handle that state for now and eventually that will be inside the helper modal.


jasonhale:
This is looking good - but what about _unselecting_ an option?  Are there requirements or stories around that?

troyblank:
@jasonhale  I talked to casey about this yesterday, you can not unselect an item from the multiselect. The only way to do it is to change the prop of `activeValue` which the left side of the help modal will control. He wanted a very simple flow with no way to do things two ways.

PR #1300, opened 2019-07-18
TFW-4087 - updated style of select drop down label

troyblank:
Screen Shot 2019-07-18 at 10 17 06 AM


components/selectDropDown/selectDropDown.beta.scss
```diff
@@ -90,9 +90,10 @@ $select-drop-down-separator-color: $color-grey-4;
 }
 
 .select-drop-down__label {
-    padding: $small-spacing $medium-spacing $tiny-spacing;
-    color: $color-grey-1;
+    padding: $space-xs $space-sm;
+    color: $color-black;
     font-size: 1.2rem;
+    font-weight: 700;
```
jasonhale: This could also be `$font-weight-bold`.

troyblank: updated this and will merge once jenkins is cool with it, thanks totally forgot about those vars.

PR #1303, opened 2019-07-19
TFW-4075 - Update query node styles

tVeratti:
Screen Shot 2019-07-19 at 3 12 26 PM


components/queryNodeList/queryNode.beta.scss
```diff
@@ -8,12 +8,20 @@ $node-height: $space-md;
...
+    border: 0;
     border-radius: $border-radius;
-    background-color: $color-white;
+    background-color: $color-grey-5;
     font-size: 1.2rem;
-    line-height: calc(#{$node-height} - 2px);
     white-space: nowrap;
+
+    &:hover {
+        background-color: $color-blue-2;
```
troyblank: I am cool with doing this when we add a tooltip since that is the action but as a note the design shows a pointer during hover.

tVeratti: Noted!

PR #1305, opened 2019-07-22
TFW-4127 - added new modal styles

troyblank:
Two things to note here 1. the cancel button styles are not done so I just made it a normal button for now. and 2. padding of content isn't 100% since it will probably be tweaked on the new design layout (since it's variable between examples in design) but works decently with older modals.

Screen Shot 2019-07-22 at 9 48 44 AM


components/modal/modal.beta.scss
```diff
@@ -22,16 +23,33 @@ $modal-width--small: 56rem;
...
+}
 
-    header {
-        @extend %clear-fix;
+.modal__header {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding: $space-xs;
+    border-bottom: 1px solid $color-grey-4;
```
tVeratti: This looks a little strange for modals without headlines. I'm not certain where those are in the app, but there's an example of it on the demo here.
Screen Shot 2019-07-22 at 10 34 02 AM

troyblank: Talked to Casey about this, and he is cool with it. BUT he did notice the X button was off by 1px center-wise so I removed `line-height` from it to fix it. Also made sure the header is explicit height and made a new story to add ellipses to header when it gets too large.

troyblank: Just added a new push that keeps the height at min of 40px on the header but expands as needed.

components/modal/modal.beta.scss
```diff
@@ -85,9 +111,8 @@ $modal-width--small: 56rem;
 
 .modal__close-btn {
     @extend %naked-button;
-    margin: -2rem -2rem 0 0;
-    padding: 2rem;
-    float: right;
+    margin: -.2rem 0;
```
jasonhale: with the line-height, doesn't seem like the negative margin is necessary anymore.

troyblank: You are correct; I removed just this line and will merge once jenkins is good.

PR #1306, opened 2019-07-22
TFW-4125 - added footer buttons to bookmark modals

troyblank:
This builds on a current PR, but does three things.

1. adds new button styles to bookmark forms.
2. allows multiple buttons to be added to the footer.
3. allows the cancel button to be hidden if desired.

Also as a bonus I fixed a weird spacing issue when the bookmark form had not edited information. It was causing a big white space if there is no info, now the container doesn't even render if there is none.


components/modal/modal.js
```diff
@@ -79,20 +80,26 @@ export default class Modal extends Component {
...
-                  <Button
-                    styleVariations={['medium', 'primary']}
-                    onClick={footerButton.action}
-                  >
-                    { footerButton.label }
-                  </Button>
+                  }
+                  {footerButtons.map(b => (
+                    <Button
+                      styleVariations={['medium', 'primary']}
```
tVeratti: Should they be allowed to pass a style variation? If nothing given, default to primary?

troyblank: Currently there is no style variation difference in the design, all buttons in the footer are `medium, primary`. Until there is a difference I see no reason to pass this personally as it has the double action of providing a safeguard for design standards.

troyblank:
I rebased this with `develop` so the negative margin on the close button is removed here as well.

PR #1307, opened 2019-07-22
TFW-4112 - Add cancel button style variation

tVeratti:
I decided to include a padding of 8px. The design isn't super clear for these, only that there is 16px between it and other buttons. This will allow for margins of 8px.


assets/js/components.js
```diff
@@ -100,6 +100,8 @@ renderHtml(
       <br /><br />
       <Button styleVariations={['standard', 'nakey']} icon={<IconPlus size={'large'} />}>Nakey</Button>
       <br /><br />
+      <Button styleVariations={['standard', 'cancel']}>Cancel</Button>
```
troyblank: Not sure this make sense being "cancel" as it is used a lot more places than canceling:

Screen Shot 2019-07-22 at 2 27 08 PM

tVeratti: 'Cancel' is an easier way to identify it instead of yet another synonym for hollow/naked. If you notice, even `Remove all` is a cancel-ish type of action - as in, they don't ever seem to be affirmative. The name is purely for our own reference, and it indicates generally how it used, even if the labeling isn't always exactly 'Cancel'.

troyblank: That is fair; it's just a name not a huge deal. The show stopper is the active styles, I won't stand in the way of how things are named as I am horrible at it :)

tVeratti: Yup, will update the active in a moment - thanks :D

PR #1308, opened 2019-07-22
TFW-4106 - Render New Dimension & open HelperModal

tVeratti:
This adds just a stub for the HelperModal. There's a story to build out its structure next.


troyblank:
ummm why no jenkins yet on this pr? 😕 @mathom any ideas ?

mathom:
It looks like its in the works:

Screen Shot 2019-07-22 at 4 17 36 PM

mathom:
NVM it needs the process from https://github.com/topbox-rci/admin-ui-legacy/pull/47 ported in like the rest of the node repos

components/universalSearch/searchBar.js
```diff
@@ -58,22 +58,12 @@ class SearchBar extends PureComponent {
 
     onDimensionUpdate = (selection) => {
         const { match: dimension } = selection;
-        // const { dimension: prevDimension } = this.state;
-        const { onDimensionChange } = this.props;
+        const { onDimensionUpdate } = this.props;
         const data = dimension && this.getDefaultData(dimension.id);
 
         this.setState({ dimension, data });
```
jasonhale: This is another bit I don't think we'll need.  I've commented it out in my PR in case it's needed as reference.  If you don't feel it's needed, just let me know and I'll clean out some of the commented code.

tVeratti: Actually I think you're right here. I was thinking of keeping it for the re-add, but this wouldn't be the right place for that anyway. I will remove it as well.

PR #1309, opened 2019-07-22
 TFW-4111 - added header children optional prop to modals

troyblank:
This builds off my other PRS; I stole some of @tVeratti  styles for this PR, but this allows you to add any custom header content to the top right of the modal.

Screen Shot 2019-07-22 at 3 56 14 PM


troyblank:
I rebased this with develop so the negative margin on the close button is removed here as well.

troyblank:
I rebased this with develop so the negative margin on the close button is removed here as well.

PR #1310, opened 2019-07-22
TFW-4072 - Use new query structure in US/Workflow

jasonhale:
_Goes with TFW-4072 on RCI_

Utilizes new query shape for Universal search (and Workflow).

Having concerns over the commented code in SearchBar.  If you see something commented out that maybe shouldn't be, please holler.


troyblank:
Why is there a lot of commented code in this PR? Wanted to know it's purpose before really looking at it deeply thx.

jasonhale:
The commented code is just building off the already commented code in SearchBar from previous commits.  Some code was commented out for later re-use in the HelperModal - I don't know if y'all need any of these extra comments that I've added, or if I can just delete.

components/universalSearch/searchBar.js
```diff
@@ -1,11 +1,11 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { isEqual } from 'lodash';
+import { isEqual, cloneDeep } from 'lodash';
 import { AutoComplete, QueryNodeList, Button } from '..';
 import { IconPlus } from '../icons';
 import optionPropType from '../util/propTypes';
-import helperMap from './helperMap';
+// import helperMap from './helperMap';
```
troyblank: any chance we can leave a comment on any commented code on what it's waiting for? my worry is we get further down the road and it's not removed and we do not know why it's commented out? Just something small like "waiting for new dimension modal picker".

troyblank:
@jasonhale  I would say document the stuff that is needed later with a small comment and remove the rest that is no longer needed. Right now it's sorta confusing what is going to be moved and what is dead.

jasonhale:
That's really where my concern is at.  I'd rather have deleted all the commented code and just use history in Git to retrieve what's needed.  My original comment is more for @tVeratti since she originally commented the code in case of need earlier.

@tVeratti - how much of this code do you think we'll need as reference?  I only added a few lines to the already commented code in `searchBar.js` as it's not used/needed from my POV.  If you'd like to keep, i'll comment it.  If not, i'll delete it.

components/universalSearch/searchBar.js
```diff
@@ -115,93 +111,92 @@ class SearchBar extends PureComponent {
...
+    //     const match = allDimensions.filter(d => d.id === dimensionId)[0] || {};
+    //     return match.attributes || ['autocomplete'];
+    // }
 
     render() {
         const { queryIdCount, dimensionInputFocused } = this.state;
         const { popularDimensions, dimensions, queries, messages } = this.props;
 
-        const hasQueries = Boolean(queries && queries.length);
+        const hasQueries = Boolean(queries && queries.nodes) && queries.nodes.length > 0;
```
tVeratti: Any reason we wouldn't want to just `queries && queries.nodes && queries.nodes.length`?

tVeratti: I also noticed in the UniversalSearch component, it just directly looks at `queries.nodes.length` - do we expect the structure to be super questionable at each level or can some of it be relied on, like if queries exists, nodes always will too?

test/components/workFlow/workFlow.spec.js
```diff
@@ -114,10 +114,10 @@ describe('Work Flow', () => {
         const resultsGrid = wrapper.find(`.${styles['results-grid']}`);
         const splash = wrapper.find(`.${styles.splash}`);
 
-        assert.isFalse(pendingAnimation.exists());
-        assert.isTrue(filterBar.exists());
-        assert.isFalse(resultsGrid.exists());
-        assert.isTrue(splash.exists());
+        assert.isFalse(pendingAnimation.exists(), 'pending animation should not exist');
```
tVeratti: Big fan of the messages here :) Thanks for adding those.

tVeratti:
Re: comments - Jason and I talked briefly about it, and I'd personally really rather leave them in until we're further along with the new helper. I will personally volunteer to do the cleanup at the end of the release if that makes you feel better about not leaving clutter when we're done :)

test/components/workFlow/workFlow.spec.js
jasonhale: Things started failing... had to hunt stuff down.  ;)

components/universalSearch/searchBar.js
jasonhale: Good catch.  That boolean shouldn't be necessary as the query should always have a value.  Updated!

jasonhale:
Brought in changes from develop, updated a bit from @tVeratti 's comments, and updated the Universal Search demo.

PR #1312, opened 2019-07-23
TFW-4107 - Render Helper Modal

tVeratti:
Also added a test case for the ConflictModal in manage since it surfaced after fixing our test coverage :)


components/modal/modal.beta.scss
```diff
@@ -164,7 +165,15 @@ $modal-width--small: 56rem;
     .modal__close-btn {
         display: none;
     }
+}
+
+.modal--top {
+    .modal__content {
+        margin-top: 7.1rem;
```
jasonhale: I wonder if that's supposed to be 72, not a 71 like in the designs... since it's not divisible by 8.

troyblank: Yea i would do 72, but it's not a huge dealyo

PR #1313, opened 2019-07-23
TFW-4110 - Render Helper Modal Header

tVeratti:
Screen Shot 2019-07-23 at 4 32 46 PM


components/modal/modal.beta.scss
```diff
@@ -28,7 +28,7 @@ $modal-width--small: 56rem;
 .modal__header {
     display: flex;
     align-items: center;
-    justify-content: space-between;
+    justify-content: flex-start;
```
jasonhale: 'flex-start' is default.  Not really necessary if 'flex-start' is what you're intending and not overwriting something.

troyblank:
Sorry I think this was just for the header.... content is placeholder I assume.

PR #1314, opened 2019-07-23
TFW-4120 - Update DurationPicker styles

jasonhale:
Screen Shot 2019-07-23 at 4 34 49 PM

Screen Shot 2019-07-23 at 4 36 20 PM


jasonhale:
@Troy - Updated, which shortened the height a little.  Better matches the design for height now.  (updates on right in img)

comparison

PR #1315, opened 2019-07-23
 TFW-4119 - updated date picker styles

troyblank:
Updated calendar styles and make sure it would work outside a drop down as well as in workflow and create form.

Screen Shot 2019-07-23 at 5 13 47 PM

Screen Shot 2019-07-23 at 5 13 58 PM


components/durationPicker/durationPicker.scss
```diff
@@ -2,11 +2,11 @@
 
 .duration-picker {
     display: inline-block;
-    border: 1px solid $color-grey;
+    border: 1px solid $color-grey-3;
```
jasonhale: I updated these in my PR, too.  💯

troyblank:
@jasonhale updated this PR so that the border only shows between calendars and not when there is only one. Thanks for that call out.

PR #1317, opened 2019-07-24
TFW-4080 -  Add tooltip to <QueryNode /> component

tVeratti:
Screen Shot 2019-07-24 at 3 35 30 PM


assets/js/components.js
```diff
@@ -777,7 +777,25 @@ renderHtml(
                     nodes: [
                         { displayLabel: 'Date', displayValue: 'Interaction : 00/00/00 – 00/00/00' }
                     ]
-                }
+                },
+                {
+                  operator: 'not',
+                  nodes: [
+                      { displayLabel: 'Agent', displayValue: 'Captain Marvel' },
```
troyblank: Bull; where is Rouge in here??? (everyone knows she can kick Captain Marvel's ass).

Rogue_(Anna_Marie)_(Earth-TRN259)_001

troyblank:
> The tooltip drops down, but floats over the querynode. Should be below it.

Yea, might need to provide and offset, not sure why that is happening as it's on the top element of the component :(

jasonhale:
Outside Troy's questions, this looks good.  Thanks!

components/queryNodeList/queryNode.js
```diff
@@ -1,8 +1,16 @@
...
 import { IconUIX } from '../icons';
 import './queryNode.beta.scss';
 
+const MAX_VALUES = 10;
+
 export default class QueryNode extends PureComponent {
+    // componentDidUpdate() {
+    //     ReactTooltip.rebuild();
+    // }
+
```
troyblank: this meant to be left out?

tVeratti: Nope, removing now! Thanks!

components/queryNodeList/queryNode.beta.scss
```diff
@@ -16,6 +16,7 @@ $node-height: $space-md;
 
     &:hover {
         background-color: $color-blue-2;
+        cursor: pointer;
```
troyblank: I hate to suggest this, but I feel like we need class that turns this on or off given the query node has more than 1 thing in it. Because it's weird when the cursor is pointer when there is no tooltip.

troyblank: Actually ignore me, I realize these will now be interact-able in the near future, so no point in my suggestion.

PR #1319, opened 2019-07-25
TFW-4118 - added selection display for date picker

troyblank:
Soooo, a lot of this is place holder mechanics so it's a little rough, but the main thing is getting the buttons hooked up to add a state of `selection`. If this is good I will next be working on a way to clear the calendar selection by adding a prop to optionally control selection for `<DatePicker />`.


components/universalSearch/helperModal.js
```diff
@@ -8,21 +9,39 @@ import './helperModal.beta.scss';
...
 
     onHelperChange = (displayValue, value) => {
         this.setState({ currentSelection: { displayValue, value } });
     }
 
+    onAddToSelection = () => {
+        const { currentSelection, selections: oldSelections } = this.state;
+        const selections = clone(oldSelections);
+
+        selections.push(currentSelection);
```
jasonhale: Next state - i think this is were we take the `currentSelection` add `id` and `displayLabel` to it so it's already ready to send back as a query.

test/components/universalSearch/helperModal.spec.js
```diff
@@ -26,50 +24,16 @@ describe('Helper Modal', () => {
...
-              { instance.state.currentSelection.displayValue }
-            </div>
-            <div className={'helper-modal__helper'}>
-              <HelperSwitch
-                attributes={activeDimension.attributes}
-                onChange={instance.onHelperChange}
-              />
-            </div>
-          </div>
-        ), 'did not render body');
```
troyblank: Iet me know if you want me to add these back but with the placeholder thing and such it seemed like too much moving stuff, but i sorta feel bad making it simpler.

jasonhale: There's a lot of moving parts.  This works as it's being built out, i think.  So, expectation to build back out tests once this bit stops moving?

components/universalSearch/helperModal.js
```diff
@@ -8,21 +9,39 @@ import './helperModal.beta.scss';
...
+        selections: []
     };
 
     onHelperChange = (displayValue, value) => {
         this.setState({ currentSelection: { displayValue, value } });
     }
 
+    onAddToSelection = () => {
+        const { currentSelection, selections: oldSelections } = this.state;
+        const selections = clone(oldSelections);
```
tVeratti: I'd love to try and use `[ ...oldSelections ]` since it's native syntax (even though we transpile it at the moment). One less dependency :)

tVeratti: That would allow you to write:
```
const selections = [
  ...oldSelections,
  currentSelections
]
```

troyblank: good call I changed to this; will merge soon as jenkins is good.

PR #1321, opened 2019-07-26
TFW-4116 - Add HelperList to HelperModal

jasonhale:
Adds the new HelperList to the universal search modal and to demo.

Screen Shot 2019-07-26 at 12 01 51 PM


components/universalSearch/helperModal.js
```diff
@@ -16,12 +16,32 @@ export class HelperModalComponent extends PureComponent {
         selections: []
     };
 
-    onHelperChange = (displayValue, value, rawValue) => {
-        this.setState({ currentSelection: { displayValue, value, rawValue } });
+    onHelperChange = (type, displayValue, value, rawValue) => {
```
tVeratti: Looks like we'll have a bit on conflicting code here. Shouldn't be too difficult to sort out, though.

components/universalSearch/helperModal.js
```diff
@@ -101,8 +118,10 @@ export class HelperModalComponent extends PureComponent {
               <div className={'helper-modal__helper'}>
                 <HelperSwitch
                   selection={rawValue}
+                  selections={selections}
                   attributes={activeDimension.attributes}
                   onChange={this.onHelperChange}
+                  onHelperValueChange={this.onHelperValueChange}
```
tVeratti: Could we rename this to `onValueChange` or `onInput`? It seems a bit redundant with the Helper included :) If you disagree it's cool to keep it as-is, but then I'd suggest `onChange` should be consistent.

jasonhale: `onValueChange` makes sense to me.  Will make that change.

jasonhale: @tVeratti - updated

PR #1322, opened 2019-07-29
TFW-4113 - Render Selections

tVeratti:
Screen Shot 2019-07-29 at 10 15 06 AM
Screen Shot 2019-07-29 at 10 15 20 AM


components/universalSearch/helperModal.js
```diff
@@ -64,11 +108,75 @@ export class HelperModalComponent extends PureComponent {
...
+                    <li className={
+                      classnames(
+                        'helper-modal__selection', {
+                          'helper-modal__selection--active': i === selections.length - 1 && showCurrentValue
+                        }
+                      )}
+                    >
+                      {s.displayValue}
+                      <IconButton
+                        icon={<IconClose color={'currentColor'} size={'small'} />}
```
jasonhale: This is `IconUIX` (with no size)... i know, confusing.  It's a slightly different build than IconClose.

components/universalSearch/helperModal.js
```diff
@@ -64,11 +108,75 @@ export class HelperModalComponent extends PureComponent {
...
+            <div className={
+              classnames(
+                'helper-modal__add',
+                'helper-modal__selection', {
+                  'helper-modal__selection--active': !selections.length
+                })}
+            >
+              {addButtonText}
+              <IconButton
+                icon={<IconPlus color={'currentColor'} />}
```
jasonhale: pretty sure this icon should be `size={'large'}`.

Screen Shot 2019-07-29 at 3 19 08 PM

components/universalSearch/helperModal.js
```diff
@@ -64,11 +108,75 @@ export class HelperModalComponent extends PureComponent {
         );
     }
 
+    renderSelections() {
```
jasonhale: I think this is just complex enough that it should be its own component.  Though, that isn't a show-stopper.

jasonhale:
Things are working good.  The selection buttons should be 32px tall - currently 40px tall.  Mentioned a couple other tiny icon things.

components/universalSearch/helperModal.js
tVeratti: I actually really considered doing this, but in the end I don't think it's really as complex as it is just a lot of lines of render. I think it would actually be more complex to manage to a new set of pass-throughs and callbacks to keep state in sync between the two, so the tradeoff didn't really seem like a win to me.

jasonhale: Looks like it could be a "dumb" component, and only work off props.  Granted, i haven't taken a dive into all the workings.  Either way, it's working.

PR #1324, opened 2019-07-30
TFW-4121 - Add Duration to HelperModal

jasonhale:
Allows duration picker to be set by props.


jasonhale:
EDIT: ...allows resetting.  Right now it's working in the helper modal and resets when a new selection is added.

Spent too much time as is and should have just converted it to be a completely controlled component.  Will look into that (may put a ticket into the backlog to address) - until then, PR is up.  Would appreciate thoughts.

components/durationPicker/durationPicker.js
```diff
@@ -72,26 +96,47 @@ class DurationPicker extends PureComponent {
...
             <TimePicker
               label={messages.MIN}
               onChange={this.onFromChange}
+              defaultValue={from}
             />
             <TimePicker
               label={messages.MAX}
               onChange={this.onToChange}
-              min={this.state.from}
+              min={from}
```
tVeratti: I've noticed that this always defaults to 00:00:01 even when the thing initially loads. In the design it looks like it's all 00:00:00, though.

jasonhale: Max always has to default to 00 00 01 - otherwise it's a duration of 0 seconds.  Minimum duration is 1 sec.  I'll confirm with Casey.

tVeratti: yeah that makes sense to me - just differed from the design. Whichever way you guys decide I'm on board with, just wanted to confirm.

jasonhale: @tVeratti - you're right, Casey and docs say it should be 0's all around until modified.  My oversight.

PR #1326, opened 2019-07-31
TFW-4124 - Allow helper to have mode, pass node/mode back on submit

tVeratti:
I haven't hit the tests for this yet because I want to go over one particular detail with you regarding the query structure before I call it ready.

Functionally you can add the query nodes now and they show up properly with the nested tooltips. I just need to meet you about the omitting and operators since I'm just missing that final detail and how ti set it up :)


tVeratti:
#1327 builds and improves on this, so this PR will probably be closed to allow the focus to be on the updated work.

PR #1331, opened 2019-08-01
TFW-4251 - HelperList - fix logic to show/hide search

jasonhale:
Fixes logic to show/hide the search bar in HelperList.

If options are less than 10, and there has been no search values entered, hide the search field.
If options are less than 10, and there has been a search value entered (even if a blank string), keep search field visible.

Also consolidated some of the functions communicating changes to the HelperModal as related to autocomplete fields.


components/universalSearch/helperList.js
```diff
@@ -46,7 +45,7 @@ export default class HelperList extends PureComponent {
 
         // Hide input when less than 10 options are shown...
         // unless that list was shortened by search input
-        const hideInput = options.length < 10 && !value;
+        const hideInput = options.length < 10 && value === undefined;
```
tVeratti: I feel like this could be a little weird if our api ever changes to return more than 10, but honestly I'm fine with making the assumption for now since we'd need to change this code in that scenario anyway.

components/universalSearch/helperList.js
```diff
@@ -11,32 +11,31 @@ export default class HelperList extends PureComponent {
...
     }
 
     onItemSelect = (selection) => {
-        const { value, label } = selection;
-        const { onBlur } = this.props;
-        this.setState({ value: '' });
-        onBlur({ value, displayValue: label });
+        const { value } = selection;
+        const { onChange } = this.props;
+        onChange({ value, isSelection: true });
```
tVeratti: I'm cool with removing `onBlur` and going with the `isSelection` option.

components/universalSearch/helperList.js
jasonhale: Yeah.  It should really be a constant and controlled by RCI itself. But I believe this is the only place that even cares about it.

PR #1332, opened 2019-08-01
TFW-4244 - Rework input interaction for non-autocomplete helper selections

tVeratti:
Screen Shot 2019-08-01 at 11 45 18 AM
Screen Shot 2019-08-01 at 11 46 07 AM
Screen Shot 2019-08-01 at 11 45 49 AM


components/universalSearch/helperModal.js
```diff
@@ -227,57 +206,57 @@ export class HelperModalComponent extends PureComponent {
...
                       classnames(
                         'helper-modal__selection', {
-                          'helper-modal__selection--active': i === selections.length - 1 && showCurrentValue
+                          'helper-modal__selection--active': i === selections.length - 1
                         }
                       )}
                     >
-                      {s.displayValue}
+                      {s.displayValue || messages.NEW_SEARCH_VALUE}
                       <IconButton
```
jasonhale: The "X" shouldn't be showing if there is no current value to display, if I remember correctly.

tVeratti: Good catch, the only time it would show in this case is when they had a value and deleted it with backspace. I'll add a check :)

components/universalSearch/helperModal.js
```diff
@@ -61,36 +65,37 @@ export class HelperModalComponent extends PureComponent {
...
+                this.setState({ selections });
+                return;
             }
         }
 
-        if (type === 'autocomplete') {
-            this.setState({ currentSelection }, () => {
-                this.onAddToSelection();
-            });
+        if (type !== 'autocomplete') {
```
jasonhale: Code style suggestion... would be more clear in this instance to keep the _if_ comparison specific (_if autocomplete, else all the other things_).  It's usually better to keep comparisons positive if possible.

not a deal-breaker or anything.

components/universalSearch/helperModal.js
```diff
@@ -187,81 +185,86 @@ export class HelperModalComponent extends PureComponent {
...
     renderSelections() {
-        const { selections, currentSelection } = this.state;
-        const hasSelections = selections.length || null;
+        const { messages } = this.props;
+        const { selections } = this.state;
+
         const showAddButton = !this.isInstantAdd();
-        const showCurrentValue = this.showValue();
-        const addButtonText = hasSelections ? 'OR' : 'New Search value';
+        const activeSelections = selections.filter(s => !!s.displayValue);
```
jasonhale: sneeaaakyyyyy

PR #1334, opened 2019-08-01
TFW-4253 - Add auto-scroll and new OR button

tVeratti:
Screen Shot 2019-08-01 at 3 43 13 PM


components/universalSearch/helperModal.beta.scss
```diff
@@ -4,7 +4,7 @@ $color-inactive: rgba($color-black, .5);
 
 .helper-modal {
     display: flex;
-    min-height: 20rem;
+    min-height: #{20rem - 5.7rem - 2.2rem};
```
jasonhale: Why the math?  A comment would help here.  And are any of those needing to be vars?

components/universalSearch/helperModal.js
```diff
@@ -74,12 +81,12 @@ export class HelperModalComponent extends PureComponent {
             }
         }
 
-        if (type !== 'autocomplete') {
+        if (type === 'autocomplete') {
```
jasonhale: 💯

components/universalSearch/helperModal.beta.scss
tVeratti: The minimum height of the modal was meant to be 200 from top of the header to the bottom, not just the content area. I subtracted the size of the header and the surrounding margins from the modal component itself. I can definitely explain that in a comment.

jasonhale: Yeah, that'd really help.  We have no reference to those measurements in case something changes.

jasonhale: yeah, that'd help.  or just set it to the right value.  as it is is a little confusing as to where those come from.

PR #1335, opened 2019-08-01
TFW-4226 - update option styles

jasonhale:
Fixes an issue with the row options in Manage Bookmarks from conflicting styles in MultiSelectDropDown.
Since MultiSelectDropDown doesn't have any extra html wrapper, moved the styles to SelectDropDown to keep things grouped.

Screen Shot 2019-08-01 at 3 51 28 PM

And a few minor tweaks to keep other styles from being affected.


components/selectDropDown/selectDropDown.beta.scss
```diff
@@ -199,3 +199,27 @@ $select-drop-down-separator-color: $color-grey-4;
         top: 0;
     }
 }
+
+.select-drop-down--multiselect {
```
tVeratti: This makes sense to nest the overriding styles under.

PR #1340, opened 2019-08-02
TFW-4252 - Hide options without labels from SelectDropDown

tVeratti:
There's a backend fix coming to remove the null data - but just to prevent bad data from presenting to the user, we can hide invalid options.


jasonhale:
I'm really concerned about this one and don't think we should.  It hides data issues, and can cause confusion in results.

tVeratti:
I'm going to cancel this for now since we don't want to hide the broken values and isntead rely on backend cleanup data.

PR #1342, opened 2019-08-02
TFW-4268 - Handle data flowing to freetext

jasonhale:
Data coming back to Input as type 'freetext' can be formatted for generic use or with channels.


components/universalSearch/helperMap.js
```diff
@@ -56,16 +56,27 @@ export default function (type, onChange, options, messages = defaultMessages, on
             }
         };
     case 'freetext':
-    default:
+    default: {
+        // freetext is used by channel selection also, so check for different formatting.
+        let defaultValue = null;
+        const value = selection && selection.value;
+        if (value) {
+            if ('object' === typeof value) {
```
tVeratti: Would you lose any logic by simplifying to:
```
const defaultValue = value.value || value;
```

jasonhale: No.  I don't believe so.  Will also protect with arrays.  Thanks

jasonhale:
@tVeratti - updated logic.

PR #1343, opened 2019-08-02
TFW-4270 - Disable commit button when no value

tVeratti:
Screen Shot 2019-08-02 at 4 03 14 PM


components/universalSearch/helperModal.js
```diff
@@ -185,6 +185,7 @@ export class HelperModalComponent extends PureComponent {
                   ariaID={'universalSearchConfirmButton'}
                   text={messages.SEARCH__CONFIRM}
                   onClick={this.onSearchQueryEdit}
+                  disabled={!activeDimension || !selections.length}
```
jasonhale: Would there ever be an instance where there would not be an `activeDimension`?

tVeratti: Hopefully not, but I went ahead and matched the logic to the existing Add/Omit - we might be able to remove from both but I went for the safety route :D

PR #1345, opened 2019-08-05
TFW-4239 - updated search header vertical spacing to better match design

troyblank:
Just some adjustments to vertical space of the header to better achieve the design spacing. Image below is of design overlaid on execution.

Screen Shot 2019-08-05 at 10 16 30 AM


components/headerPane/headerPane.beta.scss
```diff
@@ -28,7 +28,8 @@ $header-pane-height-small: $space-xxl;
     flex-wrap: wrap;
     align-items: center;
     justify-content: space-between;
-    min-height: $header-pane-height;
```
tVeratti: Does this `$header-pane-height` var still exist and is it referenced by anything else?
Can it be removed?

troyblank:
Closing this for a bit; noticed I screwed up workflow page

troyblank:
I am reopening this after casey update pattern lib to better match search. Header Pane's now have `0.8rem` top and bottom padding.

components/headerPane/headerPane.beta.scss
troyblank: Thanks, yea made sure these was not needed anymore removed it.

PR #1346, opened 2019-08-05
TFW-4239 - updated search header vertical spacing to better match design

troyblank:
Just some adjustments to vertical space of the header to better achieve the design spacing. Image below is of design overlaid on execution.

62475335-7478c900-b76a-11e9-9c38-4068e27a7f0b


components/headerPane/headerPane.beta.scss
```diff
@@ -2,23 +2,18 @@
 @import './breakpoints';
 @import './sizes';
 
-$header-pane-height: $space-xxxl;
```
tVeratti: 👍  :D

components/headerPane/headerPane.beta.scss
```diff
@@ -88,7 +89,7 @@ $header-pane-height-small: $space-xxl;
 }
 
 .header-pane__content {
-    padding: 0 0 $space-sm;
+    //padding: 0 0 $space-xs;
```
jasonhale: Left commented style here

troyblank: oops, yea and that class is also no longer needed, removed! thanks!

PR #1347, opened 2019-08-05
TFW-4240 - updated select labels to semi bold

troyblank:
* Shadow at the top of the menu should be removed
* Popular and All Dimensions labels should be in Semibold weight
* Fixed z-index issues with static Select Drop downs.

Screen Shot 2019-08-05 at 2 37 49 PM


components/selectDropDown/selectDropDown.beta.scss
```diff
@@ -223,3 +224,10 @@ $select-drop-down-separator-color: $color-grey-4;
         }
     }
 }
+
+// AUTOCOMPLETE PARENT
+.auto-complete {
```
tVeratti: In order to match the new design, I added a style variation called `floating` to the select drop down. I'm wondering if this override better belongs under that variation, since that's really what detaches the options from the top (and I think that's why we don't want a shadow?)

troyblank: But ummm I made a style variation called static that does the same thing... hmmmm

troyblank: nevermind after talking to you these are two different things.

troyblank: I moved this to the floating style, makes a lot of sense to me, thanks!

components/selectDropDown/selectDropDown.beta.scss
```diff
@@ -185,6 +189,7 @@ $select-drop-down-separator-color: $color-grey-4;
 .select-drop-down--static {
     position: relative;
     padding: 0;
+    z-index: $z-index-base;
```
jasonhale: Should we instead `unset` z-index here?  However, if this doesn't affect anything, I'm good with it also.

troyblank: `unset` still isn't very ubiquitous yet. So I would think keeping it with a new `z-index` is a little more helpful while doing the same thing.

jasonhale: `unset` is covered really well.  https://caniuse.com/#search=unset

No worries.  Just a thought.

troyblank: losing something like IE 11 is not enough to choose something else that does the same thing is my thought. If it was providing a new thing that was not comprisable it might be but worth considering, but like you said it doesn't really affect anything, might as well side with further support than the less supported value.

PR #1348, opened 2019-08-05
TFW-4237 - Update selection styles

tVeratti:
Screen Shot 2019-08-05 at 2 39 04 PM


components/universalSearch/helperModal.beta.scss
```diff
@@ -53,8 +54,25 @@ $color-inactive: rgba($color-black, .5);
         color: $color-black;
     }
 
-    .icon-button {
+    .helper-modal__or {
+        height: auto;
+        padding-top: .2rem;
```
jasonhale: I don't think this padding is necessary.  Without, the "OR" section is 16px tall, which matches the design.

tVeratti: I followed the comments in the story exactly, otherwise I would've agreed.
`OR divider should have 2px Padding on top and bottom (besides line-height of text) and be Regular weight`

troyblank: Yea not a fan that this doesn't fit in 8 grid... but I guess if casey is cool with this little tweak I am as well.

jasonhale: Nevermind.  I was looking at a non-updated design.  Sorry!  :)

PR #1350, opened 2019-08-05
TFW-4278 - Adjust defaultValue & reset HelperSwitch

tVeratti:


components/universalSearch/helperModal.js
```diff
@@ -101,18 +102,22 @@ export class HelperModalComponent extends PureComponent {
...
+            lastAction: 'added',
+            reset: !reset
+        });
     }
 
     onRemove = (i) => {
         const selections = [...this.state.selections];
         selections.splice(i, 1);
-        this.setState({ selections, lastAction: 'removed' });
+        this.setState({ selections, lastAction: 'removed', reset: !this.state.reset });
```
jasonhale: Due to how `setState` can be deferred or grouped with other state updates, toggles like this should use the functional callback version instead of passing an object.
```
this.setState(prevState => ({ selections, lastAction: 'removed', reset: !prevState.reset }));
```

tVeratti: I was actually a little worried if I used the callback that in the rare scenario where it does happen more than once in a cycle, it would end up just flipping back before the new key got applied to the render. I think you're a bit more familiar with the callback nuances, can you confirm that wouldn't happen?

jasonhale: deferred wasn't a good word.  Batched is more like it.  The flipping back is more likely in the case it's being used here since current state is referenced.

If you're worried about it flipping back and forth, use a unique key on each reset instead of toggling. Keys are meant to be unique anyway (which is why using indexes are frowned upon).

[React docs about state](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous)

jasonhale: OR.  since the issue was that the values weren't gettign reset... our Input ignores a defaultValue of `null`.

we can just coerce it to an empty string in helperMap instead of passing a `null`?  /shrug

jasonhale: OR. since the issue was that the values weren't getting reset... was it because of how Input interprets a `null` value?

Maybe we can just coerce it to an empty string in helperMap instead of passing a null - bypassing the need for a reset in the first place.

?

tVeratti: Regarding keys, I'm very well aware of how they are typically used as unique values in lists, but I've used this trick many times to great effect to force a component to reset (remount, etc). There are a lot of pieces of helper state that won't be directly tied to value, and it gives us a reliable clean slate every time.

As we discussed in slack, I think this gives us a bit of peace of mind for any and all future helpers that they truly are reset when the selection changes. Of those three functions, they should never get batched together since they're callbacks on specific events in the modal.

jasonhale: I'm OK with the reset via key - don't think it's necessary but won't argue against it's virtues.  However, the format of the setState in this instance is well documented and typically considered bad practice.

As far as `setState` referencing state directly as it is here, I'm ok with it if @troyblank is.

troyblank: Yea I am not sure here maybe we can do a quick call; but generally I agree with the sentiment that you should not use `this.state` in a `setState` as it's async, and if you need to do that you probably should use a callback such as:

```this.setState((prevState, props) => {
  return {counter: prevState.counter + props.step};
})```

But maybe after a more detail explanation I would change my mind. Although I am not super concerned about this, it doesn't seem like it would be the type to  cause a race condition considering it's on a button (may consider a debounce if react doesn't already do a debounce).

PR #1352, opened 2019-08-06
TFW-4259 - allowed trending positives and negatives on donut and vert bar graphs

troyblank:
Not 100% sure the data will come back this way as props of `Trending Positive` and `Trending Negative` (included @weeznerps on here to maybe confirm) but the ticket says it will be these names and it's easy to change so I think this work stands on it's own legs minus that detail.

Screen Shot 2019-08-06 at 10 52 56 AM

Screen Shot 2019-08-06 at 10 53 11 AM


components/alert/alert.scss
```diff
@@ -50,9 +50,9 @@
 }
 
 .success {
```
tVeratti: This is a classname I could see colliding - we should BEM it to be safe.
`.alert--success`

tVeratti: Might actually be something we make a separate story for since this is a pre-existing problem.

components/interactionEditor/timeline/timelineNode.scss
```diff
@@ -158,20 +158,20 @@
 
 .negative {
```
tVeratti: Same here, we should create a story to update these styles as BEM as well.

tVeratti: negative/positive are just too generic.

components/alert/alert.scss
troyblank: Yea agree with new story, this will never collide till we make it beta.scss anyway.

PR #1353, opened 2019-08-06
TFW-4288 - Use label for displayValue on campaignName

tVeratti:
Most dimensions work with `value` as the `displayValue` as well, but `campaignName` is one exception.

@jasonhale  I'm not sure if you have any ideas on a better solution that's less hard-coded, but I'm not sure how to tell the difference between campaign and the rest. We talked about this last week (using value) so I tried to avoid changing the behavior for any dimension other than Campaign Name.


components/universalSearch/helperModal.js
```diff
@@ -42,14 +42,19 @@ export class HelperModalComponent extends PureComponent {
...
+    onHelperInput = (type, displayValue, value, rawValue, label) => {
         const { activeDimension } = this.props;
         let currentSelection = { displayValue, value, rawValue };
 
         // Update the latest selection so that it's always ready
         // to be committed to the query (and visually updated).
         let selections = [...this.state.selections];
 
+        if (activeDimension.name === 'campaignName') {
+            // Campaign value is not the same as label, so it needs a specific override.
```
troyblank: Not fully sure this comment makes sense, shouldn't it be campaign display value is just not existent thus you need to use label?

tVeratti: No it exists, but it's the campaign id which we don't want to render in the selections :)

troyblank: Well why the hell is the displayValue the id? seems odd to me have we asked why?

jasonhale:
@tVeratti, this is the same place I would've suggested.  I don't know if you want to pass the label back, or pass campaigns (via channel store) to the modal and do a lookup there based on value that's returned on `onChange`.

components/universalSearch/helperModal.js
jasonhale: Because on Autocomplete, the default displayLabel is the value being returned - which is the campaign ID.  So it needs to be explictly set here (much like having to format the display value for dates).

PR #1361, opened 2019-08-07
TFW-4272 - Add new MagnifyingGlass Icon

jasonhale:
~Added as a new one as it doesn't replace the current search icon.~  (Meant for the navigation only)
Replaced 24px version of current IconSearch with updated icon.

Screen Shot 2019-08-07 at 2 20 12 PM

Updated the demo a little, too, to reflect the related changes in RCI for the search page and navigation.


components/icons/search.js
```diff
@@ -7,8 +7,8 @@ export default function IconSearch({ size = 'large', filled = false }) {
 
     if ('large' === size) {
         path = filled
-            ? 'M8.498 17.623l-5.937 5.938a1.5 1.5 0 0 1-2.122-2.122l5.938-5.937A9.706 9.706 0 0 1 4.5 9.75C4.5 4.365 8.865 0 14.25 0S24 4.365 24 9.75s-4.365 9.75-9.75 9.75a9.706 9.706 0 0 1-5.752-1.877zM14.25 16.5a6.75 6.75 0 1 0 0-13.5 6.75 6.75 0 0 0 0 13.5z'
-            : 'M8.527 16.887l-6.822 6.817A1 1 0 0 1 .291 22.29l6.821-6.817A9.46 9.46 0 0 1 5 9.5a9.5 9.5 0 1 1 9.5 9.5 9.46 9.46 0 0 1-5.973-2.113zM14.5 17a7.5 7.5 0 1 0 0-15 7.5 7.5 0 0 0 0 15z';
+            ? 'M7.926 18.193l-3.868 3.865a1.5 1.5 0 0 1-2.12-2.122l3.867-3.864A9.457 9.457 0 0 1 4 10.5a9.5 9.5 0 1 1 9.5 9.5 9.457 9.457 0 0 1-5.574-1.807zM13.5 17a6.5 6.5 0 1 0 0-13 6.5 6.5 0 0 0 0 13zm4.5-6a1 1 0 0 1-2 0 3 3 0 0 0-3-3 1 1 0 0 1 0-2 5 5 0 0 1 5 5z'
+            : 'M8.237 17.175l-4.532 4.53a1 1 0 0 1-1.414-1.415l4.532-4.53a8.5 8.5 0 1 1 1.414 1.414zM13.5 17a6.5 6.5 0 1 0 0-13 6.5 6.5 0 0 0 0 13z';
     }
 
     if ('medium' === size) {
```
troyblank: this is not a huge deal but seems like all new ones use a `switch` statement. which I am pretty sure you were the one to come with ;) Look at the save icon for an example if you want to conform; but it's up to you, not a blocker.

jasonhale: nah.  that makes sense.  would keep it from crashing if and unsupported value for size is passed.  I'll add that before rebasing.

jasonhale: @troyblank - added

PR #1362, opened 2019-08-07
TFW-4093 - Move Search Title to header

tVeratti:
Screen Shot 2019-08-07 at 12 34 00 PM
Screen Shot 2019-08-07 at 12 37 28 PM
Screen Shot 2019-08-07 at 1 52 07 PM


components/universalSearch/title.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import classNames from 'classnames';
+import './title.beta.scss';
+
+export default function Title({ title, queries, bookmarks = {}, messages = {} }) {
+    const { activeBookmark, isDirty } = bookmarks;
+
+    let searchTitle;
```
jasonhale: The only tiny thing i can see is this name.  It's confusing since it's actually the bookmark title, but is called `searchTitle`.  Threw me off.

Not a blocker.  Everything works.  This one bit was confusing.

troyblank: yea I agree with this, why not `bookmarkTitle` ?

tVeratti: If there is no bookmark but queries exist, it is still present to say `New Search` so I didn't want it to seem like the direct pull of the bookmark title when it's not that.

tVeratti: I can change it if after that explanation it's still confusing.

jasonhale: the "New Search" part is fine.  It's the var name `searchTitle` that is confusing because that's actually the bookmark title.

However, looking through the design and UI flow, the concept of "search" changed... any query that is started is given the prompt for a name (aka bookmark).

It's still... awkward.  But I get it.  I'm ok with it as is, then.  Thanks!

PR #1363, opened 2019-08-07
TFW-4094 - Updated edited bookmark title style

tVeratti:
REBASE BEFORE MERGE
Screen Shot 2019-08-07 at 3 09 15 PM
Screen Shot 2019-08-07 at 3 09 39 PM

Turns out I already did a lot of the work for myself on TFW-4093, so the styles were pretty straightforward here!


components/universalSearch/title.beta.scss
```diff
@@ -29,13 +29,11 @@
...
-
-.search-title--edited {
-    .search-title__text span {
-        @include italic;
-    }
+    margin-left: $space-xs;
+    padding: 0 $space-xxs;
+    border: 1px solid $color-grey-4;
+    border-radius: $border-radius;
+    font-size: 1rem;
```
jasonhale: Looks like this uses the same styles as `%label-5`

jasonhale: Sorry, don't mind me.  There's already a ticket in the backlog to update the header styles.  These header styles applied to the search header are supposed to be universal page header styles, not just on Search.  So I imagine little tweaks like above can be handled with the effort in the other ticket.

PR #1364, opened 2019-08-08
TFW-4145 - added icons to search helper selections for channel

troyblank:
Screen Shot 2019-08-08 at 9 26 42 AM

I attempted to pass the info from the input to the selection first but realized it's a very unique ask to have an icon associated to a value so I glued them together using the helper map.


components/universalSearch/helperModal.beta.scss
```diff
@@ -29,7 +29,6 @@ $color-inactive: rgba($color-black, .5);
     .helper-modal__or {
         display: flex;
         align-items: center;
-        justify-content: space-between;
```
jasonhale: removing this affects the "Add another (OR) +" button.

Screen Shot 2019-08-08 at 11 15 56 AM

troyblank: OK fixed this thanks!

Screen Shot 2019-08-08 at 1 36 19 PM

PR #1367, opened 2019-08-08
TFW-4082 - moved bookmark button to the search inputs row

troyblank:
This is just to MOVE the bookmark button, there is another story to handle the new styles of the button I will take on next.

Screen Shot 2019-08-08 at 12 06 41 PM


components/universalSearch/searchBar.js
```diff
@@ -131,6 +132,38 @@ class SearchBar extends PureComponent {
     //     return match.attributes || ['autocomplete'];
     // }
```
jasonhale: Would you mind terribly removing those comments above?  We don't need those anymore.  Or we can do in a different commit.  No bigs.

troyblank: Yea there is tons of these that need to be removed; considering it's not really part of this work I feel we need just a general cleanup commit at this point there is probably half of them no longer needed for reference. I will hold off for now but will get this in my next PR.

PR #1370, opened 2019-08-08
TFW-4090 - Add results counts to US header

jasonhale:
Screen Shot 2019-08-08 at 4 55 55 PM


jasonhale:
Yeah, i know what you mean.  But there's a ticket in the backlog already to update the header styles (TFW-4081).  Would be better to handle there.

troyblank:
@jasonhale  good point, i made a comment on that ticket to tackle it there.

PR #1374, opened 2019-08-12
TFW-4122 - add to HelperMap for 'range' attribute

jasonhale:
Agent Tenure is going to be using the new attribute 'range'.


components/universalSearch/helperSwitch.js
```diff
@@ -66,6 +66,12 @@ class HelperSwitch extends PureComponent {
 
         return onValueChange('autocomplete', value);
     }
+
+    onChange_range = e => this.onChange(
```
tVeratti: We will just need to be certain that there are never more than 10 options, otherwise typing into the filter field will cause chaos with the selections. We might want to consider some kind of check at some place for that, otherwise we could have an issue to deal with later.

jasonhale: I worried about that, too.  Asked Jordan if he knew if that'd ever be more than 10.  However, if they're ever more than 10, this will have to be converted to an autocomplete since the design won't allow more than 10 items... which will require reworking a little on the API and here.

we can confirm with Casey.

tVeratti: Sounds good. I'll approve here and if anything comes from that convo it'd prob be another story anyway.

jasonhale: @tVeratti  - talked to Casey and Josh.  If the tenure list is ever more than what it is now, there'd be a lot more considerations going on and this can be addressed then.  Thanks for bringing it up, though.

PR #1376, opened 2019-08-12
TFW-4144 - Add RangePicker

tVeratti:
I have a few more interaction tests to add, but mostly this should be the working component.
Screen Shot 2019-08-12 at 2 38 47 PM


components/durationPicker/numberPicker.js
```diff
@@ -17,19 +21,27 @@ export default class NumberPicker extends PureComponent {
         }
     }
 
+    componentDidUpdate() {
+        this.positionSuffix();
+    }
+
+    positionSuffix = () => {
```
jasonhale: This works good.  Could also be done via similar markup and css - to keep js use down if preferred.

tVeratti: How would you recommend it be done without js? Figured measuring a span and applying its width value to another would be cleanest in js but open to suggestions!

troyblank: Yea I am a bit confused why JS is needed on this. On the surface it seems overkill; I would start to tackle this by just adding some padding to the left of the suffix and let it horizontal center inline with the value with something like flexbox. BUT i could be missing a major hurdle you encountered.

troyblank: As a bonus if you did center the value + suffix in the container you would get a more centered overall look as the smaller it gets the more off to the side it appears.

Screen Shot 2019-08-12 at 4 02 37 PM

tVeratti: The issue is that for a few levels deep (`Input` component as well as `NumberPicker`) the inputs are blocks that are 100% width. This doesn't really leave a convenient spot for an element to live in between it and the buttons.
Screen Shot 2019-08-12 at 4 15 14 PM

In order to push a suffix element down into the position it we'd have to squish the input down and that would make it only focusable on the left side - unless click handlers were added to the suffix to focus the field - at which point I don't think any complexity has been saved.

jasonhale: Sorry for the delay in response.
My suggestion on CSS-only was based on the structure before your latest commit.

The idea would be the same... put in hidden content and place suffix based on that, but using styles to visibly hide content and place the suffix.  Below is rough. Could be refined a little.

NumberPicker styles...

```css
// ... remove this bit...
// .number-picker__suffix {
//     width: 50%;
//     color: $color-grey-1;
//     pointer-events: none;
// }

.number-picker__ghost-decoration {
    position: absolute;
    right: 32px;
    left: 32px;
    text-align: center;
    pointer-events: none;
}

.number-picker__ghost-input {
    display: inline-block;
    position: relative;
}

.number-picker__ghost-value {
    visibility: hidden;
}

.number-picker__ghost-input:after {
    display: block;
    position: absolute;
    top: 0;
    bottom: 0;
    left: calc(100% + 8px);
    content: attr(data-suffix);
}
```

NumberPicker jsx
```jsx
{suffix && (
<div className="number-picker__ghost-decoration">
  <span className="number-picker__ghost-input" data-suffix={suffix}>
    <span className="number-picker__ghost-value">
      {displayValue}
    </span>
  </span>
</div>
)}
```

result...

Screen Shot 2019-08-13 at 9 08 58 AM

jasonhale: Caveat: this solution doesn't address @troyblank 's note about small input fields where the suffix collides with the "+" button (looking off-centered).  That could be mitigated by restricting the min-width.  The center could also use some flex-box (or grid) lovin' to give the inputs some space in narrower containers by shrinking that dividing area.

tVeratti: I think the approach I have now does center things a little better, and I have an easy fix for the clicking which I will push up shortly (using a label instead of a span with `htmlFor` linking back to the input). I'm going to stick with it unless you guys see any major issues - though i appreciate the examples and the initial push to move to CSS.

jasonhale: I dunno, the centering is still awkward to me with the 1/2 width input.  But, it's also not the easiest problem to solve.  (and there's often more than one way to solve things)

troyblank: What about adding the suffix to the input itself and just filter it out when needed? This will allow you to center easily. Also if the user does not re-add just add it back programatically. I am on the fence though your solution is super nice because the user can not alter it and it's obviously not something they can edit.

tVeratti: I did that originally, and it throws tons of wrenches into the NumberPickers themselves because it becomes a string and a bunch of assumptions about a `Number`Picker go out the window.

troyblank: That's fair... just a thought I didn't really think hard about it :)

troyblank: Actually. this sparked a lot of ideas for me.... it appears we are treating the inputs as numbers when they really are strings anyway.... so maybe we do need a filter for not only possibly suffix stuff but for stuff like "e" which is a valid number input :( so Number picker seems to be flawed even in it's current state.

Screen Shot 2019-08-13 at 1 58 02 PM

We might have to throw out our assumptions about Number picker just to make things better anyway... thoughts?

tVeratti: I think you're getting pretty far outside of our requirements and scope for these thoughts, even though they aren't false :)

troyblank: Oh yes I know it's outside of this work for sure sure, but arguably maybe we need to push for more time is what I am saying. With this thought in mind we ask for time to do it a little better. I am fine either way though the solution you came up with works for sure it all depends on how refined we wanna get.

components/durationPicker/numberPicker.js
```diff
@@ -1,13 +1,18 @@
...
 import './numberPicker.beta.scss';
 
 export default class NumberPicker extends PureComponent {
     state = {
-        value: 0
+        value: this.props.value || 0
+    }
+
+    componentWillMount() {
+        this.id = random(100);
```
jasonhale: Why not just require an ID to be passed?  We have a convention in place with the visualizations.

tVeratti: What's the benefit of adding the extra prop everywhere it's used? It'd need to be unique across each numberpicker, which there are at least two of in each range picker, and more in the duration? Seems more cluttery when we just need something for the labels to hook into.

jasonhale: This has the chance - however rare - to have the same ID as another numberPicker.  At least if it's a prop, it's be more guaranteed to be unique.

It's an unfortunate consideration for using an ID.

tVeratti: Beefed up the random so it won't collide except for an absolute miraculous scenario :)

troyblank: wait what about using lodash unique? i know js native doesn't have one, but i know lodash unique does https://lodash.com/docs/4.17.15#uniqueId

troyblank: I agree with Jason here we should make sure this number is more unique though :)

tVeratti: Updated to use uniqueId!

PR #1378, opened 2019-08-13
TFW-4286 - converted comboBarLine to use external totals for calculating percentages

troyblank:
DO NOT MERGE THIS YET UNTIL THE CURRENT SPRINT IS FINISHED OR ISOLATED.

This is step 1 of 2.

For `<ComboBarLine />` this uses another set of data (the one without any search filtering on it) to calculate percent.

This will require work on RCI to get data without dimension filtering for the combo bar line if it doesn't get it in the first place and then switch it to no longer loading once it has that.

Screen Shot 2019-08-13 at 4 52 38 PM


components/visualizations/comboBarLineChart.js
```diff
@@ -26,14 +26,18 @@ export default class ComboBarLineChart extends PureComponent {
...
+        const allDataXMatches = allData.filter(d => xValues.includes(d[xKey]));
+        const filteredAllData = filterDataByDate(xAxisMode, allDataXMatches, { xKey, yKeys: [yKey], zKeys }, ['zKeys']);
         let filteredData = filterDataByDate(xAxisMode, data, { xKey, yKeys: [yKey], zKeys }, ['zKeys']);
 
         if (Number.isInteger(precision)) {
             filteredData = updateDataDecimalPrecision(filteredData, zKeys, precision);
         }
 
-        return addPercentsToData(filteredData, yKey);
+        return addAllRecordCountPercentsToData(filteredData, filteredAllData, xKey, yKey);
```
tVeratti: We map/filter/loop over these data points a lot of times for slightly different reasons. I can't think of a super simple way to avoid that right now, but consider it food for thought. If we could somehow reduce those loops once or twice and do multiple operations, we'd save ourselves quite a bit of extra browser power there :)

PR #1381, opened 2019-08-15
TFW-4280 - allowed line graph to accept all data calculations

troyblank:
This does the same thing for line as my previous PR for combo line when it comes to adding up percents using the total counts prior to dimension filtering. (to be paired with this PR on rci > https://github.com/topbox-rci/RCI/pull/1293)

I also hooked the all data props up using vizzy nizzy and fixed a 0 division bug I had with the util to calculate all these percentages. Enjoy!


troyblank:
I just realized I forgot to update the `loading` prop of any component that needs both sets of data (all and filtered). I will be right back with this :)

PR #1382, opened 2019-08-15
NextJs

tVeratti:
Here's an initial move to NextJS for the component demos.
It's not 100% solid for all uses yet (I haven't tested building for publishing, so I didn't remove all of the grunt stuff yet).
Screen Shot 2019-08-15 at 4 19 52 PM

We should be able to take the raw demo files and basically yoink their contents into these `demo.js` files. This might require reorganizing a few things along the way, and ending css-modules along the way.

Let me know what you guys think of this overall. I still want to get more done here before calling it good, but you can at least run and see the demos as of now.


assets/js/demo/colorCopy.js
```diff
@@ -14,13 +14,15 @@ function injectColorInfo() {
     });
 }
 
-export function onColorClick() {
+// Sorry Troy :(
```
troyblank: download

troyblank: This if fine actually, the demo stuff does not need to be tested really since it's not customer facing. If we can't we can't.

jasonhale:
Haven't got to go through it yet.  One thing to check before i forget... packaging this for publishing.  Should consider checking that so the demo files are not included.  It's just bloat at that point.

Unless there's an expectation devs using this in projects would want to run the demo as a reference.  That's a whole different can of worms.

tVeratti:
> 1. given our `next.config.js` does not have `exportPathMap` I assume we have yet to build this site statically. If we want to continue to host on s3 or even move to something like netlify or now we will need to make sure we build this to continue to be static. Was the plan to publish this site statically?

We won't need to create an `exportPathMap` because we don't need to have super pretty urls, nor do we need save queried data for static pages. If we move to calling a mock api, we can shift things, but for now it's not part of our demo project at all.

> 2. Might go along with #1 but to me it seems like the site is broken until a lot of the compiling is done on the fly (clicking buttons seem unresponsive); is this the compile on the fly something that we are going to keep and if so how can we let the user/developer know the site is not broken? Some options I have thought about could add a giant load graphic or do the build step prior to launching the site with something like `exportPathMap`. If we are going to publish statically anyway this might be killing two birds with one stone.

The very first render of any page will take a sec to build, then you're good. Should only take 1-2 seconds for the HMR to load your changes onto the page. If you see anything slower after the initial building, let me know.

> 3. Will all the components be ready to see/play with after this is merged in? or do we expect to be building this as we have to work on it? Asking as this will bloat some time it takes to update old components.

I will add as many as I can with any time "leftover" from my 21pt estimate - however I expect this will be something we all contribute to as we move forward. Some components might want to be converted from modules and stuff as we go, it should be used as an opportunity to make those long overdue changes at the same time.

> 4. Keeping the color and icon reference charts are important; noticed everything is in one giant list. When i was clicking through I thought it would be super nice to have base components and large components separated along with a home for guides like the colors.

I've changed this to be an explicit list of demos. This will allow us to create any of demo with any grouping or components etc. So colors etc will fit nicely with their own demo exported.

> Also one small thing be sure to update `.nvmrc` if this solution requires a higher version of node than the old components does, lot of robots and humans use that file to make sure they are running the repo with the right version.

I didn't seem to have any issues with our version that I could tell. Dynamically importing was causing some issues when I didn't have all of the component dependencies loaded first.

PR #1383, opened 2019-08-16
 TFW-4343 - Update Header styles

jasonhale:
Applies the new header styles to HeaderPane for consistent page titles.

The design with the slash separating sections of the title is going to be used in more than just Search, so moved that feature introduced by `Title` into `HeaderPane`.


components/headerPane/headerPane.js
```diff
@@ -22,9 +22,37 @@ export default class HeaderPane extends PureComponent {
...
+
+        let attrTitle = titleAttr || title;
+        let htmlTitleArr = title;
+
+        if (Array.isArray(title)) {
+            if (!titleAttr) {
+                attrTitle = title.length > 1 ? title.join(' / ') : title[0];
+            }
+            const titleSeparator = <span className={'header-pane__title-separator'}>/</span>;
+            // add dashes...
```
troyblank: This function rambles a little; this could easily be a util function called `addDashes` to make things easier to test and more modular and documented. What do you think?

jasonhale: I don't think it's necessary as a util, since it's only used in this one spot. And tested well (i think) in the tests currently.

If you feel it's necessary, that's not a problem. I can move this bit up a bit. Does make it a little more complicated with arguments. But should be ok.

troyblank: Well maybe not a util, but at least a separate testable function. I would be fine with it being in the same file.

components/headerPane/headerPane.js
```diff
@@ -22,9 +22,37 @@ export default class HeaderPane extends PureComponent {
         return '0';
     }
 
+    renderTitle = () => {
+        const { title, titleAttr } = this.props;
+        if (!title) return (<h1 />);
```
troyblank: Should this be `null` ? Why a blank h1?

components/headerPane/headerPane.js
troyblank: To clarify this function seems like it does two things which I have always learned is room for improvement; it renders a title but then also adds dashes if it's needed. Why not render the title and then add dashes in another function so both functions have a specific task?

components/headerPane/headerPane.js
jasonhale: Previously was rendering an empty h1.  I think it was originally like that to keep the header from collapsing - but not currently the case.  Updating.

jasonhale: now `null`.  :)

components/headerPane/headerPane.js
jasonhale: updated

components/headerPane/headerPane.js
```diff
@@ -22,9 +22,38 @@ export default class HeaderPane extends PureComponent {
...
     }
 
+    addTitleSeparator = (list) => {
+        const newList = [];
+        list.forEach((t, i) => {
+            newList.push(t);
+            if (i !== list.length - 1) newList.push(<span key={i} className={'header-pane__title-separator'}>/</span>);
+        });
+        return newList;
+    }
```
troyblank: cool thanks so much or doing this, keeps both those nice and shallow and easier to read.

PR #1384, opened 2019-08-16
TFW-4280 - allowed line graph to accept all data calculations

troyblank:
This does the same thing for line as my previous PR for combo line when it comes to adding up percents using the total counts prior to dimension filtering. (to be paired with this PR on rci > topbox-rci/RCI#1293)

I also hooked the all data props up using vizzy nizzy and fixed a 0 division bug I had with the util to calculate all these percentages. Enjoy!


components/universalSearch/visualizations/dashboards.js
```diff
@@ -100,7 +105,12 @@ export default function Dashboards() {
 
     const errorComponent = chart => (
       <Error key={chart.id} label={chart.label} type={chart.type}>
-        <VizzyNizzy {...chart} results={modifiedResults[chart.id] || {}} timezone={timezone} />
+        <VizzyNizzy
+          {...chart}
+          results={modifiedResults[chart.id] || {}}
```
tVeratti: It seems like the `||` might be redundant since you set them above on lines 76/77 right?

troyblank: Yup you totally are right... I removed the ` || {}`

PR #1385, opened 2019-08-16
TFW-4359 - Fix percentage helper query format

jasonhale:
API is expecting to/from, instead of min/max.


jasonhale:
None of those specific models for requesting dimension values are.  I'll pass the note to the gang.

PR #1386, opened 2019-08-16
TFW-4344 -  Fix modified/unsaved search title label

jasonhale:
Screen Shot 2019-08-19 at 10 59 49 AM
Screen Shot 2019-08-19 at 10 59 41 AM


jasonhale:
@troyblank - updated per Casey.  Font to semibold.  Updated colors for unsaved and fixed a height prob.

PR #1388, opened 2019-08-16
TFW-4345 - US - results count small screen

jasonhale:
Screen Shot 2019-08-16 at 5 08 05 PM


jasonhale:
@troyblank - that placement fix is in the other PR #1386.  Let me see if Casey wants to update those styles, i'll merge and can show the result.

troyblank:
@jasonhale  ah! ok cool pulled that in and looks good.

Screen Shot 2019-08-19 at 9 58 24 AM

PR #1391, opened 2019-08-19
TFW-4361 - Wordbars, only show necessary sentiments

jasonhale:
This checks the sentiments passed and only renders out the correct ones in Wordbars that use sentiment binning (for cases that use 3 sentiments vs 5).


components/visualizations/wordBar.js
```diff
@@ -47,6 +47,10 @@ class WordBar extends PureComponent {
         )
     }
 
+    static defaultProps = {
+        hasTabNavigation: true
```
tVeratti: Nice :)

components/visualizations/wordBar.js
```diff
@@ -147,18 +152,52 @@ class WordBar extends PureComponent {
...
+                    labels.push(s.label);
+                    colorMap[s.label] = convertLabelToColorKey(s.key);
+                }
+            });
+        } else {
+            SENTIMENTS.forEach((s) => {
+                labels.push(s.label);
+                colorMap[s.label] = convertLabelToColorKey(s.key);
+            });
+        }
```
troyblank: Instead of making this and if and else, couldn't you make an array above that just includes all colors so that you can do everything inside the `SENTIMENTS.forEach((s) => {`. .... For example if sentimentObj is not defined set the `receivedSentiments` to `[...SENTIMENTS]` (or whatever that array would reduce to be with just keys in it). I think you could reduce that last if else into just one `forEach` and set the green lighted sentiments to everything if there is no `sentimentObj`.

jasonhale: @troyblank - Updated.   Can't help but feel a little over-defensive here in this bit.  Trying to cover all bases.

troyblank: 💥

PR #1392, opened 2019-08-19
TFW-4364 - fixed auto complete helper selection count

troyblank:
Ok, might need a second pair of eyes on this; but for the life of me I could not figure out why on auto complete helpers we were keeping a blank "active" selection. I understand why on one's that have user input it's needed though.

I removed it because it was causing the total selections to be 1 when all were auto complete selections where removed causing the disabling of add and omit button to not work correctly. I have tested this code against RCI and can not find any downsides or errors to this removal. But please holler if you do as I was sorta out of town when this all was built :).


components/universalSearch/helperModal.js
```diff
@@ -102,11 +102,10 @@ export class HelperModalComponent extends PureComponent {
                     value,
                     channel: get(stateCurrentSelection, 'value.channel', 'bodyFull')
                 };
+
                 currentSelection.displayValue = displayValue;
+                selections.push(currentSelection);
```
jasonhale: So, looking into this, it changes how channel+autocomplete (Transcript Topic) works.  With these changes, after selecting a topic, you're still able to affect the last selection in the selection list - which isn't expected behavior.  May have to run this one through the biz peeps.

jasonhale: Possibly, in `renderHeader()`, check `selections.length` and if there's a `selections[selections.length -1].value.value || selections[selections.length -1].value`... ?

troyblank: I did not intend to change the way autocomplete works; I was trying to fix the bug where autocomplete never disables after selections are added. I will dig into this more as I am not sure how they are different with this change.

tVeratti: Well damn, I tried all of the freetext and freetext channel demos but didn't try the autocomplete channel. Another one for the eagle eye :)

troyblank: @jasonhale  thanks for spotting this; I didn't realize that is what I did so  I adjusted the `renderHeader()` check instead now.

PR #1395, opened 2019-08-20
TFW-4352 - allowed rendering of the geo horizontal bar visualization type

troyblank:
This is pretty basic; just needed a new vizzy nizzy case to get all 3 tabs working and rendering.

I confirmed this is the `type` that will show up from the API, the rest is all made up but should be good based on some small convos.

Screen Shot 2019-08-20 at 12 56 24 PM


troyblank:
@jasonhale yea good call, i will make a demo to add to the search vizzy page soon so it's not just in the base layout crap.

troyblank:
@jasonhale added demos to this mug; all is good!

PR #1398, opened 2019-08-20
TFW-4141 - Render Add Again button

tVeratti:
I spoke with Casey about when this should reset (navigating away or refreshing) and he said it should clear out in those scenarios because the workflow this is assisting would be terminated if they are navigating away anyway.

Screen Shot 2019-08-20 at 3 27 05 PM

Screen Shot 2019-08-20 at 3 27 13 PM


tVeratti:
I moved the button 20px closer to the `Add New Dimension` button compared to the screenshots shown, fyi.

assets/sass/skins/reactTooltip.beta.scss
```diff
@@ -52,7 +52,6 @@ $color-tooltip-backing: $color-black-2;
     &.place-right {
         &::after {
             right: -5px !important;
-            border-width: 5px 5px 5px 0 !important;
```
jasonhale: Why just removing this one and not the others?

components/universalSearch/searchBar.beta.scss
```diff
@@ -37,7 +32,17 @@ $dimensions-width-end: 40rem;
...
+    @include breakpoint-below($breakpoint-mobile) {
+        flex-direction: column;
+        align-items: flex-start;
+    }
+
+    display: flex;
+    align-items: center;
+
+    .auto-complete {
+        display: inline-block;
```
jasonhale: It's inside a `flex` object - and since it has a declared width, this `inline-block` doesn't seem necessary.  ?

assets/sass/skins/reactTooltip.beta.scss
tVeratti: I didn't want to accidentally change things I wasn't testing as thoroughly. I don't even fully understand why this override is here since it breaks the arrow's display. But the up/down arrows seem to look fine so I was worried about breaking those.

components/universalSearch/searchBar.beta.scss
tVeratti: Yeah honestly just a leftover property from the first few changes :) Removing!

assets/sass/skins/reactTooltip.beta.scss
jasonhale: AH.  In that case, the problem isn't the `border-width`, it's the `right: -5px`.  That should be `left: -5px' since the triangle should extend to the left of the tooltip. Then, keep the border-width as it was.

Removing, the arrow of the tooltip goes to a 6px triangle instead of 5px... not a big deal, really, but the lack of styles for the `.place-right` is _really_ confusing here.

tVeratti: I originally tried adjusting the position but it didn't seem to remove the problem, only allowing the border-width to dip back to its default fixed it. I can try again to see if flipping it works, but it seemed unresponsive to changes there.

PR #1399, opened 2019-08-21
TFW-4357 - Close bookmark selector on select

jasonhale:


components/universalSearch/bookmarking/header.js
```diff
@@ -5,13 +5,13 @@ import { IconBookmark } from '../../icons';
...
-        this.onSelectorOpenHandler = this.onSelectorOpenHandler.bind(this);
+    onSelectorOpenHandler = () => {
+        this.selector.clearSearchValue();
     }
 
-    onSelectorOpenHandler() {
-        this.selector.clearSearchValue();
+    closeBookmarkSelector = () => {
+        // referencing `instanceRef` via HOC onClickOutside
+        this.hamburgerNavigation.instanceRef.handleClickOutside();
```
troyblank: This bothers me a tad because we are forcing a click outside action when it is not actually clicking outside. And if the click outside behavior changes it could potentially cause some confusion.  Would love to call something like `closeNavigation` or anything that is indicative of closing the hamburger menu.

jasonhale: Completely reasonable.  Update incoming.

jasonhale: @troyblank - updated per suggestion, and consolidated a bit.

components/universalSearch/bookmarking/selector.js
```diff
@@ -29,9 +29,10 @@ export default class BookmarkSelector extends PureComponent {
     }
 
     onItemSelect = (item) => {
-        const { history, location } = this.props;
+        const { history, location, onBookmarkSelect } = this.props;
```
jasonhale: This was just to cut down on some of the warning noise in tests.

PR #1418, opened 2019-08-30
TFW-4224 - Update Button styles

jasonhale:
This updates the new design buttons to the latest iteration.  Tried to cleanup/consolidate a little, and updated instances where needed.

Also updated legacy buttons in place.
Didn't delete legacy styles yet... If you think we're ready, I'll do that here, too.

Updated demo app for positioned components (namely `SignIn` and `ResetPassword`).

One thing missing now is a visible active state (differentiated from hover) on the updated buttons (now that the dropshadow is removed).  That's an ongoing convo with Casey.  ~Might see about doing that in another ticket.~ (there's a ticket now: TPB-2349)


troyblank:
Wait maybe ignore me I just realized this might be built on the nextjs stuff >< we need to remove the build command here...

PR #1419, opened 2019-08-30
TFW-4444 - Remove score from wordcloud tooltip

tVeratti:
Screen Shot 2019-08-30 at 12 38 05 PM

Added `ReactTooltip` to the demo page base as well.


troyblank:
Also sorry one more thing the command `npm run build` is broken here, consider removing to avoid confusion for people like me 20 mins ago :P

